PR #183, opened 2017-03-13
recipes/switchdev/l3-000-minimal.py: add ipv6 route before ping

L8LiLiang:
Signed-off-by: liali <liali@redhat.com>

Hi pirko,

I am liliang from redhat, i am testing switchdev now.
I found a issue in recipes/switchdev/l3-000-minimal.py.
It should add ipv6 route before ping, or it will fail.


jpirko:
please send this to the mailing list lnst-developers@lists.fedorahosted.org

Thanks.

PR #205, opened 2021-06-29
Poetry split

Axonis:
Description:

With redesign of LNST we have started importing the main package to different dependent projects. However, the process of importing is suboptimal and requires knowledge of workspace with all of the projects. 
We've decided to move the installation and setup part from setuptools to newer poetry dependency manager, that handles installation of the framework as well.

With this move I've also decided to remove some unused files and directories, such as recipes/, results_xslt/, etc.

Reviews:
@jtluka @olichtne


recipes/examples/python_recipe.py
```diff
@@ -1,128 +0,0 @@
-#!/bin/python3
```
olichtne, line 1: This is a valid python recipe I think, we should probably just move this into the `lnst/Recipes/...` hierarchy

olichtne (commented):
Other than the one comment, looks good imo.

jtluka:
Are we fine with following code? I'm just wondering why we need to load the **/dev/null**. Can this be completely removed with the same result?

https://github.com/LNST-project/lnst/pull/205/files#diff-afec160d924a703fb763bc1bafe952d977ee526ec5870dee4833682882c5d76eR63

@olichtne @Axonis

jtluka:
Another concern here is that likely we'll break the Fedora package once the **setup.py** is removed. Should not be a blocker for  merging this in, just a heads up.

@olichtne is this an issue?

jtluka:
> Another concern here is that likely we'll break the Fedora package once the **setup.py** is removed. Should not be a blocker for merging this in, just a heads up.
> 
> @olichtne is this an issue?

Now I see that lnst.spec refers to a specific tarball, so this should not be an issue.

olichtne:
> Are we fine with following code? I'm just wondering why we need to load the **/dev/null**. Can this be completely removed with the same result?
> 
> https://github.com/LNST-project/lnst/pull/205/files#diff-afec160d924a703fb763bc1bafe952d977ee526ec5870dee4833682882c5d76eR63
> 
> @olichtne @Axonis

Yeah, I already talked about this with jozef and experimented with it, it should work just fine without `/dev/null`, but since it also doesn't break anything we can resolve this later.



> > Another concern here is that likely we'll break the Fedora package once the **setup.py** is removed. Should not be a blocker for merging this in, just a heads up.
> > @olichtne is this an issue?
> 
> Now I see that lnst.spec refers to a specific tarball, so this should not be an issue.

Yeah, this shouldn't be a problem, the fedora package builds from an already uploaded tarball release that we did a long time ago and doesn't pull new sources from github.

olichtne (commented):
Sorry for not specifying this earlier... This is a generic example recipe. `ENRT` is a specific group/set/package of recipes that we use at red hat. The generic example recipe is not part of this package.

Please move this up by one directory - `lnst/Recipes/ExampleRecipe.py`

recipes/examples/python_recipe.py
olichtne: resolved with commit https://github.com/LNST-project/lnst/pull/205/commits/658427390b75698fdc47080154bcbf6ff53f4a19

PR #212, opened 2021-07-23
lnst: add pkgutil style namespace packages

Axonis:
Description:
Removing top level init that was redundant and not used.

Updating the pyproject.toml, bumping project
version to 16 and python version to 3.6 and higher
to speed up installation.

Whenever LNST was installed as a dependent project
poetry was unable to stitch together multiple
packages and their modules.
Easiest solution is to update paths accordingly [0].

[0] - https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages

Tests:
J:5614799

Reviews: 
@olichtne @jtluka

### DIFF
--- a/install/lnst-slave.service
+++ b/install/lnst-slave.service
@@ -5,7 +5,7 @@ After=network.target
 
 [Service]
 Type=simple
-ExecStart=/usr/bin/lnst-slave
+ExecStart=/usr/local/bin/lnst-slave
 StandardOutput=null
 
 [Install]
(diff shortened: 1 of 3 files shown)


olichtne (approved):
looks good to me

Axonis:
Test: J:5629226

I'll also update the lnst-slave.service path of `ExecStart` due to the script being installed in the `/usr/local/bin` instead of `/usr/bin` when poetry config `virtualenvs.create` is set to false.

olichtne (approved):

PR #213, opened 2021-07-27
Evaluating flows in BaseFlowMeasurement and fixed evaluating infinity in NonzeroFlowEvaluator

enhaut:
Description:  
Replaced hardcoded result pass with evaluating in `BaseFlowMeasurement._report_flow_results()` to correctly evaluating flows with invalid duration.  
Fixed evaluating `inf` as valid passed result in `NonzeroFlowEvaluator.evaluate_results()`.

Reviews:  
@Axonis @olichtne @jtluka

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
@@ -268,6 +268,11 @@ class BaseFlowMeasurement(BaseMeasurement):
             for flow_results in aggregated_flow_results:
                 cls._report_flow_results(recipe, flow_results)
 
+    @staticmethod
+    def _invalid_flow_duration(result: FlowMeasurementResults) -> bool:
+        if result.duration <= 0:
+            return True
+        return False
 
     @classmethod
     def _report_flow_results(cls, recipe, flow_results):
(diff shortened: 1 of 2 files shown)


lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
```diff
@@ -297,8 +302,16 @@ def _report_flow_results(cls, recipe, flow_results):
                         cpu_deviation=receiver_cpu.std_deviation,
                         cpu_unit=receiver_cpu.unit))
 
-        #TODO add flow description
-        recipe.add_result(True, "\n".join(desc), data = dict(
+        recipe_result = True
+        flows = {"Generator": generator, "Generator process": generator_cpu,
```
olichtne, line 306: nitpick here... these are **metrics** or **measurement results**, not flows. all of these metrics are related to the same flow measurement.

olichtne (commented):
I don't think my nitpick is imporant enough to block this PR but I'll wait for review from others before approval.

lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
enhaut: Thanks, i just changed that.

olichtne (dismissed):

olichtne (approved):

PR #214, opened 2021-07-29
Add optional parameters for VLAN IDs

naggety:
Some switches are configured to allow only a small number of VLAN IDs (such as in klab.bos lab).

Allow passing an optional argument vlan_id/vlan_ids to specify what VLAN tags to use in the tests.

Note: only tested VlansRecipe

### DIFF
--- a/lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
+++ b/lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
@@ -5,7 +5,7 @@ from lnst.Common.IpAddress import (
     Ip4Address,
     Ip6Address,
 )
-from lnst.Common.Parameters import Param
+from lnst.Common.Parameters import Param, IntParam
 from lnst.Devices import GreDevice, VlanDevice
 from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
 from lnst.RecipeCommon.PacketAssert import PacketAssertConf
@@ -63,7 +63,7 @@ class GreTunnelOverVlanRecipe(
     host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
     host2.eth1 = DeviceReq(label="net1", driver=RecipeParam("driver"))
 
-    vlan_id = Param(default=10)
+    vlan_id = IntParam(default=10)
 
     offload_combinations = Param(
         default=(
--- a/lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
+++ b/lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
@@ -63,6 +63,8 @@ class GreTunnelOverVlanRecipe(
     host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
     host2.eth1 = DeviceReq(label="net1", driver=RecipeParam("driver"))
 
+    vlan_id = Param(default=10)
+
     offload_combinations = Param(
         default=(
             dict(gro="on", gso="on", tso="on"),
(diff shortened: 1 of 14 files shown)


lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
```diff
@@ -63,6 +63,8 @@ class GreTunnelOverVlanRecipe(
     host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
     host2.eth1 = DeviceReq(label="net1", driver=RecipeParam("driver"))
 
+    vlan_id = Param(default=10)
```
olichtne, line 66: I'd prefer using the `IntParam` class to do a little bit of type checking for the parameter. It would avoid potentially situations when someone would assign a string value to the parameter.

lnst/Recipes/ENRT/VirtualBridgeVlansOverBondRecipe.py
```diff
@@ -40,6 +40,8 @@ class VirtualBridgeVlansOverBondRecipe(VlanPingEvaluatorMixin,
     guest4 = HostReq()
     guest4.eth0 = DeviceReq(label="to_guest4")
 
+    vlan_ids = Param(default=[10, 20])
```
olichtne, line 43: i think it would probably be better if this was 2 parameters instead of a single parameter that expects a list of 2 numbers.

Right now you could do:

`VirtualBridgeVlansOverBondRecipe(vlan_ids=[123])` which would cause a traceback during test configuration

or

`VirtualBridgeVlansOverBondRecipe(vlan_ids=[1, 2, 3])` which just wouldn't use the value `3` and that could potentially be confusing?

instead if we had:
```
vlan1_id = IntParam(default=10)
vlan2_id = IntParam(default=20)
```
It resolves both situations and is potentially more clear? What do you think?

Another option would be to dynamically create as many vlans as the number of values in the `vlan_ids` list but at this point that's an unnecessarily big change to the recipes...

olichtne (changes requested):

lnst/Recipes/ENRT/VirtualBridgeVlansOverBondRecipe.py
naggety: I tried to use the same names everywhere (vlan_id for single integer and vlan_ids for arrays) for consistency and ease to use. I though in the case you provide a wrong number of elements, but I though that this won't be used very often, and if you use you probably know what you're doing.
But anyway, your suggestion seems good to me: typechecking, avoid backtraces, number of required IDs more clearly indicated, naming still consistent...

lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
naggety: Done

lnst/Recipes/ENRT/VirtualBridgeVlansOverBondRecipe.py
naggety: Done

naggety:
Pushed new commit resolving the changes requests. However I don't have now machines on loan to test them, and I don't have time the next days for it, so you can wait until I can make, at least, a basic testing.

olichtne:
Cool, looks good to me. @jtluka you worked with Vlan related recipes internally, can you also take a look? Do we want to schedule a beaker job test run for this?

jtluka:
> Cool, looks good to me. @jtluka you worked with Vlan related recipes internally, can you also take a look? Do we want to schedule a beaker job test run for this?

I will schedule a beaker test run.

jtluka:
> > Cool, looks good to me. @jtluka you worked with Vlan related recipes internally, can you also take a look? Do we want to schedule a beaker job test run for this?
> 
> I will schedule a beaker test run.

Beaker job id J:5686356.

jtluka:
> > > Cool, looks good to me. @jtluka you worked with Vlan related recipes internally, can you also take a look? Do we want to schedule a beaker job test run for this?
> > 
> > 
> > I will schedule a beaker test run.
> 
> Beaker job id J:5686356.

This merge request need to rebase on the current master to include the https://github.com/LNST-project/lnst/commit/1a862ee31623c9599489d8394dac93eaae16e66a

Without that patch we're unable to run tests in our lab.

jtluka:
Rescheduled tests in J:5686591.

jtluka:
> 5686591

@olichtne @naggety the beaker job completed without any issues, the vlan_id parameter is present in the recipe parameters. The defaults seem to be working fine. I'm ok with merging this in.

olichtne (approved):

olichtne:
thanks!

PR #215, opened 2021-07-30
ENRT.DevInterruptHWConfigMixin: refactor irq pinning

olichtne:
After some discussion with Paolo Abeni, the /proc/interrupt file should
be the source of information for the device irqs that we need to be
pinning to specific cpus. This is because it contains the "in use" list
of irqs in contrast with the "msi_irqs" file which contains the
"reserved" but potentially "not in use" irqs.

At the same time for the /proc/interrupt method to work properly the
device needs to be up when we read the file. I've added this
functionality to the irq number list detection method.

### DIFF
--- a/lnst/Recipes/ENRT/ConfigMixins/DevInterruptHWConfigMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/DevInterruptHWConfigMixin.py
@@ -82,7 +82,23 @@ class DevInterruptHWConfigMixin(BaseHWConfigMixin):
 
     def _pin_dev_interrupts(self, dev, cpus):
         netns = dev.netns
-        cpu_info = netns.run("lscpu", job_level=ResultLevel.DEBUG).stdout
+        self._check_cpu_validity(netns, cpus)
+
+        intrs = self._get_dev_interrupts(dev)
+
+        for i, intr in enumerate(intrs):
+            try:
+                cpu = cpus[i % len(cpus)]
+                netns.run(
+                    "echo -n {} > /proc/irq/{}/smp_affinity_list".format(
+                        cpu, intr
+                    )
+                )
+            except ValueError:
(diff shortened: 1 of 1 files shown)


olichtne:
functional tests in beaker job `J:5649606`

Axonis:
Makes sense, if tests are successful, then Ack.

jtluka:
@olichtne I Just have question whether the interrupts are populated instantly when the device is brought up here: https://github.com/LNST-project/lnst/pull/215/files#diff-eddf59fa7373907fd9a2859d4db07f4ab3c0b3edf48f82cca3634f002b0914efR117

If not, we may need to wait some time. But if this generally works, we can merge this as is.

olichtne:
> @olichtne I Just have question whether the interrupts are populated instantly when the device is brought up here: https://github.com/LNST-project/lnst/pull/215/files#diff-eddf59fa7373907fd9a2859d4db07f4ab3c0b3edf48f82cca3634f002b0914efR117
> 
> If not, we may need to wait some time. But if this generally works, we can merge this as is.

From my experiments they do populate "immediately", I ran the basic SimpleMacsecRecipe twice for each driver - once without my patch and once with the patch. I then compared the configured irqs, in each case they were the same, except for `i40e` where the old code attempted to configure more irqs which were "unavailable" (the main reason behind this PR).

So I think this is safe. I'll schedule one more test run for the production ready tests to double check.

The functional test `J:5649606` finished ok - the test-runner tasks failed due to no baselines found as expected, however the irq configuration was successful in all cases.

jtluka:
Ack to merge.

jtluka (approved):

olichtne:
Scheduled the production test in `J:5659579`, if this passes ok I'll merge this in.

olichtne:
tests passed, merging in

Issue #227, opened 2022-01-13
Create unit test for NonzeroFlowEvaluator


Axonis:
With changes proposed in the https://github.com/LNST-project/lnst/issues/226 it would make sense to test it with unit test instead of running full LNST.

This effort could be a start of using unit tests for LNST as a whole.

We need to create the structure to reflect the lnst package.

Issue #226, opened 2022-01-13
Extend NonzeroFlowEvaluator to check each PerfInterval


Axonis:
Currently NonzeroFlowEvaluator only checks whether the metric's average is higher than 0, but this could lead to fake passes if traffic stops flowing after a moment vice versa.

NonzeroFlowEvaluator could be extended to check each PerfInterval value whether its higher than 0, thus meaning it is a continuous stream of data.

PR #228, opened 2022-01-17
Nonzero evaluator upgrade + unit tests

Axonis:
Description:

With new use-cases for LNST, selecting evaluators based on parameter is required.
To achieve this `skip_perf_evaluation` parameter has been replaced by the `perf_evaluation_strategy` that allows user to specify which evaluators are supposed to be used based on the string representation in the parameter.
Currently it only supports values of: "all", "nonzero" and "none" and there is no mapping present as it was not required in this simple use case. 

However if evaluators are ever extended this parameter should be taken into consideration and mapping string-list[EvaluatorClass] shoud be made.

NonzeroFlowEvaluator could report false positive if some traffic flowed through the link at the start and then suddenly stop or vice versa. To address this issue the evaluator does not only check the average of whole flow in its entirety but instead verifies that each interval in its duration is non-zero.

With this change to NonzeroFlowEvaluator it made sense to introduce unit tests to speed up the development process of the framework. Thus, whole new `tests/` directory structure was created that reflect the lnst package and currently contains basic test cases for NonzeroFlowEvaluator.

New dev dependency was added in the form of pytest.

Tests:

Beaker job - 6191309

Reviews: @olichtne @jtluka 

Closes: https://github.com/LNST-project/lnst/issues/227
Closes: https://github.com/LNST-project/lnst/issues/226


lnst/RecipeCommon/Perf/Evaluators/NonzeroFlowEvaluator.py
```diff
@@ -43,6 +43,12 @@ def evaluate_results(
                             )
                         )
                     elif metric.average > 0:
+                        for interval in metric:
+                            if interval.value == 0:
+                                result = False
+                                result_text.append(
+                                    "{} reported zero throughput".format(metric_name)
+                                )
```
jtluka, line 51: With this implementation it may happen that due to average not being zero, the report may contain both the "zero" and "non-zero". Maybe add an "else" statement for the for loop.

Axonis: Oh right, I was testing it only with unit tests and I did not see any output even in aforementioned case.

I will fix this.

olichtne (approved):
looks good to me

PR #229, opened 2022-02-08
Recipes.ENRT.BaseEnrtRecipe: fix BaseCPUMeasurement conditional

Axonis:
Description:

With introduction of new parameter `perf_evaluation_strategy` there is a logical bug in the conditional 
of `BaseCPUMeasurement` evaluator registration. It is being evaluated as follows in default scenario:

```
parameter is "nonzero" = 0

OR

"none" = 1

which results into 1, removing the CPU perf evaluator
```

To fix this logical bug `self.params.perf_evaluation_strategy` is checking whether its string value exists in the given list instead.

Test: J:6280045

Reviews: @olichtne @jtluka


olichtne (approved):
looks ok, merging

PR #231, opened 2022-02-15
LNST Agent

pgagne:
This is the core part of LNST's effort to make our project more inclusive. 

Most of the changes here relate to changing the name of the "LNST Slave" to "LNST Agent" and all associated references. 

Also, as we were going through code we noticed a few places where the was an opportunity to remove old, unused code or rewrite things with new modern methods. Although this isn't intended to be a major part of this effort. It was mostly done in cases where the would be a fair amount of work refactoring code, when the code wasn't actually used. So it would be better to remove it. 

In some places we decided to go with whole new names for things, to better reflect there current use. For example the`lnst.Slave.NetTestSlave:SlaveMethod` class was renamed `lnst.Agent.Agent.RemoteMethods`. If you have any suggestions for other items that you think could use better names, please suggest them.


pgagne:
I was looking to rename this method, and wanted to see if this was references the actual lnst-slave or some thing else:
https://github.com/pgagne/lnst/blob/adc4562d1e5f77ff0096a8c3df71a71aff6f6e4b/lnst/Agent/NetConfigCommon.py#L13

But then I started searching for references to the methods in this file and couldn't find any place where any of the methods are actually used (outside of dead imports or in this file itself). 

Should I just remove this whole file?

lnst/Agent/__main__.py
```diff
@@ -0,0 +1,51 @@
...
+    parser.add_argument("-m", "--no-colours", action='store_true', help="disable coloured terminal output")
+    parser.add_argument("-p", "--port", type=int, default=None, help="LNST RPC port to listen on")
+    parser.add_argument("-c", "--config", default=DEFAULT_USER_CFG, help="Config file")
+    args = parser.parse_args()
+
+    agent_config = AgentConfig()
+
+    if os.path.isfile(args.config):
+        agent_config.load_config(args.config)
+    #else:
```
jtluka, line 33: Is this commented out code needed for any future work? If not, just remove it.

jtluka:
Besides my other minor comments, this looks ok imo. Thanks Perry!

olichtne (approved):
looks ok otherwise

olichtne (approved):
looks good to me now

PR #232, opened 2022-02-28
CI tests

enhaut:
Pushes and pull requests to `master` triggers Github action,
which tries to merge upstream `master` branch and pull request's
source branch and also runs `SimpleNetworkRecipe` using
2 containers whose try to ping eachother. `runner.py` is taken from
our [wiki](https://lnst.readthedocs.io/en/latest/installation.html#creating-an-executable-helloworld-test-script).

GitHub action [does not support Fedora](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), so I had to use Ubuntu instead,  
basically the requirements are the same but some of them have quite different name.
Actions have [limits](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) but it's pretty generous so I don't think it would be problem in future.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,0 +1,63 @@
+name: CI test
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  FunctionalTest:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.6"
+
+    - name: Source branch checkout
+      uses: actions/checkout@v2
+
+    - name: Set up system requirements
+      run: |
+        sudo apt-get install podman -y
+        sudo systemctl enable --now podman.socket
+        curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3.6 -
+
+    - name: Set up Podman network requirements
+      run: |
+        sudo sysctl -w net.ipv4.ip_forward=1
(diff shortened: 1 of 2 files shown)


enhaut:
@olichtne at the end, I've removed rebasing to upstream master, github also checks the merging ability and it won't trigger the action if PR is not able to merge automatically so doing it again is not useful at all.  
I've used `action/checkout` to checkout to source branch, regards to it's [documentation](https://github.com/actions/checkout#usage) it checkouts to `HEAD` of branch that triggers the action.
  
[Here](https://github.com/enhaut/lnst/runs/5376985573?check_suite_focus=true) is log of activity triggered by pull request to my forked LNST repo.

olichtne (dismissed):
Looks good to me.

Before merging I'll let others comment too @Axonis  @pgagne @jtluka

.github/workflows/ci-test.yml
```diff
@@ -0,0 +1,63 @@
...
+        libvirt-clients \
+        bridge-utils \
+        libvirt-dev \
+        libnl-3-200 \
+        libnl-route-3-dev \
+        git \
+        libnl-3-dev
+        $HOME/.poetry/bin/poetry install
+
+    - name: Build LNST workers image
```
pgagne, line 54: Shouldn't this be `LNST agent image`

enhaut: That was just typo, it should be `LNST agents`, thanks

pgagne:
Ack looks good.

pgagne (approved):
looks good.

olichtne (approved):
reviewed by everyone, merging

PR #235, opened 2022-03-09
Add performance testing to functional tunnel recipes

Axonis:
Description:

Internal need requires us to run performance tests through VXLAN and GENEVE tunnels.
To achieve this `generate_perf_endpouints` methods has been added which triggers the whole hierarchy of performance benchmarks.

Just a note to keep in mind is that when user wants to run pure functional test he has to specify empty `perf_tests` list, instead of being null. If the list would not exist in parameters default value of all `perf_tests` would be used, which results in running all available streams to us. 

Tests: 

J:6377886

The first test-runner has `perf_tests=[]`

While second one has `perf_tests=['tcp_stream']`

Reviews: @olichtne @jtluka


jtluka:
Ack. Looks sane.

olichtne (approved):
looks ok to me, merging

PR #238, opened 2022-03-28
Implement performance analysis with Linux Perf tool

Kuba314:
Runs `perf record` on all Agents on an interrupt-receiveing CPU and on a CPU where `iperf` is running. All collected data is then copied to Controller.

- [x] LinuxPerf test module
- [x] LinuxPerf measurement
- [x] LinuxPerf measurement generator
- [x] Copy `perf-record` result files from Agents to Controller

### DIFF
--- a/lnst/Controller/Namespace.py
+++ b/lnst/Controller/Namespace.py
@@ -15,6 +15,7 @@ olichtne@redhat.com (Ondrej Lichtner)
 """
 
 import logging
+from typing import Optional
 from abc import ABCMeta
 from lnst.Controller.Job import DEFAULT_TIMEOUT
 from lnst.Devices.Device import Device
@@ -84,6 +85,17 @@ class Namespace(object):
         returns None for the init namespace
         returns a string name for any other namespace"""
         return self._name
+
+    def copy_file_to_machine(
+        self,
+        local_path: str,
+        remote_path: Optional[str] = None,
+        netns: Optional['Namespace'] = None,
+    ) -> str:
+        return self._machine.copy_file_to_machine(local_path, remote_path, netns)
+
+    def copy_file_from_machine(self, remote_path: str, local_path: str):
+        self._machine.copy_file_from_machine(remote_path, local_path)
 
     def prepare_job(self, what, fail=False, json=False, desc=None,
                     job_level=ResultLevel.DEBUG):
--- a/lnst/Controller/Namespace.py
+++ b/lnst/Controller/Namespace.py
@@ -89,10 +89,9 @@ class Namespace(object):
     def copy_file_to_machine(
         self,
         local_path: str,
-        remote_path: Optional[str] = None,
-        netns: Optional['Namespace'] = None,
+        remote_path: Optional[str] = None
     ) -> str:
-        return self._machine.copy_file_to_machine(local_path, remote_path, netns)
+        return self._machine.copy_file_to_machine(local_path, remote_path, self)
 
     def copy_file_from_machine(self, remote_path: str, local_path: str):
         self._machine.copy_file_from_machine(remote_path, local_path)
--- a/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
@@ -1,4 +1,5 @@
 from typing import Any, Optional
+from datetime import datetime
 import logging
 import signal
 import time
@@ -15,6 +16,8 @@ from lnst.Controller.Host import Host
 from lnst.Controller.Recipe import BaseRecipe
 from lnst.Controller.RecipeResults import ResultLevel
 
+def timestamp() -> str:
+    return datetime.now().strftime("%Y-%m-%d_%H:%M:%S")
 
 class LinuxPerfMeasurementResults(BaseMeasurementResults):
     _host: Host
@@ -174,7 +177,7 @@ class LinuxPerfMeasurement(BaseMeasurement):
 
             # copy agent's perf.data file to a controller
             src_filepath: str = job.result["filename"]
-            new_filename: str = f"{filename}.{self._collection_index}"
+            new_filename: str = f"{filename}.{self._collection_index}.{timestamp()}"
             dst_filepath: str = os.path.join(self._data_folder, host.hostid, new_filename)
 
             host.copy_file_from_machine(src_filepath, dst_filepath)
--- a/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Any, Optional
+from typing import Any, Optional
 import logging
 import signal
 import time
@@ -19,7 +19,7 @@ from lnst.Controller.RecipeResults import ResultLevel
 class LinuxPerfMeasurementResults(BaseMeasurementResults):
     _host: Host
     _filename: str
-    _cpus: List[int]
+    _cpus: list[int]
 
     _start_timestamp: float
     _end_timestamp: float
@@ -29,7 +29,7 @@ class LinuxPerfMeasurementResults(BaseMeasurementResults):
         measurement: "LinuxPerfMeasurement",
         host: Host,
         filename: str,
-        cpus: List[int],
+        cpus: list[int],
         start_timestamp: float,
         end_timestamp: float,
     ):
@@ -66,14 +66,14 @@ class LinuxPerfMeasurementResults(BaseMeasurementResults):
 
 class AggregatedLinuxPerfMeasurementResults(BaseMeasurementResults):
 
-    _individual_results: List[LinuxPerfMeasurementResults] = []
+    _individual_results: list[LinuxPerfMeasurementResults] = []
 
     def __init__(self, result: Optional[LinuxPerfMeasurementResults] = None):
         if result:
             self._individual_results = [result]
 
     @property
-    def individual_results(self) -> List[LinuxPerfMeasurementResults]:
+    def individual_results(self) -> list[LinuxPerfMeasurementResults]:
         return self._individual_results
 
     def add_results(self, result: LinuxPerfMeasurementResults):
@@ -83,26 +83,26 @@ class AggregatedLinuxPerfMeasurementResults(BaseMeasurementResults):
 class LinuxPerfMeasurement(BaseMeasurement):
     _MEASUREMENT_VERSION: int = 1
 
-    hosts: List[Host]
+    hosts: list[Host]
     intr_filename: str
-    intr_cpus: List[int]
+    intr_cpus: list[int]
     iperf_filename: str
-    iperf_cpus: List[int]
+    iperf_cpus: list[int]
     _start_timestamp: float
     _end_timestamp: float
     _data_folder: str
 
-    _version: Optional[Dict[str, Any]] = None
+    _version: Optional[dict[str, Any]] = None
     _collection_index: int = 0
-    _jobs: List[Job] = []
+    _jobs: list[Job] = []
 
     def __init__(
         self,
-        hosts: List[Host],
+        hosts: list[Host],
         intr_filename: str,
-        intr_cpus: List[int],
+        intr_cpus: list[int],
         iperf_filename: str,
-        iperf_cpus: List[int],
+        iperf_cpus: list[int],
         data_folder: str,
         recipe_conf: Any = None,
     ):
@@ -121,7 +121,7 @@ class LinuxPerfMeasurement(BaseMeasurement):
             os.mkdir(os.path.join(self._data_folder, host.hostid))
 
     @property
-    def version(self) -> Optional[Dict[str, Any]]:
+    def version(self) -> Optional[dict[str, Any]]:
         if self._version:
             return self._version
 
@@ -165,9 +165,9 @@ class LinuxPerfMeasurement(BaseMeasurement):
                 logging.error(f"timeout while waiting for linuxperf job to finish on host {host.hostid}")
         self._end_timestamp = time.time()
 
-    def collect_results(self) -> List[BaseMeasurementResults]:
+    def collect_results(self) -> list[BaseMeasurementResults]:
         self._collection_index += 1
-        results: List[BaseMeasurementResults] = []
+        results: list[BaseMeasurementResults] = []
         for job, (host, filename, cpus) in zip(self._jobs, self._generate_configurations()):
             if job.result is None:
                 continue
@@ -195,7 +195,7 @@ class LinuxPerfMeasurement(BaseMeasurement):
     def report_results(
         cls,
         recipe: BaseRecipe,
-        aggregated_results: List[AggregatedLinuxPerfMeasurementResults],
+        aggregated_results: list[AggregatedLinuxPerfMeasurementResults],
     ):
         for aggregated_result in aggregated_results:
             cpus: str = ",".join(map(str, aggregated_result.individual_results[0].cpus))
@@ -212,13 +212,13 @@ class LinuxPerfMeasurement(BaseMeasurement):
     @classmethod
     def aggregate_results(
         cls,
-        old: Optional[List[AggregatedLinuxPerfMeasurementResults]],
-        new: List[LinuxPerfMeasurementResults],
-    ) -> List[AggregatedLinuxPerfMeasurementResults]:
+        old: Optional[list[AggregatedLinuxPerfMeasurementResults]],
+        new: list[LinuxPerfMeasurementResults],
+    ) -> list[AggregatedLinuxPerfMeasurementResults]:
         if old is None:
             return [AggregatedLinuxPerfMeasurementResults(result) for result in new]
 
-        aggregated: List[AggregatedLinuxPerfMeasurementResults] = []
+        aggregated: list[AggregatedLinuxPerfMeasurementResults] = []
         for old_measurements, new_measurement in zip(old, new):
             aggregated.append(
                 cls._aggregate_linux_perf_results(old_measurements, new_measurement)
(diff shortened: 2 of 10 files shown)


lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,241 @@
...
+    def collect_results(self) -> List[BaseMeasurementResults]:
+        self._collection_index += 1
+        results: List[BaseMeasurementResults] = []
+        for cfg in self.configurations:
+            if cfg.job is None or cfg.job.result is None:
+                continue
+
+            # copy agent's perf.data file to a controller
+            perf_folder_path: str = os.path.abspath("linuxperf-data")
+            src_filepath: str = os.path.abspath(cfg.job.result["filename"])
```
olichtne, line 179: careful about the use of `os.path.abspath` here as this is running in the **Controller** process, while the `cfg.job.result["filename"]` was generated in the **Agent** process context which may be running in a different working directory, imo this call should be moved to the test module itself.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,241 @@
...
+)
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
+from lnst.Controller.Job import Job
+from lnst.Controller.Host import Host
+from lnst.Controller.Recipe import BaseRecipe
+from lnst.Controller.RecipeResults import ResultLevel
+from lnst.RecipeCommon.Perf.Results import LinuxPerfResult
+
+
+class LinuxPerfMeasurementResults(BaseMeasurementResults, LinuxPerfResult):
```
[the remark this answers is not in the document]
Kuba314: I'm not sure about this either. `PerfRecipe` expects the Results object to have properties `start_timestamp` and `end_timestamp` and a `time_slice()` method. So I'd expect that every results object should inherit from the new class `BasePerfResult` which acts like an interface.
I don't think storing `LinuxPerfResult` as an attribute of `LinuxPerfMeasurementResults` and forwarding calls to these 3 "methods" to it is the correct OO way. But then, I don't know the class hierarchy that well...

lnst/Controller/Namespace.py
```diff
@@ -85,6 +86,17 @@ def name(self):
         returns a string name for any other namespace"""
         return self._name
 
+    def copy_file_to_machine(
+        self,
+        local_path: str,
+        remote_path: Optional[str] = None,
+        netns: Optional['Namespace'] = None,
+    ):
+        self._machine.copy_file_from_machine(local_path, remote_path, netns)
```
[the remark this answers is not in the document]
Kuba314: yes, a copy-paste method indeed. also made it return the remote path

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: yes, this was a mistake. I already called `os.path.abspath` in the Agent, but threw it in once more here for some reason. It's obviously not needed nor wanted here

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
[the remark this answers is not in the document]
Kuba314: Will forward the three methods and only use composition.

lnst/Tests/LinuxPerf.py
```diff
@@ -0,0 +1,62 @@
...
+        self.wait_for_interrupt()
+
+        stdout, stderr = process.communicate()
+
+        if stdout:
+            log_output(logging.debug, "Stdout", stdout.decode())
+        if stderr:
+            log_output(logging.debug, "Stderr", stderr.decode())
+
+        self._res_data["filename"] = os.path.abspath(self._parse_filename_from_output(stderr.decode()))
```
jtluka, line 40: Could you please explain why we need this when the `output_file` is a mandatory parameter? Can't we use it instead?

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,234 @@
...
+            self._jobs.append(host.run(LinuxPerf(output_file=filename, cpus=cpus), bg=True))
+
+    def finish(self) -> None:
+        for job, (host, _, _) in zip(self._jobs, self._generate_configurations()):
+            job.kill(signal=signal.SIGINT)
+            if not job.wait(timeout=120):
+                logging.error(f"timeout while waiting for linuxperf job to finish on host {host.hostid}")
+        self._end_timestamp = time.time()
+
+    def collect_results(self) -> List[BaseMeasurementResults]:
```
jtluka, line 168: I think this should return a list of `LinuxPerfMeasurementResults`.

lnst/Tests/LinuxPerf.py
Kuba314: I just now realized, that I'll still need to pass the filename through `_res_data["fileame"]`, because when copying the result file from agent to controller, we'd rather have an absolute path, just to be sure. Don't know if agent executes rpc calls from the same directory.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: Since this method implements `BaseMeasurement`'s method `collect_results()`, I wanted to keep the type the same, because it essentially implements the interface and so the return values should only be used as `BaseMeasurement`.

lnst/Tests/LinuxPerf.py
Kuba314: I removed `--timestamp-filename` and use `self.parsed.output_file` directly.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,240 @@
...
+        if version_job.passed:
+            match = re.match(r"perf version (.+?)", version_job.stdout)
+            if match:
+                return match.group(1)
+        return None
+
+    def _generate_configurations(self):
+        for host in self.hosts:
+            yield (host, self.intr_filename, self.intr_cpus)
+            yield (host, self.iperf_filename, self.iperf_cpus)
```
jtluka, line 160: Since this is supposed to be used as a "common" code for recipes not necessarilly coming from ENRT recipe set, I think the concept of dev_intr_cpu and perf_tool_cpu should be unified and instead have a list of cpus to be monitored. With that we will not need to include 'intr' or 'perf' substrings in the generated files, only the CPU id. This can be done as a followup pull request, just adding a note here.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def start(self) -> None:
+        self._start_timestamp = time.time()
+        for host, filename, cpus in self._generate_configurations():
+            self._jobs.append(host.run(LinuxPerf(output_file=filename, cpus=cpus), bg=True))
+
+    def finish(self) -> None:
+        for job, (host, _, _) in zip(self._jobs, self._generate_configurations()):
```
jtluka, line 168: I'm a bit concerned about this and another call in `collect_results` of `_generate_configurations()`. Can't we instead save the job:configuration in the first call of that method on line 165 and reuse this here? Is there a reason why to do it this way? I understand that the functionality is the same, but it just looks weird to generate them again and again.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: I thought it mattered on what CPU the data was captured? What would someone had to do in order to distinguish between the output files, if they were named after the CPU they ran on?

But I agree, I could rework the `LinuxPerfMeasurement` class to accept a list of CPUs.

jtluka: Of course, we want to distinguish the perf.data for individual cpus. The thing is that if an lnst user creates a recipe that DOES NOT use the `BaseEnrtRecipe` and only the `RecipeCommon` (which is independent of ENRT), there are no `dev_intr_cpu` and `perf_tool_cpu` parameters. So, to abstract this, we can substitute this with e.g. cpu0 and cpu5. The ENRT developer knows that cpu0 was used as dev_intr_cpu and cpu5 was used as perf_tool_cpu. The non-ENRT developer can use this differently.

So, you would have filenames:
```
cpu0.data.1.2022-04-23_14:28:32
cpu5.data.1.2022-04-23_14:28:32
```
instead of:
```
intr.data.1.2022-04-23_14:28:32
iperf.data.1.2022-04-23_14:28:32
```

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
jtluka: Yeah, I think the property would look better.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
jtluka: Hmm, I looked at the code again and actually your argument is kind of valid. Still I think we should do the abstraction and remove the dev_intr_cpu and perf_tool_cpu relation from the code.

To elaborate more on my idea:
* the `linuxperf_cpus`
  * will substitute the `LinuxPerfMeasurement.intr_cpus` and `LinuxPerfMeasurement.iperf_cpus`
  * will be a list of lists of cpus, e.g.
    * `linuxperf_cpus = [ [0, 5] ]` ... would run one perf-record for both cpu0 and cpu5
    * `linuxperf_cpus = [ [0], [5] ]` ... would run two perf records individually for cpu0 and cpu5
    * `linuxperf_cpus = [ [0,1,2,3] [4,5,6,7] ]` ... would run two perf records individually for the lists of cpus

The filenames will be generated based on linuxperf_cpus items:
* `linuxperf_cpus = [ [0], [5] ]`
  * cpus.0.data...
  * cpus.5.data...
* `linuxperf_cpus = [ [0,1,2,3] [4,5,6,7] ]`
  * cpus.0.1.2.3.data...
  * cpus.4.5.6.7.data...

So the `generate_configurations` code would look like this:
```
    for host in self.hosts:
        for cpu_list in self.linuxperf_cpus:
            yield (host, cpu_list)
```

@olichtne does this make sense?

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: resolved in a00d00d

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
olichtne: yes, i like this, additionally the ENRT recipe integration of this could then be:

```
@property
def linuxperf_cpus(self):
    return [self.dev_intr_cpus, self.perf_tool_cpus]
```

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -0,0 +1,240 @@
...
+        old: Optional[list[AggregatedLinuxPerfMeasurementResults]],
+        new: list[LinuxPerfMeasurementResults],
+    ) -> list[AggregatedLinuxPerfMeasurementResults]:
+        if old is None:
+            return [AggregatedLinuxPerfMeasurementResults(result) for result in new]
+
+        aggregated: list[AggregatedLinuxPerfMeasurementResults] = []
+        for old_measurements, new_measurement in zip(old, new):
+            aggregated.append(
+                cls._aggregate_linux_perf_results(old_measurements, new_measurement)
```
jtluka, line 231: I think the code between lines 226-230 can be simplified:
```
for old_measurements, new_measurement in zip(old, new):
    old_measurements.add_results(new)
```

And that's it. You don't need `_aggregate_linux_perf_results()` at all, imo.

If I'm missing something please let me know. Here's my code flow analysis:
```
perf_test()
    perf_test_iteration()
        """
        for measurement in recipe_conf.measurements:
            measurement.start()
        for measurement in reversed(recipe_conf.measurements):
            measurement.finish()
        for measurement in recipe_conf.measurements:
            measurement_results = measurement.collect_results()
            results.add_measurement_results(
                measurement, measurement_results
        """

        measurement_results = measurement.collect_results()
            -> measurement_results is a list of LinuxPerfMeasurementResults
            [
                # iteration 1, all hosts, all cpus
                LinuxPerfMeasurementResults(host1, cpu0),
                LinuxPerfMeasurementResults(host2, cpu0),
                LinuxPerfMeasurementResults(host1, cpu5),
                ...
            ]

        results.add_measurement_results(measurement, measurement_results)
        ->
            appends individual measurement results for the measurement, then
            aggregates:
                measurement.aggregate_results()
                    first iteration:
                        old=None, new=measurement_results
                        [
                            AggregatedLinuxPerfMeasurementResults(LinuxPerfMeasurementResults(host1, cpu0)),
                            AggregatedLinuxPerfMeasurementResults(LinuxPerfMeasurementResults(host2, cpu0)),
                            AggregatedLinuxPerfMeasurementResults(LinuxPerfMeasurementResults(host1, cpu5)),
                            ...
                        ]
                    next iteration:
                        old=previous, new=measurement_results
                        aggregated=[
                            AggregatedLinuxPerfMeasurementResults:[
                                LinuxPerfMeasurementResults(host1, cpu0) # first iteration
                                LinuxPerfMeasurementResults(host1, cpu0) # second iteration
                            ],
                            AggregatedLinuxPerfMeasurementResults:[
                                LinuxPerfMeasurementResults(host2, cpu0) # first iteration
                                LinuxPerfMeasurementResults(host2, cpu0) # second iteration
                            ],
                            AggregatedLinuxPerfMeasurementResults:[
                                LinuxPerfMeasurementResults(host1, cpu5) # first iteration
                                LinuxPerfMeasurementResults(host1, cpu5) # second iteration
                            ], ...
                        ]
```

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: @olichtne in which file should I put this property?

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
Kuba314: You're right. This was done to mimic `BaseFlowMeasurement` and `BaseCPUMeasurement`, which both use two functions. I will simplify it, since there's no logic needed in the child function.

Kuba314: resolved in 18a13800

jtluka:
Besides my other comments this look ok. Ack to merge once the pending issues are resolved.

olichtne (changes requested):

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
jtluka: You should also update the lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py where there are still filename parameters:
```python
class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):                  
    do_linuxperf_measurement = BoolParam(default=False)                         
    linuxperf_intr_fname = StrParam(default="intr.data")                        
    linuxperf_iperf_fname = StrParam(default="iperf.data")                      
```

And once you add the new property, replace the following code:
```python
        for combination in combinations:                                        
            measurement: BaseMeasurement = LinuxPerfMeasurement(                
                self.matched,                                                   
                [self.params.dev_intr_cpu, self.params.perf_tool_cpu],          
                data_folder=linuxperf_data_folder,                              
                recipe_conf=config,                                             
            )                                                                   
```

Kuba314: @olichtne you meant this, right?
```py
@property
def linuxperf_cpus(self):
    return [self.params.dev_intr_cpu, self.params.perf_tool_cpu]
```

Kuba314: @jtluka I suppose I could use the newly created `linuxperf_cpus` property instead of the `[self.params.dev_intr_cpu, self.params.perf_tool_cpu]`, right?

lnst/Controller/Namespace.py
```diff
@@ -85,6 +86,17 @@ def name(self):
         returns a string name for any other namespace"""
         return self._name
 
+    def copy_file_to_machine(
+        self,
+        local_path: str,
+        remote_path: Optional[str] = None,
+        netns: Optional['Namespace'] = None,
```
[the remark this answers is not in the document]
Kuba314: Should I change it to this then?
```py
    def copy_file_to_machine(
        self,
        local_path: str,
        remote_path: Optional[str] = None
    ) -> str:
        return self._machine.copy_file_to_machine(local_path, remote_path, self)
```

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
olichtne: > @olichtne you meant this, right?
> 
> ```python
> @property
> def linuxperf_cpus(self):
>     return [self.params.dev_intr_cpu, self.params.perf_tool_cpu]
> ```

yes

Kuba314: resolved in 49a0b807

olichtne (approved):
looks ok to me now

PR #241, opened 2022-05-12
ENRT.LinuxPerfMeasurementGenerator: fix generator return implementation

olichtne:
When 'yield' is used in python it automatically changes the
function/method into a generator. This is regardless of if the yield is
reached or not - this is decided when the function is parsed, not during
runtime.

At the same time, the 'return' statement works differently in a
generator. The return statement raises the 'StopIteration' exception
that is used by loops etc. to detect that the generation has stopped. If
'return' is used with a value, this value is provided to the exception
object:

`return val` is sort of equivalent to `raise StopIteration(val)`

However, when we write:

    for i in generator:

The `StopIteration(val)` simply exits the loop, the `val` itself isn't
assigned to `i` to start one more iteration of the loop.

Because of this the old code of the LinuxPerfMeasurementGenerator when
it simply called `return super().generate_perf_measurements_combinations`
the previous level of the collaborative mixin hierarchy stopped
iteration and didn't receive the previously generated list of
measurement combinations.

To fix this I restructured the method to avoid the use of the `return`
statement, and instead relied on the use of `yield` for the individual
previous values.

### DIFF
--- a/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
+++ b/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
@@ -29,6 +29,7 @@ class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
                 pass
 
         for combination in combinations:
+            res: list[BaseMeasurement]
             if self.params.do_linuxperf_measurement:
                 measurement: BaseMeasurement = LinuxPerfMeasurement(
                     self.matched,
--- a/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
+++ b/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
@@ -18,24 +18,26 @@ class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
     def generate_perf_measurements_combinations(self, config):
         combinations = super().generate_perf_measurements_combinations(config)
 
-        if not self.params.do_linuxperf_measurement:
-            return combinations
-
-        # create linuxperf data folder
-        linuxperf_data_folder: str = os.path.abspath(
-            os.path.join(self.current_run.log_dir, "linuxperf-data")
-        )
-        try:
-            os.mkdir(linuxperf_data_folder)
-        except FileExistsError:
-            pass
+        if self.params.do_linuxperf_measurement:
+            # create linuxperf data folder
+            linuxperf_data_folder: str = os.path.abspath(
+                os.path.join(self.current_run.log_dir, "linuxperf-data")
+            )
+            try:
+                os.mkdir(linuxperf_data_folder)
+            except FileExistsError:
+                pass
 
         for combination in combinations:
-            measurement: BaseMeasurement = LinuxPerfMeasurement(
-                self.matched,
-                self.linuxperf_cpus,
-                data_folder=linuxperf_data_folder,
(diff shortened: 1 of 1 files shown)


olichtne:
@Kuba314 feel free to also provide feedback for this as it touches code originally written by you, github isn't allowing me to add you as a reviewer...

olichtne:
more in depth explanation of the python generator yield vs return logic is in https://stackoverflow.com/a/71165082

jtluka:
The code looks ok. We should do a test run in RH labs to confirm this fixes the issue we hit there.

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
```diff
@@ -18,9 +18,6 @@ class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
     def generate_perf_measurements_combinations(self, config):
         combinations = super().generate_perf_measurements_combinations(config)
 
-        if not self.params.do_linuxperf_measurement:
-            return combinations
-
         # create linuxperf data folder
         linuxperf_data_folder: str = os.path.abspath(
```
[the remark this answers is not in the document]
olichtne: yeah, good catch, updated

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
```diff
@@ -18,24 +18,26 @@ class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
...
+                pass
 
         for combination in combinations:
-            measurement: BaseMeasurement = LinuxPerfMeasurement(
-                self.matched,
-                self.linuxperf_cpus,
-                data_folder=linuxperf_data_folder,
-                recipe_conf=config,
-            )
+            if self.params.do_linuxperf_measurement:
```
Kuba314, line 32: I'd personally add a type hint for `res` before this line, if we know that only `BaseMeasurement` objects will be present in the list.
```py
res: list[BaseMeasurement]
```
Everything else LGTM, learning something new every day.

enhaut (dismissed):

Axonis (dismissed):
Discussed this face to face, ack

olichtne:
test running in beaker job `J:6620012`

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
olichtne: added, for consistency with the rest of the method.

Kuba314: @olichtne The type hint should be placed before the code flow branches into the `if-else`. Currently the second type definition throws `Name "res" already defined.`

```py
res: list[BaseMeasurement]
if self.params.do_linuxperf_measurement:
    ...
    res = [measurement] + combination
else:
    res = combination```

olichtne: ok, i don't get these errors so I didn't notice...

olichtne:
> test running in beaker job `J:6620012`

this test passed without issues and the measurements ran correctly even with `do_linuxperf_measurement ==  False` so i think this should now be ok to merge.

@jtluka @Axonis  i'll need one of you to approve the pr for that

jtluka (approved):
Looks good now.

PR #242, opened 2022-05-12
Simple SR-IOV Network namespace OvS and Linux bridge recipe

Axonis:
Description:

Adding new SR-IOV recipe with a single VF placed in the network namespace to mimic the container behaviour. 

There are changes to Controller.Machine and Host, as well as Agent.InterfaceManager to handle the new scenario,
where we are remapping the device of the same class just with changed network namespace.

To actually remap the VF device, we cannot find it by the MAC address, as that is randomly generated, so there is a new method for Machine to locate the device by its interface name, which is consistent.

Test:

EL8 - 7284398
EL9 - 7284270

Reviews: @olichtne @jtluka

### DIFF
--- a/lnst/Controller/Host.py
+++ b/lnst/Controller/Host.py
@@ -57,16 +57,21 @@ class Host(Namespace):
         return ret
 
     def map_device(self, dev_id, how):
-        #TODO if this is supposed to be public it should be better than dict["hwaddr"]!!!!
-        hwaddr = how["hwaddr"]
-        dev = self._machine.get_dev_by_hwaddr(hwaddr)
+        if "hwaddr" in how:
+            hwaddr = how["hwaddr"]
+            dev = self._machine.get_dev_by_hwaddr(hwaddr)
+        elif "ifname" in how:
+            ifname = how["ifname"]
+            dev = self._machine.get_dev_by_ifname(ifname)
+        else:
+            raise ControllerError(f"Unknown how ({how}) mapping method.")
+
         if dev:
             self._objects[dev_id] = dev
             dev._id = dev_id
             dev._enable()
         else:
-            raise ControllerError("Device with macaddr {} not found on {}."
-                                  .format(hwaddr, self.hostid))
+            raise ControllerError(f"Device not found on {self.hostid} when searching by {how}.")
 
     def _custom_setattr(self, name, value):
         if not super(Host, self)._custom_setattr(name, value):
--- a/lnst/Controller/Host.py
+++ b/lnst/Controller/Host.py
@@ -79,6 +79,7 @@ class Host(Namespace):
                 self._machine.add_netns(value)
                 self._objects[name] = value
                 value._machine = self._machine
+                value.params = self.params
                 return True
             else:
                 return False
--- a/lnst/Controller/Machine.py
+++ b/lnst/Controller/Machine.py
@@ -279,6 +279,13 @@ class Machine(object):
                 return dev
         return None
 
+    def get_dev_by_ifname(self, ifname, netns=None):
+        ns_instance = self._get_netns_by_name(netns)
+        for dev in self._device_database[ns_instance].values():
+            if dev.name == ifname:
+                return dev
+        return None
+
     def _find_device_in_any_namespace(self, ifindex, peer_ifindex=None):
         for ns in list(self._namespaces.keys()) + [None]:
             dev = self.dev_db_get_ifindex(ifindex, ns)
--- a/lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
+++ b/lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
@@ -1,0 +1,203 @@
+import time
+
+from lnst.Common.Parameters import Param, StrParam
+from lnst.Common.IpAddress import ipaddress
+from lnst.Controller import HostReq, DeviceReq, RecipeParam
+from lnst.Controller.NetNamespace import NetNamespace
+from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
+from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
+from lnst.Recipes.ENRT.ConfigMixins.OffloadSubConfigMixin import (
+    OffloadSubConfigMixin,
+)
+from lnst.Recipes.ENRT.ConfigMixins.CommonHWSubConfigMixin import (
+    CommonHWSubConfigMixin,
+)
+from lnst.Devices import BridgeDevice
+
+
+class SRIOVNetnsBridgeRecipe(
+    CommonHWSubConfigMixin, OffloadSubConfigMixin, BaremetalEnrtRecipe
+):
+    """
+    This recipe implements Enrt testing for a SRIOV network scenario
+    with VF located in the network namespace to mimic container network.
+    PF with VF representor is plugged into the Linux bridge
+    and setups looks following.
+
+    .. code-block:: none
+
+                      +--------+
+               +------+ switch +-------+
+               |      +--------+       |
+       +-------|------+        +-------|------+
+       |    +--|--+   |        |    +--|--+   |
+    +--|----|eth0|--- |--+  +--|----|eth0|--- |--+
+    |  |    +----+    |  |  |  |    +----+    |  |
+    |  |       |      |  |  |  |       |      |  |
+    |  |vf_representor|  |  |  |vf_representor|  |
+    |  |              |  |  |  |              |  |
+    |  +-Linux bridge-+  |  |  +-Linux bridge-+  |
+    |         |          |  |         |          |
+    |    +-namespace-+   |  |    +-namespace-+   |
+    |   |    vf0     |   |  |   |    vf0     |   |
+    |   +-----------+    |  |   +-----------+    |
+    |      host1         |  |       host2        |
+    +--------------------+  +--------------------+
+
+    All sub configurations are included via Mixin classes.
+
+    The actual test machinery is implemented in the :any:`BaseEnrtRecipe` class.
+    """
+    host1 = HostReq()
+    host1.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    host2 = HostReq()
+    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    """
+    This parameter was created due to the difference between various kernel and distro
+    versions, not having consistent naming scheme of virtual function.
+    
+    Solution here is to expect deterministic VF name, which is derived from the PF name.
+    With specific kernel parameter `biosdevname=1` we can expect default suffix on 
+    VF to be created to be `_n`, where n is the index of VF created.
+    """
+    vf_suffix = StrParam(default="_0")
+
+    offload_combinations = Param(default=(
+        dict(gro="on", gso="on", tso="on", tx="on", rx="on"),
+        dict(gro="off", gso="on", tso="on", tx="on", rx="on"),
+        dict(gro="on", gso="off", tso="off", tx="on", rx="on"),
+        dict(gro="on", gso="on", tso="off", tx="off", rx="on"),
+        dict(gro="on", gso="on", tso="on", tx="on", rx="off")))
+
+    def test_wide_configuration(self):
+        """
+        Test wide configuration for this recipe involves switching the device to switchdev
+        mode, adding single VF and mapping the VF, as well as its representor.
+        New namespace is created to mimic container networking, where the VF is moved.
+        Next, VF is assigned an IPv4 and IPv6 address on both hosts.
+        Finally, new Linux bridge is created, where PF and VF representor are added, to
+        ensure correct switching of network traffic between the hosts.
+
+        host1.eth0 = 192.168.101.1/24 and fc00::1/64
+
+        host2.eth0 = 192.168.101.2/24 and fc00::2/64
+        """
+        host1, host2 = self.matched.host1, self.matched.host2
+        configuration = super().test_wide_configuration()
+        configuration.test_wide_devices = []
+
+        for i, host in enumerate([host1, host2]):
+            host.run(f"devlink dev eswitch set pci/{host.eth0.bus_info} mode switchdev")
+            time.sleep(2)
+            host.run(f"echo 1 > /sys/class/net/{host.eth0.name}/device/sriov_numvfs")
+            time.sleep(3)
+
+            vf_ifname = dict(ifname=f"{host.eth0.name}{self.params.vf_suffix}")
+            host.map_device("vf_eth0", vf_ifname)
+
+            host.newns = NetNamespace(f"lnst")
+            host.newns.vf_eth0 = host.vf_eth0
+
+            vf_representor_ifname = dict(ifname="eth0")
+            host.map_device("vf_representor_eth0", vf_representor_ifname)
+
+            host.br0 = BridgeDevice()
+            host.br0.slave_add(host.eth0)
+            host.br0.slave_add(host.vf_representor_eth0)
+
+            for dev in [host.eth0, host.newns.vf_eth0, host.vf_representor_eth0, host.br0]:
+                dev.up()
+
+            host.newns.vf_eth0.ip_add(ipaddress("192.168.101." + str(i + 1) + "/24"))
+            host.newns.vf_eth0.ip_add(ipaddress("fc00::" + str(i + 1) + "/64"))
+
+            configuration.test_wide_devices.append(host.newns.vf_eth0)
+
+        self.wait_tentative_ips(configuration.test_wide_devices)
+
+        return configuration
+
+    def generate_test_wide_description(self, config):
+        desc = super().generate_test_wide_description(config)
+        host1, host2 = self.matched.host1, self.matched.host2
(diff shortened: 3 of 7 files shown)


lnst/Controller/Machine.py
```diff
@@ -280,6 +280,12 @@ def get_dev_by_hwaddr(self, hwaddr):
                 return dev
         return None
 
+    def get_dev_by_ifname(self, ifname):
+        for dev in list(self._device_database[self._initns].values()):
```
olichtne, line 284: what about other namespaces here? Since this is the `Machine` class it is used for all of the namespaces on the specific agent so wouldn't it be more sensible to have a more universal method?

Also, I see you've used this method in an older commit, ideally this would be switched - first a commit that adds a new method, then a commit that uses this new method.

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
```diff
@@ -0,0 +1,158 @@
...
+        dict(gro="on", gso="on", tso="on", tx="on", rx="on"),
+        dict(gro="off", gso="on", tso="on", tx="on", rx="on"),
+        dict(gro="on", gso="off", tso="off", tx="on", rx="on"),
+        dict(gro="on", gso="on", tso="off", tx="off", rx="on"),
+        dict(gro="on", gso="on", tso="on", tx="on", rx="off")))
+
+    def test_wide_configuration(self):
+        """
+        Test wide configuration for this recipe involves just adding an IPv4 and
+        IPv6 address to the matched eth0 nics on both hosts.
```
olichtne, line 57: is the docstring accurate or copied from a different recipe?

I would probably add information about the vf creation and it moving into a namespace...

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def generate_test_wide_description(self, config):
+        """
+        Test wide description is extended with the configured addresses
+        """
+        desc = super().generate_test_wide_description(config)
+        desc += [
+            "Configured {}.{}.ips = {}".format(
+                dev.host.hostid, dev.name, dev.ips
+            )
```
olichtne, line 97: you can probably add the namespace name here.

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
```diff
@@ -0,0 +1,158 @@
...
+            for dev in config.test_wide_devices
+        ]
+        return desc
+
+    def test_wide_deconfiguration(self, config):
+        host1, host2 = self.matched.host1, self.matched.host2
+        for i, host in enumerate([host1, host2]):
+            host.run(f"devlink dev eswitch set pci/{host.eth0.bus_info} mode legacy")
+            host.run(f"echo 0 > /sys/class/net/{host.eth0.name}/device/sriov_numvfs")
+        ""  # overriding the parent docstring
```
olichtne, line 107: the docstring override needs to be directly after the `def ...` line imo

olichtne (changes requested):

olichtne:
overall the recipe looks ok, i think the main part of the review should be based on seeing real executions of it to check that it also works.

lnst/Controller/Machine.py
Axonis: Fixed

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
Axonis: Changed whole test wide description, fixed.

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
Axonis: Fixed.

lnst/Recipes/ENRT/SRIOVNetnsRecipe.py
Axonis: Fixed.

jtluka:
Looks ok. Please double check that all ENRT tests that us network namespaces are working correctly with the changes in this PR.

Also, please fix the commit history by squashing related parts.

Additionally please fix the commit descriptions that sometimes contain confusing grammar.

Axonis:
I have cleaned up the commit history and this MR should be ready to be merged, once agreed upon with HW enablement and/or HW NIC RDMA teams.

As I've mentioned in the description, there is probably another Recipe coming with VF bonding as well.

lnst/Controller/Host.py
```diff
@@ -57,16 +57,21 @@ def namespaces(self):
         return ret
 
     def map_device(self, dev_id, how):
-        #TODO if this is supposed to be public it should be better than dict["hwaddr"]!!!!
-        hwaddr = how["hwaddr"]
-        dev = self._machine.get_dev_by_hwaddr(hwaddr)
+        if "hwaddr" in how:
```
jtluka, line 60: Why is `how` a dictionary? Why it simply can't be a string? Do you expect the dictionary would have more data in it? If so what are the expected use cases?

A type hint for how would be useful.

lnst/Controller/Host.py
```diff
@@ -57,16 +57,21 @@ def namespaces(self):
...
+            raise ControllerError(f"Unknown how ({how}) mapping method.")
+
         if dev:
             self._objects[dev_id] = dev
             dev._id = dev_id
             dev._enable()
         else:
-            raise ControllerError("Device with macaddr {} not found on {}."
-                                  .format(hwaddr, self.hostid))
+            raise ControllerError(f"Device with how {how} not found on {self.hostid}.")
```
jtluka, line 74: That string is a bit confusing for both how methods:
```Device with how ifname not found on ...```
```Device with how hwaddr not found on ...```

I'd change this to ```Device not found on {self.hostid} when searching by {how}"```

File: lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py [hunk 7]
```diff
@@ -0,0 +1,197 @@
...
+
+    .. code-block:: none
+
+                      +--------+
+               +------+ switch +-------+
+               |      +--------+       |
+       +-------|------+        +-------|------+
+       |    +--|--+   |        |    +--|--+   |
+    +--|----|eth0|--- |--+   +--|----|eth0|--- |--+
+    |  |    +----+    |  |   |  |    +----+    |  |
```
jtluka, line 33: Indentation in the topology graph is one space off. Above eth0 on second host.

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
```diff
@@ -0,0 +1,197 @@
...
+
+            vf_ifname = dict(ifname=f"{host.eth0.name}v0")
+            host.map_device("vf_eth0", vf_ifname)
+
+            host.newns = NetNamespace(f"lnst{i + 1}")
+            host.newns.vf_eth0 = host.vf_eth0
+
+            vf_representor_ifname = dict(ifname="eth0")
+            host.map_device("vf_representor_eth0", vf_representor_ifname)
+            host.vf_representor = host.vf_representor_eth0
```
jtluka, line 95: Is this just for the purpose of shortening the name? If so, I think it is confusing. Anyone reading this would think why this is needed?

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
```diff
@@ -0,0 +1,197 @@
...
+
+        return configuration
+
+    def generate_test_wide_description(self, config):
+        """
+        Test wide description consists of mirror setup of a single namespace,
+        containing virtual function. Representor of this virutal function, as well as,
+        physical function is plugged into the OvS bridge, which has hardware offloading
+        capabilities enabled.
+        """
```
jtluka, line 119: I'm not sure if this docstring is useful in th context of this method. Probably makes sense to include this in the top level class docstring or the `test_wide_configuraiton` docstring.

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
```diff
@@ -0,0 +1,197 @@
...
+        for i, host in enumerate([host1, host2]):
+            desc += [
+                f"Configured host{i} interface {host.eth0.name} to be controlled by the switchdev driver \n"
+                "and created a single virtual function via PCI SYS interface.\n"
+                f"Created OvS bridge {host.br0} with enabled hardware offloading functionality.\n"
+                f"Configured new namespace to mimic network functionality of container, where \n"
+                f"virtual function {host.newns.vf_eth0.name} was moved and had following IP addresses \n"
+                f"assigned: {host.newns.vf_eth0.ips}.\n"
+                f"Physical function {host.eth0} interface, as well as, virtual function representor \n"
+                f"interface {host.vf_representor} were plugged into the OvS bridge {host.br0}."
```
jtluka, line 132: Not necessarily need to be resolved now, but for this descriptions try to
* use short sentences - avoid 'and'
* split this to individual configurations made, so that they're each on one line - see other ENRT recipes e.g. lnst/Recipes/ENRT/VirtualBridgeVlansOverBondRecipe.py

lnst/Controller/Host.py
jtluka: Sorry .. misunderstood the concept. Resolving.

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
Axonis: fixed

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
Axonis: Removed

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
```diff
@@ -0,0 +1,187 @@
...
+
+        host2.eth0 = 192.168.101.2/24 and fc00::2/64
+        """
+        host1, host2 = self.matched.host1, self.matched.host2
+        configuration = super().test_wide_configuration()
+        configuration.test_wide_devices = []
+
+        for i, host in enumerate([host1, host2]):
+            host.run(f"devlink dev eswitch set pci/{host.eth0.bus_info} mode switchdev; sleep 2")
+            host.run(f"echo 1 > /sys/class/net/{host.eth0.name}/device/sriov_numvfs; sleep 5")
```
jtluka, line 81: Could you move `sleep` commands out of the `run()`? My concern is that if the first and more important part of the command list fails, the sleep would mask the failed return code because sleep would return 0.

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
```diff
@@ -0,0 +1,187 @@
...
+
+            vf_ifname = dict(ifname=f"{host.eth0.name}v0")
+            host.map_device("vf_eth0", vf_ifname)
+
+            host.newns = NetNamespace(f"lnst{i + 1}")
```
jtluka, line 86: We probably don't need to index the namespaces as they're on different hosts. Not a big deal, but it would probably make it easier to read the code if the index is not used here.

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def generate_test_wide_description(self, config):
+        desc = super().generate_test_wide_description(config)
+        host1, host2 = self.matched.host1, self.matched.host2
+        for i, host in enumerate([host1, host2]):
+            desc += [
+                f"Configured {host.hostid}.{host.eth0.name}.driver = switchdev\n"
+                f"Created virtual function on {host.hostid}.{host.eth0.name} = {host.newns.vf_eth0.name}\n"
+                f"Created network_namespace on {host.hostid} = {host.newns.name}\n"
+                f"Moved interface {host.newns.vf_eth0.name} from {host.hostid} to {host.hostid}.{host.newns.name}\n"
```
jtluka, line 116: I expect this would look like:
```
Moved interface some_name from host1 to host1.lnst1
```

This is a bit confusing as it does not mention "namespace". I'd change this to:
```
Moved interface some_name from host1 root namespace to lnst1 namespace.
```

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
```diff
@@ -0,0 +1,187 @@
...
+    def test_wide_deconfiguration(self, config):
+        """
+        Test wide deconfiguration means deleting the Linux bridge and returning the
+        control over the physical function to base driver.
+        Finally virtual function is deleted.
+        """
+        host1, host2 = self.matched.host1, self.matched.host2
+        for i, host in enumerate([host1, host2]):
+            host.run(f"echo 0 > /sys/class/net/{host.eth0.name}/device/sriov_numvfs; sleep 5")
+            host.run(f"devlink dev eswitch set pci/{host.eth0.bus_info} mode legacy; sleep 2")
```
jtluka, line 135: Could you move sleep commands out of the run()? My concern is that if the first and more important part of the command list fails, the sleep would mask the failed return code because sleep would return 0.

jtluka (commented):
Overall looks good. There are just few minor issues.

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
Axonis: Fixed

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
Axonis: Fixed

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
Axonis: Fixed

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
jtluka: Looks good now, resolving

jtluka (dismissed):
Looks good now.

lnst/Recipes/ENRT/SRIOVNetnsBridgeRecipe.py
```diff
@@ -0,0 +1,195 @@
...
+
+    The actual test machinery is implemented in the :any:`BaseEnrtRecipe` class.
+    """
+    host1 = HostReq()
+    host1.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    host2 = HostReq()
+    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    vf_suffix = StrParam(default="_0")
```
olichtne, line 57: as discussed on irc it should be explained that this requires the use of specific kernel boot parameters to work as expected.

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
```diff
@@ -0,0 +1,198 @@
...
+
+    The actual test machinery is implemented in the :any:`BaseEnrtRecipe` class.
+    """
+    host1 = HostReq()
+    host1.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    host2 = HostReq()
+    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    vf_suffix = StrParam(default="_0")
```
olichtne, line 57: same comment here

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
olichtne: is this removed? i still see the `host.vf_representor = host.vf_representor_eth0` in here but i'm not sure if that was the original code that was being commented on.

olichtne (changes requested):
just minor updates requested. otherwise looks good.

lnst/Controller/Machine.py
```diff
@@ -279,6 +279,13 @@ def get_dev_by_hwaddr(self, hwaddr):
                 return dev
         return None
 
+    def get_dev_by_ifname(self, ifname, netns=None):
+        ns_instance = self._get_netns_by_name(netns)
+        for dev in list(self._device_database[ns_instance].values()):
```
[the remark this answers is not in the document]
olichtne: yeah, this is a construct that is typically auto created by the `2to3` scripts which we used a while ago, we have it in a couple of places since the migration, since it's harmless i didn't really care about removing them.

but for new code i think it's unnecessary.

lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
Axonis: I wasn't testing OvS recipe and forgot about the changes there. Going to fix it now.

lnst/Controller/Machine.py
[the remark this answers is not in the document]
Axonis: fixed

olichtne (approved):

PR #245, opened 2022-05-27
Tests.TestPMD: change -w to -a argument

olichtne:
The -w (whitelist) argument has been deprecated and removed since DPDK
version 21.11 [1] and is instead replaced by the -a (allowlist)
argument.

[1] https://doc.dpdk.org/guides/rel_notes/release_21_11.html?highlight=whitelist


jtluka (approved):
Ack to merge.

PR #249, opened 2022-06-20
External.TRexLib: move import_optionals to run method

olichtne:
The optionals are runtime optionals that can be relevant to the specific
host where the trex is to be executed, and the import depends on the
sys.path variable being updated to include the specified parametrized
path.

Therefore the `_import_optionals` method should only be called in the
run method.


enhaut (approved):
makes sense, lgtm

jtluka (commented):
Looks ok.

PR #250, opened 2022-06-21
External.TRex: fix * imports and globals export

olichtne:
from x import * is only supported on the global level so we need to
specify the actual objects to import.

We also need to export them as globals for the rest of the modules to
work correctly.


olichtne:
testing in `J:6739914`

olichtne:
testing finished ok

PR #251, opened 2022-07-14
Test improvements

enhaut:
Description:  
* refactored reporting results, boolean values has been changed to `ResultType` object which allows us to implement more suitable results. For compatibility reasons, `ResultType` implements `__bool__()` method, so these changes shouldn't be breaking.
* implemented `WARNING` result type which is evaluated as a pass (by `__bool__()`)
* changed evaluation of cpu/throughput improvements - if there is a lower cpu utilization than the baseline, it could mean kernel just skipped some part of code or something haven't been activated, that should be tested by functional tests
* i wasn't sure if we want to swap boolean values by `ResultType` object across whole project so i've added a "compatibility" check in `lnst/Controller/Recipe.py:158`
* added PR desc template 
  
Tests:  
Tested using internal RH infrastructure, test jobs (`6829485` and `6829144`) have passed, here is the sample how it looks like:
```bash
    ...
    WARNING TestResult:
        Configured 5.86% difference as acceptable
        IMPROVEMENT: CPU cpu: utilization 10.94% lower than baseline
```

Reviews:  
@olichtne @jtluka @Axonis @pgagne

### DIFF
--- a/lnst/Controller/Recipe.py
+++ b/lnst/Controller/Recipe.py
@@ -6,6 +6,7 @@ __author__ = """
 olichtne@redhat.com (Ondrej Lichtner)
 """
 
+from typing import Union
 import copy
 import datetime
 import logging
@@ -152,9 +153,9 @@ class BaseRecipe(object):
         else:
             return None
 
-    def add_result(self, success, description="", data=None,
+    def add_result(self, result: Union[ResultType, bool], description="", data=None,
                    level=None, data_level=None):
-        self.current_run.add_result(Result(success, description, data,
+        self.current_run.add_result(Result(result, description, data,
                                            level, data_level))
 
     def __getstate__(self):
@@ -181,10 +182,9 @@ class RecipeRun(object):
 
         self._results.append(result)
 
-        result_str = (
-            decorate_with_preset("PASS", "pass")
-            if result.success
-            else decorate_with_preset("FAIL", "fail")
+        result_str = decorate_with_preset(
+            str(result.result),
+            str(result.result).lower()
         )
         if len(result.description.split("\n")) == 1:
             logging.info(
--- a/lnst/Controller/Recipe.py
+++ b/lnst/Controller/Recipe.py
@@ -13,11 +13,13 @@ import logging
 import lzma
 import os
 import pickle
+from functools import reduce
+
 from lnst.Common.Parameters import Parameters, Param
 from lnst.Common.Colours import decorate_with_preset
 from lnst.Controller.Requirements import _Requirements, HostReq
 from lnst.Controller.Common import ControllerError
-from lnst.Controller.RecipeResults import BaseResult, Result
+from lnst.Controller.RecipeResults import BaseResult, Result, ResultType
 
 class RecipeError(ControllerError):
     """Exception thrown by the BaseRecipe class"""
@@ -216,7 +218,17 @@ class RecipeRun(object):
 
     @property
     def overall_result(self):
-        return all([i.success for i in self.results] + [self.exception is None])
+        results = [i.result for i in self.results]
+
+        if (ResultType.FAIL in results) or (self.exception is not None):
+            return ResultType.FAIL
+
+        if ResultType.WARNING in results:
+            return ResultType.WARNING
+        elif reduce(ResultType.max_severity, results) == ResultType.PASS:
+            return ResultType.PASS
+        else:
+            raise NotImplementedError(f"Unsupported result type(s): {results}")
 
     @property
     def recipe(self) -> BaseRecipe:
--- a/lnst/Controller/RecipeResults.py
+++ b/lnst/Controller/RecipeResults.py
@@ -13,6 +13,7 @@ __author__ = """
 olichtne@redhat.com (Ondrej Lichtner)
 """
 
+import logging
 import time
 from enum import IntEnum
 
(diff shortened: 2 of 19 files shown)


enhaut:
Fixed failing due to `sd-bus: Connection reset by peer` error. It seems like github-wide issue with ubuntu - https://github.com/actions/runner/issues/1994

lnst/RecipeCommon/Perf/Evaluators/BaselineFlowAverageEvaluator.py
```diff
@@ -76,13 +76,19 @@ def _average_diff_comparison(
...
-        comparison = abs(difference) <= self._pass_difference
-        if comparison:  # TODO
+
+        cpu = "_cpu_" in name
+
+        #  (                 flow metrics                 ) or (                cpu metrics                )
+        if (not cpu and difference > self._pass_difference) or (cpu and difference < -self._pass_difference):
+            comparison = ResultType.WARNING
+            result_text = "IMPROVEMENT: " + result_text
+        elif (not cpu and difference >= -self._pass_difference) or (cpu and difference <= self._pass_difference):
```
olichtne, line 91: looking at the logs from the test jobs, i'm thinking if maybe it would make sense to add "PASS:" and "FAIL:" to the individual comparison lines same as for "IMPROVEMENT:" too?

e.g.
```
        Configured 4.34% difference as acceptable
        PASS: New generator_results average is 1.65% lower from the baseline generator_results
        PASS: New generator_cpu_stats average is 0.02% higher from the baseline generator_cpu_stats
        FAIL: New receiver_results average is 4.36% lower from the baseline receiver_results
        PASS: New receiver_cpu_stats average is 3.47% lower from the baseline receiver_cpu_stats
```

what do you think @enhaut @jtluka

olichtne (commented):
I also have a generic comment to how you write the commit header lines. Please use the format of:

`file: topic of change`

where file explains the location of the change... in some generic sense, if the change touches multiple files there's typically a file or at least a python package that is the "main" source of the change.

lnst/Controller/Recipe.py
```diff
@@ -152,9 +153,13 @@ def current_run(self):
         else:
             return None
 
-    def add_result(self, success, description="", data=None,
+    def add_result(self, result: Union[ResultType, bool], description="", data=None,
                    level=None, data_level=None):
-        self.current_run.add_result(Result(success, description, data,
+        if not isinstance(result, ResultType):  # backwards compatibility
+            result = ResultType(result)
+            logging.debug("Please use ResultType for job results instead of boolean value")
```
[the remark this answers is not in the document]
enhaut: 1. i've replaced it but still, want to keep backwards compatibility, just in case
2. it would, thanks.

lnst/Controller/RunSummaryFormatter.py
```diff
@@ -30,15 +30,14 @@ def __init__(self, level=ResultLevel.IMPORTANT, colourize=False):
...
-    def _format_success(self, success):
-        if success:
-            return (decorate_with_preset("PASS", "pass")
-                    if self._colourize
-                    else "PASS")
-        else:
-            return (decorate_with_preset("FAIL", "fail")
-                    if self._colourize
-                    else "FAIL")
+    def _format_success(self, res):
```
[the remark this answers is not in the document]
enhaut: done, thanks

lnst/Controller/RecipeResults.py
```diff
@@ -21,25 +21,51 @@ class ResultLevel(IntEnum):
...
+    PASS = 1
+
+    def __str__(self):
+        if self == ResultType.FAIL:
+            return "FAIL"
+        else:
+            return "PASS"
+
+    @staticmethod
+    def most_important(first, second):
```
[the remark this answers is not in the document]
enhaut: renamed to `max_severity()`

lnst/RecipeCommon/Perf/Evaluators/BaselineFlowAverageEvaluator.py
enhaut: nice idea, thanks

lnst/Controller/RecipeResults.py
```diff
@@ -21,25 +21,51 @@ class ResultLevel(IntEnum):
...
+        self._result = result
 
     @property
     def timestamp(self):
         return self._timestamp
 
     @property
     def success(self):
-        return self._success
+        return self._result in (ResultType.PASS, ResultType.WARNING)
```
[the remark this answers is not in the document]
enhaut: thanks

lnst/Controller/RunSummaryFormatter.py
```diff
@@ -118,6 +117,6 @@ def format_run(self, run):
                 output_lines.extend(self._format_data(res.data))
 
         output_lines.append("Overall result of this Run: {}".
-                            format(self._format_success(run.overall_result)))
+                            format(run.overall_result))
```
olichtne, line 120: why remove the formatting/colorization here?

lnst/Controller/Recipe.py
```diff
@@ -216,7 +216,17 @@ def results(self):
...
-        return all([i.success for i in self.results] + [self.exception is None])
+        results = [i.result for i in self.results]
+
+        if (ResultType.FAIL in results) or (self.exception is not None):
+            return ResultType.FAIL
+
+        if ResultType.WARNING in results:
+            return ResultType.WARNING
+        elif all([i == ResultType.PASS for i in results]):
+            return ResultType.PASS
```
olichtne, line 227: i think this can all be simplified by using the `max_severity` class method and https://docs.python.org/3/library/functools.html?highlight=reduce#functools.reduce

olichtne (changes requested):
looks almost ready, just minor comments

lnst/Controller/RecipeResults.py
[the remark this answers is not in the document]
enhaut: changed

lnst/RecipeCommon/Perf/Evaluators/BaselineFlowAverageEvaluator.py
enhaut: i've replaces `success` property with `result`. now it should be consistent across project

lnst/Controller/Recipe.py
enhaut: so i've replaced property `succes` with `result` as it could also return `WARNING` not just pass/fail

lnst/Controller/RunSummaryFormatter.py
enhaut: sure

olichtne (approved):
looks ok to me now

PR #256, opened 2022-10-06
Recipes/BaremetalEnrtRecipe: avoid duplication of perf profiled cpus

### LINKED ISSUE #255: BaremetalEnrtRecipe should aggregate the cpus in linuxperf_cpus property
Currently the BaremetalEnrtRecipe does following:
```
    @property
    def linuxperf_cpus(self):
        return [self.params.dev_intr_cpu, self.params.perf_tool_cpu]
```

This can be an issue when the `dev_intr_cpu` and `perf_tool_cpu` contain the same cpus. Moreover, since both these are lists, I'm not quite sure if the code does what is intended. For example, if `dev_intr_cpu=[0]` and `perf_tool_cpu=[5]`, then `linuxperf_cpus=[ [0], [5] ]`. I'd expect this should be `linuxperf_cpus=[0,5]`.

jtluka:
Description:  

If dev_intr_cpu and perf_tool_cpu contain the same cpus the LinuxPerf measurement would run perf profiler on these cpus twice, also writing to the same file at the same time, corrupting the data.

The two lists need to be aggregated and remove duplicates to avoid the conflict.

Fixes #255
  
Tests:  

RH internal job ids: ~~J:7083780~~ ~~J:7083775~~ J:7108529 (no override) J:7108533 (override)

Reviews:  

@olichtne @Kuba314 

Closes: #255

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
@@ -145,19 +145,19 @@ class LinuxPerfMeasurement(BaseMeasurement):
         return None
 
     @property
-    def configurations(self) -> Iterable[tuple[Host, list[int]]]:
+    def configurations(self) -> Iterable[tuple[Host, int, list[int]]]:
         for host in self.hosts:
-            for cpu_list in self.linuxperf_cpus:
-                yield (host, cpu_list)
+            for cpu_list_id, cpu_list in enumerate(self.linuxperf_cpus):
+                yield (host, cpu_list_id, cpu_list)
 
     def start(self) -> None:
         self._start_timestamp = time.time()
-        for host, cpus in self.configurations:
-            filename: str = f"cpus.{'.'.join(map(str, cpus))}.data"
+        for host, cpu_list_id, cpus in self.configurations:
+            filename: str = f"cpu_list{cpu_list_id}.cpus.{'.'.join(map(str, cpus))}.data"
             self._jobs.append(host.run(LinuxPerf(output_file=filename, cpus=cpus), bg=True))
 
     def finish(self) -> None:
-        for job, (host, _) in zip(self._jobs, self.configurations):
+        for job, (host, _, _) in zip(self._jobs, self.configurations):
             job.kill(signal=signal.SIGINT)
             if not job.wait(timeout=120):
                 logging.error(f"timeout while waiting for linuxperf job to finish on host {host.hostid}")
@@ -166,7 +166,7 @@ class LinuxPerfMeasurement(BaseMeasurement):
     def collect_results(self) -> list[BaseMeasurementResults]:
         self._collection_index += 1
         results: list[BaseMeasurementResults] = []
-        for job, (host, cpus) in zip(self._jobs, self.configurations):
+        for job, (host, _, cpus) in zip(self._jobs, self.configurations):
             if job.result is None:
                 continue
 
--- a/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
+++ b/lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
@@ -7,12 +7,19 @@ from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import (
 from lnst.RecipeCommon.Perf.Measurements.LinuxPerfMeasurement import (
     LinuxPerfMeasurement,
 )
-from lnst.Common.Parameters import BoolParam
+from lnst.Common.Parameters import (
+    BoolParam,
+    IntParam,
+    ListParam,
+)
+from lnst.Common.LnstError import LnstError
 
 import os
 
+
 class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
     do_linuxperf_measurement = BoolParam(default=False)
+    linuxperf_cpus_override = ListParam(type=ListParam(type=IntParam()), mandatory=False)
 
     def generate_perf_measurements_combinations(self, config):
         combinations = super().generate_perf_measurements_combinations(config)
@@ -27,12 +34,27 @@ class LinuxPerfMeasurementGenerator(BaseMeasurementGenerator):
             except FileExistsError:
                 pass
 
+        profiled_cpu_groups: list[list[int]] = []
+        if self.params.get("linuxperf_cpus_override"):
+            profiled_cpu_groups = self.params.linuxperf_cpus_override
+        else:
+            profiled_cpu_groups = getattr(self, "linuxperf_cpus", [])
+
+        # TODO: in case no group of cpus is in the list, we may simply run
+        # profiler without the cpu specified, however this requires additional
+        # changes in the LinuxPerfMeasurement class; for now, let's just
+        # disallow this by raising an exception
+        if self.params.do_linuxperf_measurement and not profiled_cpu_groups:
+            raise LnstError(
+                "Cannot profile empty list of cpus when do_linuxperf_measurement parameter is specified"
+            )
+
         for combination in combinations:
             res: list[BaseMeasurement]
             if self.params.do_linuxperf_measurement:
                 measurement: BaseMeasurement = LinuxPerfMeasurement(
                     self.matched,
-                    self.linuxperf_cpus,
+                    profiled_cpu_groups,
                     data_folder=linuxperf_data_folder,
                     recipe_conf=config,
                 )


Kuba314 (dismissed):
LGTM

jtluka:
> I believe that this comes from the fact that the original implementation expected following format of the linuxperf_cpus: `[ [1,2,3], [4,5,6]]. However this does not make sense too much, it should be just list(int).

Hmm, maybe it does make sense.

The LinuxPerfMeasurement may profile various sets of CPUs at the same time. So the fix could be changing the linuxperf_cpus property to `[ list(set(dev_intr_cpu + perf_tool_cpu)) ]`

jtluka:
Although this would fix the issue, it's not the intended use.

We want to be able to profile both the cpus used for interrupts as well as cpus used for iperf independently. So, we'll likely need to add some logic on creating the sets.

For the simplest case, when we ping IRQs to one cpu and iperf to another cpu, we just need to check if they're equal:
* dev_intr_cpu=[0] perf_tool_cpu=[5] -> linuxperf_cpus=[ [0], [5] ]
* dev_intr_cpu=[5] perf_tool_cpu=[5] -> linuxperf_cpus=[ [5] ]

For more complex scenarios, like parallel iperf testing, this would get complicated:
* dev_intr_cpu=[0,1,2,3] perf_tool_cpu=[0,1,2,3] -> linuxperf_cpus=[ [0,1,2,3] ]
* dev_intr_cpu=[0,1,2,3] perf_tool_cpu=[4,5,6,7] -> linuxperf_cpus=[ [0,1,2,3], [4,5,6,7 ]

So maybe just check that if the lists are same, then just use only one, otherwise use both?

Kuba314:
@jtluka why don't we just do `set([tuple(dev_intr_cpu), tuple(perf_tool_cpu)])`? This will not work correctly only if both sets of CPUs intersect only in *some* cpus. What should we do then? Either we decide to run `perf` twice on the intersected CPUs, or we run `perf` only once, but then the results would be only in one of the two files.

jtluka:
@Kuba314 that's interesting idea. I've checked this with two common scenarios and it works perfectly. Thanks for suggestion.

olichtne:
wouldn't it be easier to keep the `linuxperf_cpus=[[1, 2], [1, 2]]` logic (so allow duplicated cpu sets) but change the way we generate `LinuxPerf` classes so that each instance writes to a different file that doesn't create name conflicts?

this way you don't have to come up with a way of "what duplicate detection" method works exactly how we want it but instead accept that you may create two captures that are basically the same but since they don't conflict with each other then it shouldn't be a problem?

jtluka:
> wouldn't it be easier to keep the linuxperf_cpus=[[1, 2], [1, 2]] logic (so allow duplicated cpu sets) but change the way we generate LinuxPerf classes so that each instance writes to a different file that doesn't create name conflicts?

I'm a bit worried about additional performance hit. Of course, if we're profiliing the performance measurements are very likely to be off, but not sure if this would skew the measurements more.

On the other hand, your proposal is more generic and "safe" in a sense that we don't introduce additional issues that we would have to fix later.

jtluka:
@olichtne @Kuba314 I've force pushed another implementation of the fix for this issue, please review.

jtluka:
I also had an idea, that in some cases it would be useful to override this with a recipe parameter. That way the user can simply specify `linuxperf_cpus=[ [0,1], [3,4] ]`. So, the value would not be determined on what the `dev_intr_cpu`/`perf_tool_cpu` values are.

olichtne:
> override this with a recipe parameter

yeah i think that can make sense in situations where you may want to profile all cpus for example. i think there may be a slight problem with this in case the matched machines don't have an equal number of cpu cores.

jtluka:
I've pushed additional changes that makes possible to override the profiled cpus with a recipe parameter.

Here's a summary:
* to avoid overwriting files generated by perf profiler, the filenames include the index of the cpu group in the `linuxperf_cpus` cpu group list
* the `BaremetalEnrtRecipe.linuxperf_cpus` property remains unchanged
* a new `linuxperf_cpus_override` recipe parameter is provided by `LinuxPerfMeasurementGenerator`, that can be used to override the `linuxperf_cpus` property defined by a recipe derived from `BaremetalEnrtRecipe`
  * this can be used also when the recipe does not have the `linuxperf_cpus` defined
* there's a TODO included where it makes sense to do perf profiliing without any cpus specified, however that would require addtional changes in LinuxPerfMeasurement - to avoid snowball effect, I'll create a separate issue for this - for now, if `do_linuxperf_measurement=True` and set of cpus to be profiled turns out to be empty, that would raise an Exception

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
```diff
@@ -27,12 +34,30 @@ def generate_perf_measurements_combinations(self, config):
...
+            except AttributeError:
+                profiled_cpu_groups = []
+
+        # TODO: in case no group of cpus is in the list, we may simply run
+        # profiler without the cpu specified, however this requires additional
+        # changes in the LinuxPerfMeasurement class; for now, let's just
+        # disallow this by raising an exception
+        if not profiled_cpu_groups:
+            raise LnstError(
+                "Cannot profile empty list of cpus when do_linuxperf_measurement parameter is specified"
```
jtluka, line 52: This is misleading. The `do_linuxperf_measurement` may be both True and False.

jtluka:
New test jobs J:7108529 (no override) J:7108533 (override)

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
```diff
@@ -27,12 +34,30 @@ def generate_perf_measurements_combinations(self, config):
...
                 pass
 
+        profiled_cpu_groups: list(list(int)) = []
+        if self.params.get("linuxperf_cpus_override", None):
+            profiled_cpu_groups = self.params.linuxperf_cpus_override
+        else:
+            try:
+                profiled_cpu_groups = self.linuxperf_cpus
+            except AttributeError:
+                profiled_cpu_groups = []
```
Kuba314, line 44: You can replace this with `getattr`:
```suggestion
            profiled_cpu_groups = getattr(self, "linuxperf_cpus", [])
```

Kuba314 (changes requested):

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -145,19 +145,19 @@ def _get_host_perf_version(self) -> Optional[str]:
         return None
 
     @property
-    def configurations(self) -> Iterable[tuple[Host, list[int]]]:
+    def configurations(self) -> Iterable[tuple[int, Host, list[int]]]:
         for host in self.hosts:
-            for cpu_list in self.linuxperf_cpus:
-                yield (host, cpu_list)
+            for cpu_list_id, cpu_list in enumerate(self.linuxperf_cpus):
+                yield (cpu_list_id, host, cpu_list)
```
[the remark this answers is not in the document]
jtluka: Ok, will change that.

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
```diff
@@ -27,12 +34,30 @@ def generate_perf_measurements_combinations(self, config):
             except FileExistsError:
                 pass
 
+        profiled_cpu_groups: list(list(int)) = []
```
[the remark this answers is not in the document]
jtluka: Good catch. Will fix that, thanks.

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
jtluka: Yeah makes sense.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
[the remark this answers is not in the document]
jtluka: Fixed.

jtluka:
New test jobs: J:7111484 J:7111482

lnst/Recipes/ENRT/MeasurementGenerators/LinuxPerfMeasurementGenerator.py
jtluka: Fixed.

jtluka:
> New test jobs: J:7111484 J:7111482

The jobs completed and the new file names were used correctly both with and without override.

olichtne (approved):
looks good to me.

PR #258, opened 2022-10-12
RecipeCommon.Perf.BaselineEvaluator: add data to result object

olichtne:
Description:  

Having access to the evaluation data as part of the created final Result object can be very helpful for post run analysis and is difficult to find without this direct link.

As the data is already present somewhere in the lrc file this shouldn't add any overhead except a simple object link to one additional place.

I am also considering defining a custom `class BaselineComparisonResult(Result)` which would be used here instead which could make the filtering in analysis tools even easier because we could simply do:

`comparison_results = [i for i in lrc.results if isinstance(i, BaselineComparisonResult)]`

What do you guys think?

I've had this prepared for quite a while but I think I really require this now to implement the inclusion of analysis data into dashboard - some of the data fields that we want to send are difficult to retrieve without this.
  
Tests:  

beaker job: `J:7102548`

Reviews: 

@jtluka @enhaut @Axonis 

Closes: #

### DIFF
--- a/lnst/RecipeCommon/Perf/Evaluators/BaselineCPUAverageEvaluator.py
+++ b/lnst/RecipeCommon/Perf/Evaluators/BaselineCPUAverageEvaluator.py
@@ -1,5 +1,5 @@
 from __future__ import division
-from typing import List, Tuple, Dict
+from typing import List, Dict
 
 
 from lnst.Controller.Recipe import BaseRecipe
@@ -11,7 +11,7 @@ from lnst.RecipeCommon.Perf.Measurements.Results import (
     BaseMeasurementResults as PerfMeasurementResults,
 )
 from lnst.RecipeCommon.Perf.Evaluators.BaselineEvaluator import (
-    BaselineEvaluator,
+    BaselineEvaluator, MetricComparison
 )
 
 
@@ -77,28 +77,30 @@ class BaselineCPUAverageEvaluator(BaselineEvaluator):
         result: PerfMeasurementResults,
         baseline: PerfMeasurementResults,
         result_index: int = 0
-    ) -> Tuple[ResultType, List[str]]:
+    ) -> List[MetricComparison]:
         comparison = ResultType.FAIL
-        text = []
 
-        host = result.host.hostid
-        metric = f"{result_index}_{host}"
+        metric_name = f"{result_index}_utilization"
 
         if baseline is None:
-            comparison = ResultType.FAIL
-            text.append(
-                "CPU {cpuid}: no baseline found".format(cpuid=result.cpu)
-            )
-        elif (threshold := self._thresholds.get(metric, None)) is not None:
+            return [
+                MetricComparison(
+                    metric_name=metric_name,
+                    result=ResultType.FAIL,
+                    text=f"FAIL: CPU {result.cpu}: no baseline found",
+                )
+            ]
+        elif (threshold := self._thresholds.get(metric_name, None)) is not None:
             try:
                 difference = result_averages_difference(
                     result.utilization, baseline.utilization
                 )
 
-                text.append(
-                    "CPU {cpuid}: utilization {diff:.2f}% {direction} than baseline. "
+                text = (
+                    "CPU {cpuid}: {metric_name} {diff:.2f}% {direction} than baseline. "
                     "Allowed difference: {threshold}%".format(
                         cpuid=result.cpu,
+                        metric_name=metric_name,
                         diff=abs(difference),
                         direction="higher" if difference >= 0 else "lower",
                         threshold=threshold
@@ -107,20 +109,27 @@ class BaselineCPUAverageEvaluator(BaselineEvaluator):
 
                 if difference < -threshold:
                     comparison = ResultType.WARNING
+                    text = "IMPROVEMENT: " + text
                 elif difference <= threshold:
                     comparison = ResultType.PASS
+                    text = "PASS: " + text
                 else:
                     comparison = ResultType.FAIL
-
-                text[-1] = ("IMPROVEMENT: " if comparison == ResultType.WARNING else f"{comparison}: ") + text[-1]
+                    text = "FAIL: " + text
             except ZeroDivisionError:
-                text.append(
-                    "CPU {cpuid}: zero division by baseline".format(
-                        cpuid=result.cpu
-                    )
+                text = f"CPU {result.cpu}: {metric_name} zero division by baseline"
+            return [
+                MetricComparison(
+                    metric_name=metric_name,
+                    result=comparison,
+                    text=text,
                 )
+            ]
         else:
-            comparison = ResultType.FAIL
-            text.append(f"Metric {metric}, threshold not found")
-
-        return comparison, text
+            return [
+                MetricComparison(
+                    metric_name=metric_name,
+                    result=ResultType.FAIL,
+                    text=f"FAIL: CPU {result.cpu}: {metric_name} no threshold found",
+                )
+            ]
--- a/lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
+++ b/lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
@@ -1,4 +1,7 @@
-from typing import List, Tuple
+from typing import List
+from functools import reduce
+from dataclasses import dataclass
+
 from lnst.Controller.Recipe import BaseRecipe
 from lnst.Controller.RecipeResults import ResultType, Result
 from lnst.RecipeCommon.BaseResultEvaluator import BaseResultEvaluator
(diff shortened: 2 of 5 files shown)


jtluka:
Looks good.

> I am also considering defining a custom class BaselineComparisonResult(Result) which would be used here instead which could make the filtering in analysis tools even easier because we could simply do:

Yes, I think that would help.

jtluka:
Looks like the internal run contains an issue with regards to internal LNST extensions:
```
_pickle.PicklingError: Can't pickle <class 'lnst.RHExtensions.CatalogConnector.CatalogedCPUMeasurementResults'>: attribute lookup CatalogedCPUMeasurementResults on lnst.RHExtensions.CatalogConnector failed
```

olichtne:
yeah, i'm working on debugging the exception, while doing so i feel much more the need for the custom class so i'll also add that.

olichtne:
so the issue with pickling is that we're using a custom **dynamically** created class. the dynamic creation is basically:

```python
new_class = type("new_name", (MixinClass, old_class), {})
```

and classes like this are not picklable or unpicklable - python doesn't know where to look for the code of such a class because there isn't any.

There are some online resources about how to potentially solve this that i'll investigate, or we'll have to think of a different way to implement our internal extension mixin class...

jtluka:
> and classes like this are not picklable or unpicklable - python doesn't know where to look for the code of such a class because there isn't any.

I somehow can't understand why this is a problem now? Since this has been working till this PR, but now it does not, the problem should be within this PR.

Or is this combination of these two?

olichtne:
i think upstream wise this pr should be ok, but when we combine it dynamically with a mixin in our internal code that creates a dynamic class and i think that can't be pickled.

it worked until now because i think we weren't pickling these objects until now.

but you may be right, i have some internal test code where i made the classes not dynamic and defined them statically, if that doesn't resolve the issue then the source of the problem is somewhere else.

Axonis:
Looks good, just quick thought - What about adding `recipe` object to the `comparisons` dict as well? Won't that be useful somewhere down the line, just so we have full info?

olichtne:
i think you could walk the `recipe_conf` to find a reference to the recipe, but i don't think this is necessary yet, we're talking about a `lnst.Controller.RecipeResults.BaseResult` type object which should be tracked in `lnst.Controller.Recipe.RecipeRun` instance in the results log, so you probably already have the `recipe` reference anyway

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
```diff
@@ -53,8 +58,23 @@ def evaluate_group_results(
             )
             comparison_result = ResultType.max_severity(comparison_result, comparison)
             result_text.extend(text)
+            comparisons.append(
+                {
+                    "current_result": result,
+                    "baseline_result": baseline,
+                    "comparison_result": comparison,
```
enhaut, line 65: wouldn't be something like `run_result` more descriptive? `comparison_result` is also a variable used few lines above that represents cumulative result

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
```diff
@@ -53,8 +58,23 @@ def evaluate_group_results(
...
+                    "text": text,
+                    "recipe_conf": recipe_conf,
```
enhaut, line 67: hmm, is it necessary to keep `recipe_conf` for every result? afaik many of results have common`recipe_conf`

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
olichtne: yeah i was debating that, but i'm not sure what else to call a result of a single comparison, maybe it would make sense to rename the other vars instead?
* comparison_result for a single comparison
* cumulative_result for the max_severity result (what is now called comparison_result)

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
olichtne: sure, but if it is the same object then object should only exist once, this will only add a reference to that object for convenience, it doesn't duplicate it

it's also non-trivial to find the perf recipe configuration related to a specific measurement result without this at this point and i'm not sure if we would want to add it into some form of a grouped result...

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
Axonis: I think we should be careful not to confuse this variable with possible reference to Catalog run as well, even though it has nothing to do with Catalog just yet.

olichtne: i did the renaming here so i think this should be ok.

@Axonis  what do you mean by Catalog confusion?

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
olichtne: is this still relevant or can we resolve this @enhaut ?

enhaut: hmm, im ok with that

lnst/RecipeCommon/Perf/Evaluators/BaselineEvaluator.py
Axonis: `run_result` could refer to a single run / iteration or a single test run or even catalog run object, that's why I wanted to avoid it.

olichtne: well, that's a different context though. from the point of view of the `BaselineEvaluator` you only have a current result and the baseline result, both of which are some kind of a `*PerfMeasurementResult` type.

either way, is the current state ok and can we resolve this thread or do you propose a different naming?

Axonis: Yes, I am aware of that.

I am okay with current name.

olichtne:
The MR now is rebased on top of master.

I added the `ENRT.DevInterruptHWConfigMixin: fix grep regex when bus_info missing` commit as it fixes an issue with running the `SimpleNetworkRecipe` via containers locally for me.

I also added a new commit that refactors the `compare_result_with_baseline` method.

Started a round of internal tests: `J:7169965`

lnst/RecipeCommon/Perf/Evaluators/BaselineCPUAverageEvaluator.py
```diff
@@ -77,28 +77,30 @@ def compare_result_with_baseline(
...
         baseline: PerfMeasurementResults,
         result_index: int = 0
-    ) -> Tuple[ResultType, List[str]]:
+    ) -> List[MetricComparison]:
         comparison = ResultType.FAIL
-        text = []
 
-        host = result.host.hostid
-        metric = f"{result_index}_{host}"
+        metric_name = f"{result_index}_utilization"
```
enhaut, line 83: if we are going to rename metric, let's use something more descriptive. what about `{result_index}_{generator/receiver}_utilization`/`{result_index}_{generator/receiver}_utilization_reversed`

enhaut (changes requested):

lnst/RecipeCommon/Perf/Evaluators/BaselineCPUAverageEvaluator.py
jtluka: The evaluator has no clue who is generator/receiver, so this is not possible. But we can still include the `{host}` in the string.

olichtne: to expand on what jan said, generator/receiver refers to **Flows** specifically, cpu measurements are completely independent of that, we can run cpu measurements even on nodes that are in between the generator and receiver.

i'm not 100% sure about including the host either, it's already part of the description of the evaluation:
```
    FAIL 191_TestResult:
        FAIL: CPU cpu: 191_utilization no threshold found
```

I want to keep this simple and only have the name of the metric in there to keep some sort of consistency across all different types of measurement evaluators which look at more/fewer metrics. We could for example add more metrics to the cpu evaluator and then the host id would be repeated everywhere...

So to summarize, IMO, it is actually very descriptive to use only `utilization`, as this is already `hostX` cpu measurement evaluation and we're looking at the `utilization` metric instead of for example `soft_irq`.

jtluka: I'm fine with keeping just the `{result_no}_utilization`.

olichtne:
rebased on top of current master. still need at least one approval for this.

enhaut (approved):
reviewed once again, lgtm

jtluka (approved):
Looks good

PR #259, opened 2022-10-13
add .dockerignore

jtluka:
Description:  

When the podman container lnst-agent image is build, the content of the directory is copied to the image along with .venv that may be present.

This however may contain incorrect path to the python executable.

To avoid this, the .dockerignore is added that excludes the .venv directory from COPY directive in Dockerfile.  

Tests:  

Tested locally.

Reviews:  

@olichtne @enhaut


olichtne (approved):
looks good to me

pgagne:
Shouldn't we be using `.containerignore`

Issue #260, opened 2022-10-27
ShortLivedConnectionsRecipe: using perf_warmup_duration for neper tcp_crr tests results in exception


jtluka:
This was detected in downstream testing however should be applicable to upstream as well.

The traceback:
```
Traceback (most recent call last):
  File "/mnt/tests/gitlab.cee.redhat.com/lnst-team/beaker-tasks/-/archive/master/beaker-tasks-master.tar.gz/lnst/test-runner/./do-my-test", line 38, in main
    ctl.run(recipe)
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/Controller/Controller.py", line 172, in run
    recipe.test()
  File "/root/rhextensions-lnst/lnst/RHExtensions/RHRecipeMixin.py", line 80, in test
    super(RHRecipeMixin, self).test()
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 172, in test
    self.do_tests(recipe_config)
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 294, in do_tests
    self.do_perf_tests(recipe_config)
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 320, in do_perf_tests
    self.perf_report_and_evaluate(result)
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/RecipeCommon/Perf/Recipe.py", line 191, in perf_report_and_evaluate
    aligned_results = results.time_aligned_results
  File "/root/virtualenvs/rhextensions-lnst-Xo1BSm3a-py3.9/lib/python3.9/site-packages/lnst/RecipeCommon/Perf/Recipe.py", line 123, in time_aligned_results
    real_times = self._get_measurement_timestamps(iteration_results_group)
[...]

PR #261, opened 2022-11-07
Trimming of neper results fix

enhaut:
Description:  

* fixed trimming of neper - Neper `tcp_crr` receiver_results are saved in just single interval `perf_duration` long, so this interval could not be trimmed out of first `N` samples. Because of that, disabling trimming based on receiver_results.
* moved all the *MeasurementResults classes to separate files in `Measurements/Results/*.py`

  
Tests:  

neper-only tests in `J:7215548`  
moved results classes tests in `J:7215416` basically just full test set we had

Reviews:  

@olichtne @jtluka @Axonis 

Closes: #260


jtluka:
From the splitting measurement results classes point of view I think it makes sense to have both non-aggregated and aggregated results class in the same file. The aggregated class inherits from the non-aggregated and that way you would not have to import it. Also when reading the code, at least for myself, I would not have to skip between the files.

It is not a big deal, but wondering what is the reasoning and benefit of such granularity?

So, for example:
`CPUMeasurementResults` and `AggregatedCPUMeasurementResults` can be in CPUStatMeasurementResults.py file.

@olichtne @enhaut @Axonis what are your thoughts on that?

enhaut:
> From the splitting measurement results classes point of view I think it makes sense to have both non-aggregated and aggregated results class in the same file. The aggregated class inherits from the non-aggregated and that way you would not have to import it. Also when reading the code, at least for myself, I would not have to skip between the files.
> 
> It is not a big deal, but wondering what is the reasoning and benefit of such granularity?
> 
> So, for example: `CPUMeasurementResults` and `AggregatedCPUMeasurementResults` can be in CPUStatMeasurementResults.py file.
> 
> @olichtne @enhaut @Axonis what are your thoughts on that?

Not a benefit at all. I split it into multiple files because if i remember correctly we were discussing same question in MR to one of our internal project and ended up with splitting it into multiple files.

lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
```diff
@@ -3,9 +3,8 @@
 from lnst.Controller.RecipeResults import ResultType
 from lnst.RecipeCommon.Perf.Measurements.MeasurementError import MeasurementError
 from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
-from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurementResults
-from lnst.RecipeCommon.Perf.Results import SequentialPerfResult
 from lnst.RecipeCommon.Perf.Results import ParallelPerfResult
+from lnst.RecipeCommon.Perf.Measurements.Results import FlowMeasurementResults, AggregatedFlowMeasurementResults
```
jtluka, line 7: Please swap these two lines, so that imports are grouped by modules.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
```diff
@@ -7,81 +7,21 @@
 import re
 
 from lnst.Tests.LinuxPerf import LinuxPerf
-from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import (
-    BaseMeasurement,
+from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
+from lnst.RecipeCommon.Perf.Measurements.Results import (
     BaseMeasurementResults,
+    LinuxPerfMeasurementResults,
```
jtluka, line 13: Wondering why we need the `BaseMeasurementResults`. I see that the measurement class uses this in type hinting, but wondering why we can't simply type-hint to `LinuxPerfMeasurementResults`.

Maybe a question for Jakub.

jtluka (commented):
Overall looks ok. Just few minor issues.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
jtluka: @Kuba314 could you take a look?

Kuba314: `LinuxPerfMeasurementResults` inherits from this. We can't remove the import.

We could change the type-hints to `LinuxPerfMeasurementResults` though. When I wrote this, I wanted to follow the same type-hints as are defined by the parent class, but that was illogical. We should change the type-hint on `collect_results` to `list[LinuxPerfMeasurementResult]` and *remove* the type-hint on the `results` variable in the same function since it's just noise and mypy figures this out on its own.

olichtne:
Generic comment first, I think the I'd reorganize the commits:

1. do the refactor of moving the result classes into their own package
2. add the new neper result class

olichtne:
On the topic of `Aggregated*Results` classes I think I'd prefer to do a bit more of a refactoring of these soon where we generalize them a bit more. The aggregation works in a predictable way, I think we don't necessarily need to write out each individual aggregated class like this and we could get away with a single "mixin/wrapper" class that can be used to simply define the aggregation ability of a results class to implement it.

After that the situation with the number of files should get a bit better.

So for now I'd keep it as is - each class separated into its own file, when we do this refactor I think we can revisit this topic.

I'll open an issue to describe some ideas/thoughts I have for refactoring the perf results classes.

lnst/RecipeCommon/Perf/Measurements/LinuxPerfMeasurement.py
olichtne: @Kuba314  i think that's what the situation was before the moving refactor that @enhaut did here. right now this `LinuxPerfMeasurement` module seems to only use the `LinuxPerfMeasurementResults` imo.

Kuba314: You're right. If the type gets changed, the import can of course be removed.

lnst/RecipeCommon/Perf/Measurements/NeperFlowMeasurement.py
```diff
@@ -5,8 +5,8 @@
 from lnst.Common.Utils import pairwise
 from lnst.Controller.Recipe import RecipeError
 from lnst.Controller.RecipeResults import ResultLevel
-from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import BaseFlowMeasurement, NetworkFlowTest, \
-    FlowMeasurementResults, Flow
+from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import BaseFlowMeasurement, NetworkFlowTest, Flow
+from lnst.RecipeCommon.Perf.Measurements.Results import NeperFlowMeasurementResults
```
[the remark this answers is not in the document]
enhaut: `Measurements.Results` got `__init__.py` file which provides `NeperFlowMeasurementResults`

enhaut:
Fixed issues, reordered commits.  
Tests:  
Neper tests only: `J:7261480`
Full test set: `J:7261486`

lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
enhaut: done

jtluka:
Code wise ok. Please fix the downstream test runs and post job ids.

enhaut:
@jtluka The previous runs failed due to forgotten downstream dependency, another one passed: `J:7265616`

enhaut:
Ah sorry, here is the rest:

* neper tests only `J:7283233`
* full test set `J:7283209`
* linuxperf measurement `J:7274922`

jtluka:
> neper tests only J:7283233

The tcp_crr tests have perf_warmup_duration=0 so we're not testing the fix here.

enhaut:
> > neper tests only J:7283233
> 
> The tcp_crr tests have perf_warmup_duration=0 so we're not testing the fix here.

Thanks for catch, but i guess you meant `J:7274922` that one is `IpsecEspAhCompRecipe` recipe, where is same problem as you've fixed in 89992bb5491079517f7340782c0bd6fefc4e8a93 . 

Fixed, here is the test - `J:7293411`. As it doesn't touch another parts of LNST full test set is not necessary.

jtluka:
> Thanks for catch, but i guess you meant J:7274922 that one is IpsecEspAhCompRecipe recipe, where is same problem as you've fixed in https://github.com/LNST-project/lnst/commit/89992bb5491079517f7340782c0bd6fefc4e8a93 .

No, the job I referenced was correct, see the parameters:
```
│     ├ Completed/Fail  /lnst/test-runner https://beaker.engineering.redhat.com/recipes/13014520#task153349628
│     │  ├ LNST_MODULE_NAME = lnst.RHExtensions.ExtendedRecipes
│     │  ├ LNST_RECIPE_NAME = ShortLivedConnectionsRecipe
│     │  ├ LNST_RECIPE_PARAMS = driver='ixgbe';perf_duration=5;perf_tool_cpu=[0];dev_intr_cpu=[0];adaptive_tx_coalescing=True;adaptive_rx_coalescing=True;catalog_workflow='RHEL8';ip_versions=['ipv4'];perf_tests=['tcp_crr'];perf_msg_sizes=[5120];rx_pause_frames=False;tx_pause_frames=False;disable_turboboost=True;minimal_idlestates_latency=0;drop_caches=True;save_sosreport=False;perf_warmup_duration=0;catalog_save=False;baseline_execution=False;thresholds={}
```

jtluka:
Ok, everything looks good now. I'm fine with merging this in.

jtluka (approved):
I approve.

olichtne (approved):
looks good to me.

PR #265, opened 2022-12-05
Test Network Address Parameters

pgagne:
Description:  

This is the start of changes to LNST to allow IP subnets to be passed as parameters. Right now only SimpleNetworkRecipe and VlansRecipe have it. More to come either in this PR or future PRs. 
  
Tests:  

Will upload test script soon. 

Reviews:  

@lnst-project/developers

Closes: #

### DIFF
--- a/lnst/Common/IpAddress.py
+++ b/lnst/Common/IpAddress.py
@@ -11,7 +11,12 @@ olichtne@redhat.com (Ondrej Lichtner)
 """
 
 import socket
+from collections.abc import Iterator
+from ipaddress import IPv4Network, IPv6Network, IPv4Interface, IPv6Interface, ip_interface
+from itertools import dropwhile, islice
 from socket import inet_pton, inet_ntop, AF_INET, AF_INET6
+from typing import Union, Optional
+
 from lnst.Common.LnstError import LnstError
 
 #TODO Replace this with Python's builtin ipaddress module.
@@ -125,6 +130,37 @@ class Ip6Address(BaseIpAddress):
         #constant from linux/if_addr.h
         IFA_F_TENTATIVE = 0x40
         return IFA_F_TENTATIVE & self.flags
+
+
+def interface_addresses(
+        network: Union[IPv4Network, IPv6Network],
+        default_start: Optional[Union[str, IPv4Interface, IPv6Interface]] = None,
+        default_skip: Optional[int] = None,
+) -> Iterator[BaseIpAddress]:
+    """
+    Generator of BaseIpAddress objects. Used to allocate interface addresses
+    (host IP + prefix length) from a network.
+
+    When used together with `NetworkParam` it provides a convenient way to
+    generate network addresses for the host interfaces used in LNST recipes.
+
+    If `default_start` is specified it is passed to `ipaddress.ip_interface()`
+    if that interface is in `network` the iteration will start at that address.
+
+    `default_skip` (only used when `default_start` is specified) allows you to
+    skip every N addresses, it's passed to `itertools.islice`.
+    """
+    hosts = network.hosts()
+    if default_start:
+        default_start = ip_interface(default_start)
+        if default_start in network:
+            hosts = dropwhile(lambda x: x != default_start.ip, hosts)
+            if default_skip:
+                hosts = islice(hosts, 0, None, default_skip)
+
+    for addr in hosts:
+        yield ipaddress(f"{addr}/{network.prefixlen}")
+
 
 def ipaddress(addr, flags=None):
     """Factory method to create a BaseIpAddress object"""
--- a/lnst/Common/Parameters.py
+++ b/lnst/Common/Parameters.py
@@ -12,6 +12,10 @@ olichtne@redhat.com (Ondrej Lichtner)
 
 import copy
 import re
+from ipaddress import AddressValueError, IPv4Network, IPv6Network, NetmaskValueError
+from socket import AddressFamily
+from typing import Optional, Union
+
 from lnst.Common.DeviceRef import DeviceRef
 from lnst.Common.IpAddress import BaseIpAddress, ipaddress
 from lnst.Common.LnstError import LnstError
@@ -76,11 +80,26 @@ class BoolParam(Param):
             raise ParamError("Value must be a boolean")
 
 class IpParam(Param):
+    def __init__(
+            self,
+            family: Optional[AddressFamily] = None,
+            multicast: bool = False,
+            **kwargs,
+    ):
+        self.require_family = family
+        self.require_multicast = multicast
+        super().__init__(**kwargs)
+
     def type_check(self, value):
         try:
-            return ipaddress(value)
+            value = ipaddress(value)
         except LnstError:
             raise ParamError("Value must be a BaseIpAddress object")
+        if self.require_family and value.family != self.require_family:
+            raise ParamError(f"Value must be of type {self.require_family.name}")
+        if self.require_multicast and not value.is_multicast:
+            raise ParamError("Value must be a multicast address")
+        return value
 
 class HostnameParam(Param):
     def type_check(self, value):
@@ -177,6 +196,37 @@ class ChoiceParam(Param):
         if value not in self._choices:
             raise ParamError(f"Value '{value}' not one of {self._choices}")
         return value
+
+
+class NetworkParam(Param):
+    """Network Param
+
+    Parameters of this type are used to specify a IPv4 or IPv6 networks.
+    (network address + prefix length).
+
+    For example::
+        vlan0_ipv4 = IPv4NetworkParam(default="192.168.10.0/24")
+        vlan0_ipv6 = IPv6NetworkParam(default="fc00:0:0:1::/64")
+    """
+    def __init__(self, type: Union[IPv4Network, IPv6Network], **kwargs):
+        self._type = type
+        super(NetworkParam, self).__init__(**kwargs)
+
+    def type_check(self, value: str):
+        try:
+            return self._type(value)
+        except (AddressValueError, NetmaskValueError) as e:
+            raise ParamError(f"Value {value} failed type check") from e
+
+
+class IPv4NetworkParam(NetworkParam):
+    def __init__(self, **kwargs):
+        super().__init__(IPv4Network, **kwargs)
+
+
+class IPv6NetworkParam(NetworkParam):
+    def __init__(self, **kwargs):
+        super().__init__(IPv6Network, **kwargs)
 
 
 class Parameters(object):
--- a/lnst/Recipes/ENRT/IpsecEspAhCompRecipe.py
+++ b/lnst/Recipes/ENRT/IpsecEspAhCompRecipe.py
@@ -1,9 +1,14 @@
 import signal
 import logging
 import copy
-from lnst.Common.IpAddress import ipaddress
+from lnst.Common.IpAddress import interface_addresses
 from lnst.Common.IpAddress import AF_INET, AF_INET6
-from lnst.Common.Parameters import StrParam, Param
+from lnst.Common.Parameters import (
+    Param,
+    StrParam,
+    IPv4NetworkParam,
+    IPv6NetworkParam,
+)
 from lnst.Common.LnstError import LnstError
 from lnst.Controller import HostReq, DeviceReq, RecipeParam
 from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
@@ -26,6 +31,11 @@ class IpsecEspAhCompRecipe(CommonHWSubConfigMixin, BaremetalEnrtRecipe,
     host2 = HostReq()
     host2.eth0 = DeviceReq(label="to_switch", driver=RecipeParam("driver"))
 
+    net1_ipv4 = IPv4NetworkParam(default="192.168.99.0/24")
+    net1_ipv6 = IPv6NetworkParam(default="fc00:1::/64")
+    net2_ipv4 = IPv4NetworkParam(default="192.168.100.0/24")
+    net2_ipv6 = IPv6NetworkParam(default="fc00:2::/64")
+
     ciphers = Param(default=[('aes', 128), ('aes', 256)])
     hashes = Param(default=[('hmac(md5)', 128), ('sha256', 256)])
     ipsec_mode = StrParam(default="transport")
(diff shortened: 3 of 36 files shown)


lnst/Recipes/ENRT/VlansRecipe.py
```diff
@@ -53,6 +53,15 @@ class VlansRecipe(VlanPingEvaluatorMixin,
         dict(gro="on", gso="on", tso="off", tx="off", rx="on"),
         dict(gro="on", gso="on", tso="on", tx="on", rx="off")))
 
+    vlan0_ipv4 = IPv4NetworkParam(default="192.168.10.0/24")
```
pgagne, line 56: hey do people think we should have a standard name for the IPv4Network param objects. 

For example if we values like `net1_ipv4`, `net2_ipv4, etc. I might be able to add `net1_ipv4`,net1_ipv6` to something like BaseEnrt then for test that require more they can add more to their own class or super class.

lnst/Common/IpAddress.py
```diff
@@ -126,6 +130,16 @@ def is_tentative(self):
         IFA_F_TENTATIVE = 0x40
         return IFA_F_TENTATIVE & self.flags
 
+
+def interface_addresses(network: Union[IPv4Network, IPv6Network]) -> Generator[BaseIpAddress]:
```
Kuba314, line 134: The return type should either be the full `Generator[BaseIpAddress, None, None]` or you should use `Iterator[BaseIpAddress]` instead, because I doubt we're going to use the generator's `send()` method.

jtluka: Please ammend the fix in the commit that adds this function. We want to have a bisectable commit history and having fixup commits added on top of a patch set would break this.

lnst/Common/Parameters.py
```diff
@@ -179,6 +180,27 @@ def type_check(self, value):
         return value
 
 
+class NetworkParam(Param):
```
jtluka, line 183: Either provide some details in the commit and/or add a docstring for the new classes as this provides new API to the user.

If you docstring for these they will automatically appear in https://lnst.readthedocs.io/en/latest/parameters.html

lnst/Common/IpAddress.py
```diff
@@ -126,6 +130,16 @@ def is_tentative(self):
...
+    """
+    Generator of BaseIpAddress used that iterates through all host addresses
```
jtluka, line 136: Extra word `... used ...`? Please fix or reword.

lnst/Common/Parameters.py
pgagne: fixedup the commit history during rebase

lnst/Recipes/ENRT/VlansRecipe.py
olichtne: i think this is specific enough to each test scenario that it is on the same level as the definition of the "network requirements"

so my opinion is that we should define the network parameters in the same class where we define `HostReq` and `DeviceReq` instances.

This basically means that we need to update each recipe individually and include the network parameters definition to each of them. Which is a bit of work, but I think it makes it very clear what and how the specific recipe and it's topology will look like/be configured.

olichtne (changes requested):
overall looks good to me, but i want this extended to update all recipes in the ENRT package before merging.

jtluka:
I created a [test branch](https://github.com/jtluka/lnst/tree/add-ip-adress-param-to-tunnel-recipes) with all TunnelRecipes updated.

I'm testing the changes in internal lab, once these succeed I will update this merge request with the commits.

lnst/Common/Parameters.py
jtluka: Unresolving. Neither commit has been updated with the description how this can be used, nor docstring was added.

pgagne: opps sorry resolved the wrong one.

lnst/Common/IpAddress.py
pgagne: Fixed let me know what you think, I will squash the commit for it later

lnst/Common/Parameters.py
```diff
@@ -179,6 +180,53 @@ def type_check(self, value):
...
+    """Network Param
+
+    This parameter is used to specify the IP networks a test will have
+     interfaces on. Using it allows multiple simultaneous LNST sessions on a shared broadcast domain.
+
+     It is used with physical or created logical interfaces (vlans, bonds, etc) have
+     addresses on external networks.
```
jtluka, line 190: I don't understand this sentence. What is meant by external networks?

lnst/Common/Parameters.py
```diff
@@ -179,6 +180,53 @@ def type_check(self, value):
...
+
+     Things like tunnel recipes generally use it for interfaces in the underlying network.
+
+     For example here is VlanRecipe use of it::
+        vlan0_id = IntParam(default=10)
+        vlan1_id = IntParam(default=20)
+        vlan2_id = IntParam(default=30)
```
jtluka, line 197: These IntParam parameters are imo irrelevant.

lnst/Common/Parameters.py
```diff
@@ -179,6 +180,53 @@ def type_check(self, value):
...
+
+        vlan0_ipv4 = IPv4NetworkParam(default="192.168.10.0/24")
+        vlan0_ipv6 = IPv6NetworkParam(default="fc00:0:0:1::/64")
+
+        vlan1_ipv4 = IPv4NetworkParam(default="192.168.20.0/24")
+        vlan1_ipv6 = IPv6NetworkParam(default="fc00:0:0:2::/64")
+
+        vlan2_ipv4 = IPv4NetworkParam(default="192.168.30.0/24")
+        vlan2_ipv6 = IPv6NetworkParam(default="fc00:0:0:3::/64")
```
jtluka, line 206: Maybe adding an example together with the `interface_address` generator would be more helpful. Also, to be less verbose here, a SimpleNetworkRecipe example would be enough I think.

lnst/Recipes/ENRT/TeamVsBondRecipe.py
```diff
@@ -1,5 +1,10 @@
-from lnst.Common.Parameters import IntParam, Param, StrParam
-from lnst.Common.IpAddress import ipaddress
+from lnst.Common.Parameters import (
+    Param,
+    StrParam,
```
jtluka, line 3: You removed `IntParam` which is used in the code later.

pgagne: Yeah i noticed that, fix incoming

lnst/Common/Parameters.py
pgagne: Yeah, looking at it again your probably right.

lnst/Common/Parameters.py
pgagne: that should probably say "that have addresses". ]

Yeah I basically want to say you only need to use it when the addresses you pick are exposed to the network outside the nic. Thats what I mean by "external network" here.  I'll see if i can reword it better, if you have any suggestions I'm all ears.

lnst/Common/Parameters.py
pgagne: I kinda wanted to put a slightly less trivial example that demonstrates multiple networks. I always hate it when docs only give simplistic examples. Although now that I've gone through the recipes I found a couple 2 network ones that might be better.

pgagne: Agreed on `interface_addresses` I could probably give a better example there.

lnst/Recipes/ENRT/VhostNetPvPRecipe.py
```diff
@@ -83,8 +85,9 @@ def test_wide_configuration(self, config):
         config.generator.nics.append(self.matched.generator_req.eth0)
         config.generator.nics.append(self.matched.generator_req.eth1)
 
-        self.matched.generator_req.eth0.ip_add(ipaddress("192.168.1.1/24"))
-        self.matched.generator_req.eth1.ip_add(ipaddress("192.168.1.2/24"))
+        ipv4_addr = interface_addresses(self.params.net_ipv4)
+        self.matched.generator_req.eth0.ip_add(next(ipv4_addr))
+        self.matched.generator_req.eth1.ip_add(next(ipv4_addr))
```
olichtne, line 90: fyi, for these dpdk related recipes we may also need to update the base class that does a "warmup" ping: https://github.com/pgagne/lnst/blob/45a862b9bf7d7666a14d6791c145ca60ec660f44/lnst/Recipes/ENRT/BasePvPRecipe.py#L105

i think the warmup is actually not really related so maybe we can also completely remove it, i think the idea i started with may not have been fully functional even though the thought behind it is valid - warmup here should fill up the switch forwarding table but from what i noticed, it gets automatically dumped anyway when we proceed with the dpdk reconfiguration that we do after the warmup.

lnst/Recipes/ENRT/IpsecEspAhCompRecipe.py
```diff
@@ -1,9 +1,9 @@
 import signal
 import logging
 import copy
-from lnst.Common.IpAddress import ipaddress
+from lnst.Common.IpAddress import interface_addresses
 from lnst.Common.IpAddress import AF_INET, AF_INET6
-from lnst.Common.Parameters import StrParam, Param
+from lnst.Common.Parameters import StrParam, IPv4NetworkParam, IPv6NetworkParam
```
jtluka, line 6: You lost Param while updating.

pgagne:
I think I have done the code to convert all the recipes over. If I have missed one please let me know in this thread. 

Still need to run tests and do other things, but all recipes should be converted.

pgagne (commented):
I believe I have converted all recipes over. If I missed one please let me know. 

Still need to resolve other requests and run tests, but all `Recipes.ENRT` classes should be using the network params now.

pgagne:
kernel tests : J:7458808

OVS tests: J:7458837

lnst/Common/Parameters.py
jtluka: > I always hate it when docs only give simplistic examples.

Sure, but you should start with a simple explanation and then build on top of that for more complex use.

The problem I see with the current documentation is that it's simply missing basic explanation what the parameter type holds. You jumped right into specific use cases, which imo is less important.

So, I'd start with following.

`NetwokParam` is a parameter type that can hold either IPv4 or IPv6 network address, specified as string. It automatically checks correctness of the passed network address.

When used together with `interface_address` it provides a convenient way to generate network addresses for the host interfaces used in LNST recipes.

For example, ....

... then you can provide more specific examples. But imo, that information is sufficient.

lnst/Recipes/ENRT/VhostNetPvPRecipe.py
pgagne: Will remove warmup code.

lnst/Common/Parameters.py
olichtne: i think you're overcomplicating the docstring here. this is a `lnst.Common.Parameters` module class... which means it can be used universally for any ip network use case a tester can imagine.

So IMO the docstring should primarily focus on describing the fact that this parameter class accepts strings formatted in a way that describe an IPv4 or IPv6 network.

And only after that provide an example of a potential use case where we use this specifically to configure ip addresses that our ENRT recipes use on the baremetal nics that are connected to shared networks to avoid conflicts.

... actually, this last part about how the ENRT test set uses this may be more fit to include as documentation in the ENRT package instead of here in the `Common` module... and then maybe link it from here as an example of a potential use?

lnst/Recipes/ENRT/MPTCPRecipe.py
```diff
@@ -83,13 +97,19 @@ def test_wide_configuration(self):
...
             # Need to disable rp_filter on server side
             host2.run("sysctl -w net.ipv4.conf.all.rp_filter=0")
 
         if "ipv6" in self.params.ip_versions:
             host1.mptcp.add_endpoints(host1.eth1.ips_filter(family=AF_INET6), flags=MPTCPFlags.MPTCP_PM_ADDR_FLAG_SUBFLOW)
-            host1.run(f"ip route add fc00::/64 dev {host1.eth1.name} via fc01::2 prio 10000")
+            host1.run(
+                f"ip route add {self.params.net1_ipv6} dev {host1.eth1.name}"
+                f" via {save_addrs[host2.eth1][AF_INET6]} prio 10000"
+            )
```
olichtne, line 112: hm, looking at this now, is this supposed to only be configured in one way? shouldn't a similar route be added also on `host2` towards `host1`?

this isn't really related to the ip networks work, but you creating the `save_addrs` dictionary sort of highlighted the asymmetry to me.

olichtne (changes requested):
checked that all affected recipes are modified and roughly checked that the changes make sense.

looks good to me, with just the two minor questions to answer.

if all internal tests pass fine after the fixups and reruns i think you can rebase and squash the fixup commits so that we can proceed with merging this in.

lnst/Recipes/ENRT/MPTCPRecipe.py
pgagne: Yeah im pretty sure its correct, the asymetry is due to the direction of traffic. If you want i can look at this later when i have some time to get back into mptcp headspace.

olichtne: discussed on irc, this should be ok.

lnst/Recipes/ENRT/IpsecEspAhCompRecipe.py
pgagne: Fixed

lnst/Common/Parameters.py
pgagne: Added a docstring, resolving this thread for now, we can discuss it in other threads

lnst/Common/Parameters.py
pgagne: Re did the interface_address and NetworkParam docstrings let me know what you think.

lnst/Common/IpAddress.py
pgagne: Re did the interface_address and NetworkParam docstrings let me know what you think.

lnst/Common/Parameters.py
pgagne: Fixed the interface_addresses and NetworkParams doc string let me know what you think

lnst/Common/IpAddress.py
```diff
@@ -126,6 +131,30 @@ def is_tentative(self):
...
+        default_start : Union[None, str, IPv4Interface, IPv6Interface] = None,
+) -> Iterator[BaseIpAddress]:
+    """
+    Generator of BaseIpAddress that iterates through all host addresses
+    in an IP network.
+
+    If `default_start` is specified it is passed to `ipaddress.ip_interface()`
+    if that interface is in `network` the iteration will start at that address.
+
+
```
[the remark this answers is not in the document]
pgagne: fixed

olichtne (approved):
looks ok to me now, the CI is failing for some weird reason for all commits/PRs anyway so ignoring that and merging.

PR #266, opened 2022-12-19
Common.Utils: Use command instead of which for is_installed check

Kuba314:
### Description
`command` is POSIX compatible and also is usually a shell built-in. `which` is a program that might not be installed on the machine.
  
### Tests
Tested just the `is_installed()` function. It works as expected.

### Reviews
@jtluka @enhaut


jtluka (approved):
Looks good.

PR #273, opened 2023-01-16
Add linux bridge recipe

jtluka:
### Description

This adds new recipe into ENRT recipe set. The network configuration is:
* a linux bridge device configured on top of bare NIC
* ipv4 and ipv6 address is configured on the bridge
* the test does not have any performance tweaks compared to e.g. SimpleNetworkRecipe

### Tests

Ran tests internally.

### Reviews

@olichtne

### DIFF
--- a/lnst/Recipes/ENRT/LinuxBridgeRecipe.py
+++ b/lnst/Recipes/ENRT/LinuxBridgeRecipe.py
@@ -1,0 +1,144 @@
+from lnst.Common.Parameters import IPv4NetworkParam, IPv6NetworkParam
+from lnst.Common.IpAddress import interface_addresses
+from lnst.Controller import HostReq, DeviceReq, RecipeParam
+from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
+from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
+from lnst.Recipes.ENRT.ConfigMixins.MTUHWConfigMixin import MTUHWConfigMixin
+from lnst.Devices import BridgeDevice
+
+
+class LinuxBridgeRecipe(MTUHWConfigMixin, BaremetalEnrtRecipe):
+    """
+    This recipe implements Enrt testing for a simple network scenario that looks
+    as follows
+
+    .. code-block:: none
+
+                    +--------+
+             +------+ switch +-----+
+             |      +--------+     |
+          +--+-+                 +-+--+
+        +-|eth0|-+             +-|eth0|-+
+        | +----+ |             | +----+ |
+        |   |    |             |   |    |
+        |  br0   |             |  br0   |
+        |        |             |        |
+        | host1  |             | host2  |
+        +--------+             +--------+
+
+    All sub configurations are included via Mixin classes.
+
+    The actual test machinery is implemented in the :any:`BaseEnrtRecipe` class.
+    """
+
+    host1 = HostReq()
+    host1.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    host2 = HostReq()
+    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    net_ipv4 = IPv4NetworkParam(default="192.168.101.0/24")
+    net_ipv6 = IPv6NetworkParam(default="fc00::/64")
+
+    def test_wide_configuration(self):
+        """
+        Test wide configuration for this recipe involves adding the matched
+        NICs into a Linux bridge and configuring an IPv4 and IPv6 address
+        on the bridge device on both hosts.
+
+        host1.br0 = 192.168.101.1/24 and fc00::1/64
+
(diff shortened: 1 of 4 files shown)


lnst/Recipes/ENRT/LinuxBridgeRecipe.py
```diff
@@ -0,0 +1,116 @@
...
+        host2.br0 = 192.168.101.2/24 and fc00::2/64
+        """
+        host1, host2 = self.matched.host1, self.matched.host2
+        configuration = super().test_wide_configuration()
+        configuration.test_wide_devices = []
+
+        ipv4_addr = interface_addresses(self.params.net_ipv4)
+        ipv6_addr = interface_addresses(self.params.net_ipv6)
+
+        for i, host in enumerate([host1, host2]):
```
enhaut, line 60: `i` is not used anywhere

enhaut (changes requested):

Kuba314:
This code is, I suspect, mostly similar to every other bridge recipe. @jtluka @olichtne Is there any refactoring planned for this? Am I correct in thinking that there's a lot of boilerplate code here?

lnst/Recipes/ENRT/LinuxBridgeRecipe.py
```diff
@@ -0,0 +1,116 @@
...
+            host.br0 = BridgeDevice()
+            host.eth0.down()
+            host.br0.slave_add(host.eth0)
+
+            for dev in [host.eth0, host.br0]:
```
pgagne, line 65: is a for loop really needed here, its the same amount of lines to do:

```
python
host.eth0.up()
host.br0.up()
```
```

Kuba314: I think this should even be done automatically behind the scenes... Is it possible to get all created devices and call the `.up()` method on them in a single place? I don't think it's possible now, but it definitely could be. I'm just not sure if this is sane from a design perspective.

jtluka: > I think this should even be done automatically behind the scenes... Is it possible to get all created devices and call the `.up()` method on them in a single place? I don't think it's possible now, but it definitely could be. I'm just not sure if this is sane from a design perspective.

On one hand I agree, automating things would help the test writer. The problem comes when you want to have control over these things when analyzing issues identified by the test. I think we should follow the same workflow as if the network stack is configured with ip route. Another problem is that you as a test writer need to know that something is done automatically behind the scenes. Being explicit is imo better.

jtluka: > is a for loop really needed here, its the same amount of lines to do:
> 
> ```python
> host.eth0.up()
> host.br0.up()
> ```

Good point. I can change that.

lnst/Recipes/ENRT/LinuxBridgeRecipe.py
jtluka: Will fix.

jtluka: Fixed.

lnst/Recipes/ENRT/LinuxBridgeRecipe.py
jtluka: Fixed.

olichtne (changes requested):
I don't see you adding the `generate_perf_endpoints` property so this test won't run any performance measurements. Is this intentional?

The PR description only mentions that no tweaks are included but not that no performance tests should be run.

jtluka:
> I don't see you adding the `generate_perf_endpoints` property so this test won't run any performance measurements. Is this intentional?
> 
> The PR description only mentions that no tweaks are included but not that no performance tests should be run.

Thanks for catching this. We do want to run performance tests on that setup, so I will have to fix this.

lnst/Recipes/ENRT/LinuxBridgeRecipe.py
olichtne: > I think this should even be done automatically behind the scenes... Is it possible to get all created devices and call the `.up()` method on them in a single place? I don't think it's possible now, but it definitely could be. I'm just not sure if this is sane from a design perspective.

This is specifically because you want to be as explicit about what type of configuration you want to do for this specific test. Setting a device up/down during configuration is a very important step and as a test writer/user you want to know exactly when it happens. So the choice of doing it like this instead of behind the scenes was very much intentional.

At least for now I'd keep these as is, at some point we may want to take a closer look to see if it makes sense to refactor the test set implementations a bit more to improve some very obvious and common patterns but I don't think this is the right time for that.

olichtne (approved):
looks good to me now.

jtluka:
Reran the last version of the test with containers setup.

All of ping, ping6, iperf ipv4 tcp_stream, iperf ipv6 tcp_stream works fine.

olichtne (approved):

PR #274, opened 2023-01-17
Controller/ContainerPoolManager: enable use with firewalld

jtluka:
### Description

This fixes connectivity problems when firewalld is running on the host that configures the pool with containers.

Basically, all interfaces that are generated by ContainerPoolManager must be in the same zone.

If the host does not have firewalld enabled, the configuration is skipped.

TODO:
* error handling
* make this configurable, preferably disable by default?

### Tests

Tested locally on Fedora36.

### Reviews
@all for generic review, I'll make this a draft for now.

### DIFF
--- a/lnst/Controller/ContainerPoolManager.py
+++ b/lnst/Controller/ContainerPoolManager.py
@@ -6,6 +6,7 @@ from json import loads
 from lnst.Controller.AgentPoolManager import PoolManagerError
 from lnst.Controller.Machine import Machine
 from lnst.Common.DependencyError import DependencyError
+from lnst.Common.Utils import check_process_running
 
 
 class ContainerPoolManager(object):
@@ -59,6 +60,7 @@ class ContainerPoolManager(object):
         self._network_prefix = "lnst_container_net_"
         self._start_timeout = 5
         self._pool_check = pool_checks
+        self._firewalld_config = {} # interface to firewalld zone mapping
 
     @property
     def image(self):
@@ -247,7 +249,19 @@ class ContainerPoolManager(object):
 
         self._networks[name] = network
 
+        if check_process_running("firewalld"):
+            self._add_interface_to_firewalld_zone("cni-podman1", "internal")
+
         return network
+
+    def _add_interface_to_firewalld_zone(self, interface, zone):
+        logging.debug(f"Adding interface {interface} to firewalld zone {zone}")
+        output = subprocess.check_output(
+            f"firewall-cmd --zone={zone} --change-interface={interface}",
+            shell=True
+        )
+
+        self._firewalld_config[interface] = zone
 
     def _connect_to_network(self, container: "Container", network: "Network"):
         """There is no way to get MAC address of remote interface except
@@ -284,6 +298,31 @@ class ContainerPoolManager(object):
             "params": {"hwaddr": interface["address"], "driver": "veth"},
             "network": network.name,
         }
+
+        logging.debug(f"Added {eth} with {interface['address']}")
+
+        if check_process_running("firewalld"):
+            # add interface to trusted zone
+            host_interfaces = loads(
+                subprocess.check_output(
+                    "ip -json link",
+                    shell=True
+                ).decode("utf-8")
+            )
+            matched_host_interfaces = [
+                host_iface for host_iface in host_interfaces \
+                    if host_iface["ifindex"] == interface["link_index"]
+            ]
+
+            if not matched_host_interfaces:
+                raise Exception(
+                    "No match for {} in\n{}".format(
+                        interface["address"],
+                        host_interfaces
+                    )
+                )
+
+            self._add_interface_to_firewalld_zone(matched_host_interfaces[0]['ifname'], "internal")
 
         return True
 
@@ -329,3 +368,15 @@ class ContainerPoolManager(object):
     def cleanup(self):
         self.cleanup_containers()
         self.cleanup_networks()
+        if check_process_running("firewalld"):
+            self._cleanup_firewalld()
+
+    def _cleanup_firewalld(self):
+        for interface, zone in self._firewalld_config.items():
+            logging.debug(f"Removing interface {interface} from firewalld zone {zone}")
+            output = subprocess.check_output(
+                f"firewall-cmd --zone={zone} --remove-interface={interface}",
+                shell=True
+            )
+
+        self._firewalld_config = {}


jtluka:
The tests failed because the host does not have the firewall-cmd. I will have to adjust the code to check if the command is present.

jtluka:
> The tests failed because the host does not have the firewall-cmd. I will have to adjust the code to check if the command is present.

Actually I misread the test logs. The problem is that there's no check for firewalld when podman bridge is added to the firewall zone. The check for firewall-cmd is not required. I will repush the changes.

jtluka:
Note that the current implementation does not work correctly for IPv6.

jtluka:
> Note that the current implementation does not work correctly for IPv6.

The problem is only with LinuxBridgeRecipe that I used from https://github.com/LNST-project/lnst/pull/273

The problem is with duplicate address detection where the link local IPv6 address is the same for eth and bridge device:
```
inet6 fe80::d46d:77ff:fe41:982d/64 scope link dadfailed tentative
```

jtluka:
> > Note that the current implementation does not work correctly for IPv6.
> 
> The problem is only with LinuxBridgeRecipe that I used from #273
> 
> The problem is with duplicate address detection where the link local IPv6 address is the same for eth and bridge device:
> 
> ```
> inet6 fe80::d46d:77ff:fe41:982d/64 scope link dadfailed tentative
> ```

The problem is present with SimpleNetworkRecipe, too. Disabling firewalld resolves the issue, so it's not related to IPv6 DAD.

jtluka:
I will need to consult this with Eric Garver who's firewalld expert.

olichtne:
fyi, i think the `@ALL` you used doesn't really work on github, it links to some org/user and afaik it doesn't actually ping anyone in the lnst-project.

i think you need to do `@LNST-project/lnst-developers` instead

lnst/Controller/ContainerPoolManager.py
```diff
@@ -285,6 +299,31 @@ def _connect_to_network(self, container: "Container", network: "Network"):
...
+            # add interface to trusted zone
+            host_interfaces = loads(
+                subprocess.check_output(
+                    "ip -json link",
+                    shell=True
+                ).decode("utf-8")
+            )
+            matched_host_interfaces = [
+                host_iface for host_iface in host_interfaces \
+                    if host_iface["ifindex"] == interface["link_index"]
```
olichtne, line 314: if i understand it correctly, `interface` is defined at https://github.com/LNST-project/lnst/blob/master/lnst/Controller/ContainerPoolManager.py#L271 which is also a json output of an `ip` command

if that's true then i think `interface['ifname']` should also be defined, and so i think you don't have to run this search for the interface?

am i missing something?

olichtne: maybe `interface` is INSIDE the container and you're looking for the `ifname` OUTSIDE the container?

lnst/Controller/ContainerPoolManager.py
```diff
@@ -329,3 +368,15 @@ def cleanup_networks(self):
     def cleanup(self):
         self.cleanup_containers()
         self.cleanup_networks()
+        if check_process_running("firewalld"):
```
olichtne, line 371: at this point i think you don't have to check that the process is running but that you have done and saved some firewalld configuration in the `self._firewalld_config` dictionary

olichtne: i say that because i think you can change the firewalld configuration with the `firewall-cmd` commands even if the process isn't actively running but i may be misremembering something... probably good to check that, but i don't have firewalld on my laptop.

lnst/Controller/ContainerPoolManager.py
jtluka: Yeah, the problem is that `ip link` shows only ifindex of the device in the other namespace, not the device name.

lnst/Controller/ContainerPoolManager.py
jtluka: Yeah, I was thinking about the same, what you propose may be better. I will change that.

jtluka: > i say that because i think you can change the firewalld configuration with the firewall-cmd commands even if the process isn't actively running but i may be misremembering something...

I don't understand. If firewalld is not running I assume that the service is not enabled and therefore no need to modify firewall configuration.

olichtne: what i was trying to say is that i seem to remember that firewalld works as a database that is also editable offline when the actual firewall service isn't running - so the check process would return `False` and we'd keep our configuration in the database without cleaning it up.

this would be an edge case of when the service crashes or is somehow shut down during the test which is of course unlikely.

But, my main reason for the proposal is that it also simplifies the code a bit - you don't have to check for a running process, since the database is *always* editable (even if the service is offline), you can simply trust your internal `_firewalld_config` and clean up the database without the process check.

I checked and my assumption about the offline database is true, see:
* https://firewalld.org/documentation/configuration/runtime-versus-permanent.html
* https://firewalld.org/documentation/utilities/firewall-offline-cmd.html

But from the `firewalld-cmd` commands that you use here I think we're only changing the runtime configuration, not the permanent one, which means that in case the service crashes, our configuration automatically disappears.

Still the code is probably a bit simpler without the process check for the cleanup.

jtluka: Ok, let me try this on Fedora and if it works, I'll remove check for firewalld process running.

jtluka: Here's a try with the firewalld check removed:
```
2023-01-25 18:25:42       (localhost)        -   DEBUG: Adding interface cni-podman1 to firewalld zone internal
FirewallD is not running
Command 'firewall-cmd --zone=internal --change-interface=cni-podman1' returned non-zero exit status 252.
```

Not sure if `--permanent` would make any difference, but I'd rather avoid changing permanent configuration.

Eventually, we could check the `firewall-cmd` error code, where the exit code of 252 means the firewalld service is not running. I think this would be worth to try.

jtluka: After offline discussion here's the plan:
* checking whether firewalld is running is better, since if it's running that means the firewall-cmd is available and I don't have to check for its presence
* when check for the return code is used, I'd have to also check whether the fail is caused by firewall-cmd not being present

Additionally, I should double check whether the overall solution is valid with Eric Garver.

One more note, by default the firewalld configuration should not be touched, this should be configurable through ContainerPoolManager setting.

jtluka:
So, I tried to reproduce the issue on fresh Fedora 36 installation and I was not able to reproduce the issue there.

Most likely the setup where I hit this:
* did not have `network_backend` set to `cni` as described in https://lnst.readthedocs.io/en/latest/extensions.html#configure-podman-to-use-cni
* I had `podman-docker` rpm installed that created `docker` zone in firewalld - I think this was not the reason, as the zone still exists after removing the package and rebooting

Most likely the 'cni' backend + reboot resolved this issue for me.

With that I'm closing this pull request.

PR #279, opened 2023-01-27
Add linux bridge over bond recipe

jtluka:
### Description

This adds new recipe into ENRT recipe set. The network configuration is:

* a bond device is created over two NICs
* a linux bridge device is created and bond device is added to the bridge
* ipv4 and ipv6 address is configured on the bridge
* the test does not have any performance tweaks compared to e.g. SimpleNetworkRecipe

### Tests

Tested with containers locally (removed one DeviceReq form the code as containers do not support that).

Redhat lab job id 7503695

### Reviews

@all


jtluka:
The beaker job ran fine, this is ready to merge.

olichtne (approved):

PR #281, opened 2023-02-08
Revert "pyproject.toml: enforce pyroute2 0.7.3"

jtluka:
### Description

This reverts commit 4e9ea1f0a20b3d3213e5f272fdac22e70b0c7a94.

The pyroute2 issues has been fixed, https://github.com/svinota/pyroute2/issues/1068#issuecomment-1407366608

### Tests

github CI covers this.

### Reviews

@olichtne

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ include = ["schema-am.rng", "install/*", "lnst-ctl.conf"]
 python = "^3.9"
 lxml = "*"
 ethtool = "*"
-pyroute2 = "0.7.3"
+pyroute2 = "*"
 
 libvirt-python = {version = "*", optional = true }
 podman = {version = "*", optional = true }


jtluka:
Hmm, I think the fixed version actually was not picked up. I will have to test this differently. Probably by specifying the fixed version somehow in the pyproject.toml. Even if the tests pass this way, I will have to wait for Pete to tag the fixed version so that it can be used by poetry.

Marking as draft until I check all of this.

jtluka:
> Hmm, I think the fixed version actually was not picked up. I will have to test this differently. Probably by specifying the fixed version somehow in the pyproject.toml. Even if the tests pass this way, I will have to wait for Pete to tag the fixed version so that it can be used by poetry.
> 
> Marking as draft until I check all of this.

Used following to test the pyroute upstream fix:
```
poetry add git+https://github.com/svinota/pyroute2#f62792c6d3f952efbe348189fbd2b4334ef453d0
```

Pushed a commit that should be reverted before merging.

jtluka:
@olichtne so, we should be all set here. Do not merge this yet, we need to wait for Pete to tag the version so that it's included in pypi and poetry then can fetch it.

jtluka:
So, pyroute with the fix is now tagged as 0.7.5. I'll update this MR with update of pyroute2 to that version.

jtluka:
> So, pyroute with the fix is now tagged as 0.7.5. I'll update this MR with update of pyroute2 to that version.

Originally we used any latest version, so I'll use that instead.

olichtne (approved):

PR #282, opened 2023-02-08
Recipes/BaseTunnelRecipe: add generate_perf_endpoints()

jtluka:
### Description

This enables all recipes derived from BaseTunnelRecipe to run performance tests.

### Tests

Test run of GreTunnelRecipe in RH labs, job id 7512901.


olichtne:
> @olichtne could you just take a quick look if having the generate_perf_endpoints() in BaseTunnelRecipe (instead of in individual recipe implementations) looks sane?

yeah, i think that is sane and should be ok to do.

olichtne (approved):

Kuba314 (approved):

olichtne (approved):

PR #286, opened 2023-02-16
Setting initial state of used devices to down

enhaut:
### Description
* setting initial state of used devices to down -
  * Sometimes may happen, that used interfaces are not in down state which may leads to nondeterministic behaviour of recipes that relies on initial state of device. Setting used device to down state as part of pre-test configuration sets same initial state of devices across the recipes.
* Removed setting devices down from MACsec recipes, which was just 
workaround

### Tests
Scheduled 1 test for each recipe class we have with my changes `J:7541084` and without my changes `J:7541167` to check whether my changes breaks (changes) indexing or not (should not).

### Reviews
@LNST-project/lnst-developers

### DIFF
--- a/lnst/Agent/Agent.py
+++ b/lnst/Agent/Agent.py
@@ -131,6 +131,7 @@ class RemoteMethods:
 
         for device in self._if_manager.get_devices():
             try:
+                device.down()
                 device.store_cleanup_data()
             except DeviceDisabled:
                 pass
--- a/lnst/Recipes/ENRT/SimpleMacsecRecipe.py
+++ b/lnst/Recipes/ENRT/SimpleMacsecRecipe.py
@@ -33,9 +33,6 @@ class SimpleMacsecRecipe(CommonHWSubConfigMixin, BaremetalEnrtRecipe):
         configuration = super().test_wide_configuration()
         configuration.test_wide_devices = [host1.eth0, host2.eth0]
 
-        host1.eth0.down()
-        host2.eth0.down()
-
         self.wait_tentative_ips(configuration.test_wide_devices)
 
         configuration.endpoint1 = host1.eth0


olichtne (approved):

jtluka (approved):

enhaut:
I wrote a simple script that parses metrics from all the recipes:
```python
import os
import pprint
from rhextensions_lrc_file import LrcFile


recipe_metrics = {}  # recipe_class: metrics

# get metrics from lrc files:
for lrc in os.listdir("/tmp/changed/"):
    if not lrc.endswith(".lrc"):
        continue
    try:
        lrc_file = LrcFile(os.path.join("/tmp/changed/", lrc))
    except Exception as e:
        print(f"Could not load {lrc}: {e}")
        continue

    recipe_metrics[lrc_file.recipe_name] = sorted(list(lrc_file.metrics.keys()))


# sort recipes:
recipe_metrics = {k: recipe_metrics[k] for k in sorted(recipe_metrics.keys())}

pprint.pprint(recipe_metrics)
```
The output of script is just `{RECIPE_CLASS: ["XX_metric", "YY_metric"]}`.

Compared outputs of analysis on lrcs that includes changes and the ones without changes:
```
62,67c62,67
<  'SimpleMacsecRecipe': ['178_utilization',
<                         '179_utilization',
<                         '180_generator_cpu_data',
<                         '180_generator_flow_data',
<                         '180_receiver_cpu_data',
<                         '180_receiver_flow_data'],
---
>  'SimpleMacsecRecipe': ['176_utilization',
>                         '177_utilization',
>                         '178_generator_cpu_data',
>                         '178_generator_flow_data',
>                         '178_receiver_cpu_data',
>                         '178_receiver_flow_data'],
```


All the metric indexes **remain the same**, except `MACsec` ones. The reason is, that `MACsec` recipe was setting interface down, that was/could be down which afterwards leads to setting interface up by recipe itself, so it generated 2 more task results.


I'd say, we can proceed with merging this in.

PR #289, opened 2023-02-28
Common.Utils: Avoid using regex inline flags

Kuba314:
### Description
Regex inline flags have been deprecated since Python 3.6. The deprecation warning is an exception since Python 3.11(?).

We should use `re.IGNORECASE` flag to achieve the same behaviour. I also combined the "true" and "yes" checks to a single regex pattern.

### Tests
```py
>>> re.match("^\s*(true|yes)", "True", flags=re.IGNORECASE)
<re.Match object; span=(0, 4), match='True'>
>>> re.match("^\s*(true|yes)", "yeS", flags=re.IGNORECASE)
<re.Match object; span=(0, 3), match='yeS'>
```

### Reviews
@pgagne @olichtne

### DIFF
--- a/lnst/Common/Utils.py
+++ b/lnst/Common/Utils.py
@@ -95,9 +95,9 @@ def int_it(val):
 
 def bool_it(val):
     if isinstance(val, str):
-        if re.match("^\s*(?i)(true)", val) or re.match("^\s*(?i)(yes)", val):
+        if re.match("^\s*(true|yes)", val, flags=re.IGNORECASE):
             return True
-        elif re.match("^\s*(?i)(false)", val) or re.match("^\s*(?i)(no)", val):
+        elif re.match("^\s*(false|no)", val, flags=re.IGNORECASE):
             return False
     return True if int_it(val) else False
 


lnst/Common/Utils.py
```diff
@@ -95,9 +95,9 @@ def int_it(val):
 
 def bool_it(val):
     if isinstance(val, str):
-        if re.match("^\s*(?i)(true)", val) or re.match("^\s*(?i)(yes)", val):
+        if re.match("^\s*(true|yes)", val, flags=re.IGNORECASE):
```
pgagne, line 98: Perhaps we should also complile these regexes

pgagne:
Thanks for looking at this Jakub

lnst/Common/Utils.py
olichtne: i wonder how much gain there would be from that... the function isn't really called that frequently that this is actually slowing us down...

i mean, we could as well compile all of the regexes that we use everywhere, but honestly as noted in https://docs.python.org/3/library/re.html?highlight=re%20compile#re.compile the recently used regexes are cached and don't really need to be compiled in most cases.

so i think keeping this simpler is fine since we're not really seeing any significant performance hit from this...

olichtne (approved):

lnst/Common/Utils.py
pgagne: Didnt realized python caches recent regexes, interesting. 

I agree its okay to keep it simple.

PR #293, opened 2023-03-22
Using rules/sec for TrafficControlRecipe results

pgagne:
### Description
Previously we where evaluating `time_taken`. This changes it to evaluate `rules/sec`

### DIFF
--- a/lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
+++ b/lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
@@ -18,7 +18,7 @@ class BaselineTcRunAverageEvaluator(BaselineEvaluator):
             result_index: int = 0,
     ) -> list[MetricComparison]:
 
-        metric_name = f"{result_index}_time_taken"
+        metric_name = f"{result_index}_rule_install_rate"
 
         if baseline is None:
             return [
@@ -29,14 +29,15 @@ class BaselineTcRunAverageEvaluator(BaselineEvaluator):
                 )
             ]
         elif (threshold := self._thresholds.get(metric_name, None)) is not None:
-            difference = ((result.time_taken / baseline.time_taken) * 100 ) - 100
+            difference = ((result.rule_install_rate.average / baseline.rule_install_rate.average) * 100) - 100
             direction = "higher" if difference >= 0 else "lower"
             text = [
                 f"{self.__class__.__name__} of tc run with {metric_name}",
                 f"{result.description}",
-                f"baseline time_taken={baseline.time_taken}",
-                f"{difference:2f} {direction} than baseline ",
-                f"Allowed differences: {threshold}% ",
+                f"Baseline: {baseline.rule_install_rate.average} rules/sec",
+                f"Measured: {result.rule_install_rate.average} rules/sec",
+                f"{abs(difference):2f}% {direction} than baseline ",
+                f"Allowed difference: {threshold}% ",
             ]
             if difference < -threshold:
                 comparison = ResultType.WARNING
--- a/lnst/RecipeCommon/Perf/Evaluators/MaxTimeTakenEvaluator.py
+++ b/lnst/RecipeCommon/Perf/Evaluators/MaxTimeTakenEvaluator.py
@@ -13,7 +13,7 @@ class MaxTimeTakenEvaluator(BaseResultEvaluator):
         self._max_time = max_time
 
     def _check_interval(self, timed_run: TcRunMeasurementResults) -> ResultType:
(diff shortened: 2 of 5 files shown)


olichtne (dismissed):
would be good to see logs with the baseline evaluation using some actual baseline, but this looks ok imo.

lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
```diff
@@ -18,7 +18,7 @@ def compare_result_with_baseline(
             result_index: int = 0,
     ) -> list[MetricComparison]:
 
-        metric_name = f"{result_index}_time_taken"
+        metric_name = f"{result_index}_run_interval"
```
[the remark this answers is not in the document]
pgagne: I could maybe see that. I don't know why I picked `run_interval`.  Not sure about calling it `rules_per_sec` but maybe we need to talk about it. 

Right now `run_interval` is to `rules_per_sec` like `generator_results` is to `generator_throughput` with flows. It terms of how the results get stored and what not. I am not 100% sure why we have that distinction but lets discuss it.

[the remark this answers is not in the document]
olichtne: technically, rules per second is the unit type of a metric that should be called something like `throughput` where that refers to the throughput of the tc implementation - how quickly can it push rules into the subsystem...

the `generator_results` and `generator_throughput` are the way that they are mostly due to the fact that the implementation moved and the context changed slightly over time but we haven't updated the keys very consistently... at this point imo it should all be named {`generator,receiver}_{throughput, cpu_utilization}` but... compatibility with older code/results is more important in this at this point.

for tc, we don't need the backwards compatibility so this is a good time to come up with a good descriptive name.

lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
```diff
@@ -29,12 +29,12 @@ def compare_result_with_baseline(
...
             ]
         elif (threshold := self._thresholds.get(metric_name, None)) is not None:
-            difference = ((result.time_taken / baseline.time_taken) * 100 ) - 100
+            difference = ((result.rule_install_rate.average / baseline.rule_install_rate.average) * 100) - 100
             direction = "higher" if difference >= 0 else "lower"
             text = [
                 f"{self.__class__.__name__} of tc run with {metric_name}",
                 f"{result.description}",
-                f"baseline time_taken={baseline.time_taken}",
+                f"baseline rules_per_sec={baseline.rule_install_rate.average}",
```
olichtne, line 37: checked the newer job you sent me and i think this could be a bit more descriptive. what you have now:
```
        host1.ens5f0np0 multi tc run num_instances=4 mean time_taken=10.508329783249991s num_rules=(100000, 100000, 100000, 100000)
        baseline rules_per_sec=9552.86526092749
        -0.383185 lower than baseline 
        Allowed differences: 1.42% 
```

so you're comparing "time taken" which is printed in the `result.description` part, but comparing to `rules rate` printed here... i think it would be a bit nicer to see `rules rate` to `rules rate` comparison here.

pgagne: Yeah this description could probably be clean up a bit.

pgagne: Fixed let me know what you think heres what it looks like now:
```
    host1.ens1f0np0 multi tc run num_instances=1 mean time_taken=3.740612900000997s num_rules=(100000,)
    Baseline: 27106.8640311474 rules/sec
    Measured: 26733.586894268945 rules/sec
    1.377058% lower than baseline 
    Allowed difference: 3.0% 
```

lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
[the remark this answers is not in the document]
pgagne: Renamed to `rule_install_rate`. Let me know what you think.

pgagne:
Updated to rename `run_interval` to `rule_install_rate` . Let me know what you think 

```
J:7677844
J:7677920
J:7678011
J:7678174
```

jtluka (approved):
Looks good.

lnst/RecipeCommon/Perf/Evaluators/BaselineTcRunAverageEvaluator.py
olichtne: looks good

olichtne (approved):

PR #295, opened 2023-04-06
Use PerfParalellResult and `asyncio.subprocess` for TrafficControl results

pgagne:
### Description
Changing to perf paralell result. mostly complete. commiting for test and review purposes. 

Also use asyncio.subprocess to run multiple tc instances client side. 

For reference here is a sample of the `tc` process start timestamps: 
```
    data:
        instance_results:
            item 0:
                ...
                start_timestamp: 1681069428.9487906
                ...
            item 1:
                ...
                start_timestamp: 1681069428.9511075
                ... 
            item 2:
                ...
                start_timestamp: 1681069428.9530337
                ...
            item 3:
                ...
                start_timestamp: 1681069428.9548721
                ...
```

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
@@ -9,7 +9,7 @@ from lnst.Controller.RecipeResults import ResultLevel, ResultType
 from lnst.RecipeCommon.Perf.Measurements.MeasurementError import MeasurementError
 from lnst.RecipeCommon.Perf.Measurements.Results.AggregatedTcRunMeasurementResults import AggregatedTcRunMeasurementResults
 from lnst.RecipeCommon.Perf.Measurements.Results.TcRunMeasurementResults import TcRunMeasurementResults
-from lnst.RecipeCommon.Perf.Results import PerfInterval
+from lnst.RecipeCommon.Perf.Results import PerfInterval, ParallelPerfResult
 from lnst.Tests.TrafficControl import TrafficControlRunner
 from lnst.Controller.Job import Job
 from lnst.Controller.Namespace import Device, Namespace
@@ -159,16 +159,12 @@ class TcRunMeasurement(BaseMeasurement):
             self._running_jobs.append(job)
 
     def _prepare_jobs(self) -> list[Job]:
-        jobs = []
-        for instance in self.instance_configs:
-            batchfile = instance.generate_batchfile()
-            job = instance.host.prepare_job(
-                TrafficControlRunner(batchfile=batchfile),
-                job_level=ResultLevel.NORMAL,
-            )
-            instance.instance_job = job
-            jobs.append(job)
-        return jobs
+        batchfiles = [i.generate_batchfile() for i in self.instance_configs]
+        job = self.host.prepare_job(
+            TrafficControlRunner(batchfiles=batchfiles),
+            job_level=ResultLevel.NORMAL,
+        )
+        return [job]
 
     def finish(self):
         jobs = self._running_jobs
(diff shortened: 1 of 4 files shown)


pgagne:
J:7716681

lnst/Tests/TrafficControl.py
```diff
@@ -1,45 +1,48 @@
...
-        time_taken = et - st
-
-        return start_timestamp, time_taken, res
+        start_time = time.perf_counter()
+        proc = await asyncio.create_subprocess_exec(
+            tc_exec, "-b", batchfile,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
+        stdout, stderr = await proc.communicate()
```
olichtne, line 38: the previous version of the test module also logged any errors, this is probably a good idea to do in this version as well, especially since we're now running multiple processes at once - having more debug info could be useful in case something goes wrong.

lnst/Tests/TrafficControl.py
```diff
@@ -1,45 +1,48 @@
...
-    def run_tc(self, batchfile: str) -> tuple[float, subprocess.CompletedProcess]:
-
+        instance_results = asyncio.run(self.run_instances())
+        all_success = all((i["success"] for i in instance_results))
+        self._res_data["data"] = dict(
+            instance_results=instance_results,
+        )
+        return all_success
+
+    async def run_instances(self):
```
olichtne, line 21: this should have a return type defined.

lnst/Tests/TrafficControl.py
```diff
@@ -1,45 +1,48 @@
...
 from lnst.Tests.BaseTestModule import BaseTestModule
 
 
 class TrafficControlRunner(BaseTestModule):
+    batchfiles = ListParam(type=StrParam(), required=True)
 
-    batchfile = StrParam(required=True)
-
-    def run(self) -> bool:
+    def run(self):
```
olichtne, line 12: this should have a return type defined.

lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
```diff
@@ -193,23 +189,30 @@ def _make_instances_cfgs(self) -> list[TcRunInstance]:
...
+        )
+        if len(self._finished_jobs) > 1:
+            raise MeasurementError("Too many jobs")
+        job = self._finished_jobs[0]
+        instance_data = job.result["data"]["instance_results"]
+
+        run_result.rule_install_rate = ParallelPerfResult(
+            [self._get_instance_interval(d) for d in instance_data]
+        )
+        run_result.run_success = all((d["success"] for d in instance_data))
```
olichtne, line 205: this can just be the `job.passed` i believe, you already calculate the `all()` value in the test module and that should transfer back into this attribute

olichtne (changes requested):
not sure why you picked asyncio as the way to do this since we don't use it anywhere else in lnst... but after some studying it seems to look ok.

should be fine after the minor cosmetic updates.

olichtne:
> Test jobs:

> J:7716681
> J:7713522
> J:7720505

none of these jobs actually test the parallel functionality, i assume you also have a job for that somewhere?

lnst/Tests/TrafficControl.py
pgagne: fixed

lnst/Tests/TrafficControl.py
```diff
@@ -1,45 +1,61 @@
...
-    def run_tc(self, batchfile: str) -> tuple[float, subprocess.CompletedProcess]:
-
+        instance_results = asyncio.run(self.run_instances())
+        all_success = all((i["success"] for i in instance_results))
+        self._res_data["data"] = dict(
+            instance_results=instance_results,
+        )
+        msgs = []
+        print(f"instance_result_type{type(instance_results)}")
+        print(f"instance_result[0]_type{type(instance_results[0])}")
```
olichtne, line 22: i assume these are forgotten debug prints?

lnst/Tests/TrafficControl.py
olichtne: looks good now

lnst/Tests/TrafficControl.py
pgagne: opps, good catch. fixed.

olichtne (approved):

PR #299, opened 2023-05-11
Draft: SRIOV HW OL path instead of SW path

Axonis:
### Description
Linux bridge utilizes SW path which is not
interesting for our purposes as we want to
try HW OL path, which has to use Tc to
achieve connectivity.

TODO

### Tests
J:7836800

### Reviews
@olichtne @jtluka @enhaut


Axonis:
https://github.com/docker/docker-py/issues/3113#issuecomment-1525789561 

Failing FunctionalTest

enhaut (approved):

PR #310, opened 2023-06-22
Added XDP testing capability

enhaut:
### Description

* added `Utils.kmod_loaded` function that checks whether module is loaded or not 
* PktGen test module implementation
  * PktGen is connection-less packet generator that simply "injects" packets to NIC's TX buffer. Because of that it requires `pktgen` kernel module if kernel wasn't compiled with `CONFIG_NET_PKTGEN` flag.
  * Outputs of devices are at the end parsed to `PktGen._res_data` attr in following format: `{[OUTPUT_DEV]: {"params": {DEVICE_CONFIGURATION}, "current": {CURRENT_STATS}, "state": "[ACTUAL_STATE]"}}`
* `XDPBench` test module that utilizes [xdp-bench](https://github.com/xdp-project/xdp-tools/tree/master/xdp-bench)
  * see problem with output reading https://github.com/xdp-project/xdp-tools/issues/344
* `XDPDropRecipe` is basically just `SimpleNetworkRecipe` that uses `XDPBenchMeasurement(Generator)` instead, so it uses pktgen+xdp-bench 

### Tests
* Invidivual `PktGen`/`XDPBench` test modules (examples in commit messages)
* Manual `XDPDropRecipe` testing - reviewed generator, receiver result data as well as cpu measurement results

Minimal example:
```python
from lnst.Controller import Controller
from lnst.Controller.Recipe import export_recipe_run
from lnst.Recipes.ENRT.XDPDropRecipe import XDPDropRecipe

ctl = Controller(
    debug=1,
)

recipe_instance = XDPDropRecipe(
        driver='ice',
        perf_tool_cpu=[5,6,7,8,9],
        perf_tool_cpu_policy='all',
        perf_iterations=1,
        perf_duration=60,
        multi_dev_interrupt_config={"host2":{"eth0": {"cpus":[5],"cpu_policy":"all"}}},
        ip_versions=['ipv4'],
        perf_msg_sizes=[60],
        disable_turboboost=True,
        drop_caches=True,
        offload_combinations=[{'gro': 'on', 'gso': 'on', 'tso': 'on', 'tx': 'on'}],
        perf_warmup_duration=3)
r = ctl.run(recipe_instance)

export_recipe_run(recipe_instance.runs[0], "/root/")
```
### Reviews
@LNST-project/lnst-developers

### DIFF
--- a/lnst/Common/Utils.py
+++ b/lnst/Common/Utils.py
@@ -86,6 +86,25 @@ def kmod_in_use(modulename, tries = 1):
         return kmod_in_use(modulename, tries)
     return ret
 
+
+def kmod_loaded(modulename, tries = 1):
+    tries -= 1
+    ret = False
+    mod_file = "/proc/modules"
+    handle = open(mod_file, "r")
+    for line in handle:
+        match = re.match(r'^(\S+)\s\d+\s\d+\s\S+\s(Live|Loading|Unloading).*$', line)
+        if not match or not match.groups()[0] in re.split('\s+', modulename):
+            continue
+        if match.groups()[1] == "Live":
+            ret = True
+        break
+    handle.close()
+    if (ret and tries):
+        return kmod_loaded(modulename, tries)
+    return ret
+
+
 def int_it(val):
     try:
         num = int(val)
--- a/lnst/RecipeCommon/Perf/Measurements/Results/XDPBenchMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/XDPBenchMeasurementResults.py
@@ -1,0 +1,50 @@
+from lnst.RecipeCommon.Perf.Results import ParallelPerfResult
+from lnst.RecipeCommon.Perf.Measurements.Results.FlowMeasurementResults import (
+    FlowMeasurementResults,
+)
+from lnst.RecipeCommon.Perf.Measurements.MeasurementError import MeasurementError
+
+
+class XDPBenchMeasurementResults(FlowMeasurementResults):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._generator_results = ParallelPerfResult()  # multiple instances of pktgen
+        self._receiver_results = ParallelPerfResult()  # single instance of xdpbench
+
+    def add_results(self, results):
+        if results is None:
+            return
+        if isinstance(results, XDPBenchMeasurementResults):
+            self.generator_results.append(results.generator_results)
+            self.receiver_results.append(results.receiver_results)
+        else:
+            raise MeasurementError("Adding incorrect results.")
+
+    def time_slice(self, start, end):
+        result_copy = XDPBenchMeasurementResults(
+            self.measurement, self.flow, warmup_duration=0
+        )
+
+        result_copy.generator_results = self.generator_results.time_slice(start, end)
+        result_copy.receiver_results = self.receiver_results.time_slice(start, end)
+
+        return result_copy
+
+    def describe(self):
+        generator = self.generator_results
+        receiver = self.receiver_results
+
+        desc = []
+        desc.append(str(self.flow))
+        desc.append(
+            "Generator generated: {tput:,f} {unit} per second.".format(
+                tput=generator.average, unit=generator.unit
+            )
+        )
+        desc.append(
+            "Receiver processed: {tput:,f} {unit} per second.".format(
+                tput=receiver.average, unit=receiver.unit
+            )
+        )
+
+        return desc
--- a/lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
@@ -1,0 +1,189 @@
+from lnst.RecipeCommon.Perf.Measurements.Results.AggregatedXDPBenchMeasurementResults import (
+    AggregatedXDPBenchMeasurementResults,
+)
+from lnst.Controller.RecipeResults import ResultType
+from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import (
+    Flow,
+    NetworkFlowTest,
+)
+from lnst.RecipeCommon.Perf.Measurements.MeasurementError import MeasurementError
+from lnst.RecipeCommon.Perf.Measurements.Results.XDPBenchMeasurementResults import (
+    XDPBenchMeasurementResults,
+)
+from lnst.RecipeCommon.Perf.Results import (
+    PerfInterval,
+    ParallelPerfResult,
+    SequentialPerfResult,
+)
+from lnst.Tests.PktGen import PktGen
+from lnst.Tests.XDPBench import XDPBench
+from lnst.Controller.Job import Job
+from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import BaseFlowMeasurement
+
+
+class XDPBenchMeasurement(BaseFlowMeasurement):
+    def __init__(self, flows: list[Flow], xdp_command: str, recipe_conf=None):
+        super().__init__(recipe_conf)
+        self._flows = flows
+        self._running_measurements = []
+        self._finished_measurements = []
+
+        self.command = xdp_command
+
+    def version(self):
+        return 1.0
+
+    @property
+    def flows(self):
+        return self._flows
+
+    def start(self):
+        net_flows = self._prepare_flows()
+        for flow in net_flows:
+            flow.server_job.start(bg=True)
+            flow.client_job.start(bg=True)
+            # server starts immediately, no need to wait
+            self._running_measurements.append(flow)
+
+        self._running_measurements = net_flows
+
+    def _prepare_server(self, flow: Flow):
+        params = {
+            "command": self.command,
+            "interface": flow.receiver_nic,
+            "duration": flow.duration + flow.warmup_duration * 2,
+        }
+        bench = XDPBench(**params)
+        job = flow.receiver.prepare_job(bench)
+
+        return job
+
+    def _prepare_client(self, flow: Flow):
+        params = {
+            "src_if": flow.generator_nic,
+            "dst_mac": flow.receiver_nic.hwaddr,
+            "src_ip": flow.generator_bind,
+            "dst_ip": flow.receiver_bind,
+            "cpus": flow.generator_cpupin,
+            "pkt_size": flow.msg_size,
+            "duration": flow.duration + flow.warmup_duration * 2,
+        }
+        pktgen = PktGen(**params)
+
+        job = flow.generator.prepare_job(pktgen)
+
+        return job
+
+    def _prepare_flows(self) -> list[NetworkFlowTest]:
+        flows = []
+        for flow in self.flows:
+            client = self._prepare_client(flow)
+            server = self._prepare_server(flow)
+            net_flow = NetworkFlowTest(flow, server, client)
+            flows.append(net_flow)
+
+        return flows
+
+    def finish(self):
+        try:
+            for flow in self._running_measurements:
+                client_job = flow.client_job.what
+                flow.client_job.wait(timeout=client_job.runtime_estimate())
+                flow.server_job.wait(timeout=5)
+        finally:
+            for flow in self._running_measurements:
+                flow.server_job.kill()
+                flow.client_job.kill()
+        self._finished_measurements = self._running_measurements
+        self._running_measurements = []
+
+    def collect_results(self):
+        test_flows = self._finished_measurements
+
+        results = []
+        for test_flow in test_flows:
+            flow_results = XDPBenchMeasurementResults(
+                measurement=self,
+                flow=test_flow.flow,
+                warmup_duration=test_flow.flow.warmup_duration,
+            )
+            flow_results.generator_results = self._parse_generator_results(
+                test_flow.client_job,
+            )
+            flow_results.receiver_results = self._parse_receiver_results(
+                test_flow.server_job
+            )
+
+            results.append(flow_results)
+        return results
+
+    def _parse_generator_results(self, job: Job):
+        results = ParallelPerfResult()  # container for multiple instances of pktgen
+
+        for _, raw_results in job.result.items():
+            instance_results = SequentialPerfResult()  # instance (device) of pktgen
+            for raw_result in raw_results:
+                sample = PerfInterval(
+                    raw_result["packets"],
+                    raw_result["duration"],
+                    "packets",
+                    raw_result["timestamp"],
+                )
+                instance_results.append(sample)
+            results.append(instance_results)
+
+        return results
+
+    def _parse_receiver_results(self, job: Job):
+        result = (
+            ParallelPerfResult()
+        )  # just a placeholder to keep data structure same as other Measurements
+        results = SequentialPerfResult()  # single instance of xdp-bench
+
+        for sample in job.result:
+            results.append(
+                PerfInterval(
+                    sample["rx"], sample["duration"], "packets", sample["timestamp"]
+                )
+            )
+
+        result.append(results)
+
+        return result
+
+    def _aggregate_flows(self, old_flow, new_flow):
+        if old_flow is not None and old_flow.flow is not new_flow.flow:
+            raise MeasurementError("Aggregating incompatible Flows")
+
+        new_result = AggregatedXDPBenchMeasurementResults(
+            measurement=self, flow=new_flow.flow
+        )
+
+        new_result.add_results(old_flow)
+        new_result.add_results(new_flow)
+        return new_result
+
+    @classmethod
+    def _report_flow_results(cls, recipe, flow_results):
+        generator = flow_results.generator_results
+        receiver = flow_results.receiver_results
+
+        desc = []
+        desc.extend(flow_results.describe())
+
+        recipe_result = ResultType.PASS
+        metrics = {"Generator": generator, "Receiver": receiver}
+        for name, result in metrics.items():
+            if cls._invalid_flow_duration(result):
+                recipe_result = ResultType.FAIL
+                desc.append("{} has invalid duration!".format(name))
+
+        recipe.add_result(
+            recipe_result,
+            "\n".join(desc),
+            data=dict(
+                generator_flow_data=generator,
+                receiver_flow_data=receiver,
+                flow_results=flow_results,
+            ),
+        )
--- a/lnst/Recipes/ENRT/MeasurementGenerators/XDPFlowMeasurementGenerator.py
+++ b/lnst/Recipes/ENRT/MeasurementGenerators/XDPFlowMeasurementGenerator.py
@@ -1,0 +1,31 @@
+from lnst.Common.Parameters import ChoiceParam, StrParam
+from lnst.RecipeCommon.Perf.Measurements.XDPBenchMeasurement import XDPBenchMeasurement
+from lnst.Recipes.ENRT.MeasurementGenerators.BaseFlowMeasurementGenerator import (
+    BaseFlowMeasurementGenerator,
+)
+
+from lnst.Tests.XDPBench import XDP_BENCH_COMMANDS
+
+
+class XDPFlowMeasurementGenerator(BaseFlowMeasurementGenerator):
+    xdp_command = ChoiceParam(type=StrParam, choices=XDP_BENCH_COMMANDS)
+
+    @property
+    def net_perf_tool_class(self):
+        """
+        This method uses the concept of partial application [1].
+
+        BaseFlowMeasurementGenerator.generate_perf_measurement_combinations
+        calls net_perf_tool_class and passes some arguments to the returned class.
+        However, XDPBenchMeasurement requires additional arguments, so the
+        only way to pass them is to either redefine the entire
+        generate_perf_measurement_combinations method or return a partially 
+        "initialised" class from net_perf_tool_class.
+
+        [1] https://github.com/LNST-project/lnst/pull/310#discussion_r1305763175
+        """
+        def XDPBenchMeasurement_partial(*args, **kwargs):
+            return XDPBenchMeasurement(*args, self.params.xdp_command, **kwargs)
+
+        return XDPBenchMeasurement_partial
+
--- a/lnst/Tests/PktGen.py
+++ b/lnst/Tests/PktGen.py
@@ -1,0 +1,218 @@
+import re
+import time
+import logging
+from subprocess import Popen
+from threading import Thread
+from typing import Iterator, Union
+
+from lnst.Common.Utils import kmod_loaded
+from lnst.Common.IpAddress import Ip4Address
+from lnst.Tests.BaseTestModule import BaseTestModule, TestModuleError
+from lnst.Common.Parameters import IntParam, IpParam, StrParam, ListParam, DeviceParam
+
+
+class PktGenResultsSampler:
+    def __init__(self, devs: list[str], duration: int) -> None:
+        """
+        PktGen output is just a table with current stats of devices. Therefore,
+        each device has a separate thread that captures current status of device
+        each second for `duration`.
+        """
+        self._devs = devs
+        self._duration = duration
+
+        self._sampling_threads: list[Thread] = []
+        self._raw_samples = {}
+
+    def start_sampling(self):
+        """
+        This is a separate method just to emphasize that pktgen
+        needs to be started immediately after the start of sampling.
+        """
+        self._setup_capturing()
+
+        for thread in self._sampling_threads:
+            thread.start()
+
+    def _setup_capturing(self):
+        for device in self._devs:
+            thread = Thread(target=self._read_dev_samples, args=(device,))
+            self._sampling_threads.append(thread)
+
+    def _read_dev_samples(self, device: str):
+        self._raw_samples[device] = []
+
+        for _ in range(0, self._duration + 1):  # +1 because first sample is "empty"
+            with open(f"/proc/net/pktgen/{device}", "r") as file:
+                self._raw_samples[device].append((time.time(), file.read()))
+                # TODO: ^^^ when upgrading to python interpreter without GIL check thread safety
+                # NOTE: samples are saved at it's end => the timestamp represents ending time, not the start
+            time.sleep(1)
+
+    @property
+    def device_samples(self) -> dict[str, list[dict[str, Union[float, int, dict]]]]:
+        for thread in self._sampling_threads:
+            thread.join(timeout=2)
+
+        samples = {}
+        for device in self._devs:
+            samples[device] = []
+            packets_sofar = 0
+            start_timestamp = self._raw_samples[device][0][0]  # first "empty" sample
+            # NOTE: sample's timestamp represent the end of sampling
+            # so each sample actually starts at the timestamp of previous sample
+
+            for timestamp, raw_sample in self._raw_samples[device][
+                1:
+            ]:  # ignore first empty sample
+                params, current = self._split_output(raw_sample)
+                current = self._parse_values(current)
+
+                packets = int(current["sofar"]) - packets_sofar
+
+                samples[device].append(
+                    {
+                        "timestamp": start_timestamp,
+                        "duration": timestamp - start_timestamp,
+                        "packets": packets,
+                        "errors": int(current["errors"]),
+                        "params": params,
+                    }
+                )
+                packets_sofar += packets
+                start_timestamp = timestamp
+
+        return samples
+
+    def _read_dev_outputs(self) -> Iterator[tuple[str, str]]:
+        for device in self._devs:
+            output = ""
+            with open(f"/proc/net/pktgen/{device}", "r") as f:
+                output = f.readlines()
+            yield device, "\n".join(output)
+
+    def _split_output(self, output: str) -> tuple:
+        match = re.search(r"Params:(.+)Current:(.+)Result:\s(?:\w+)", output, re.DOTALL)
+        if not match:
+            raise TestModuleError(f"Could not parse pktgen devide output: {output}")
+
+        return match.groups()
+
+    def _parse_values(self, params) -> dict[str, str]:
+        values = {}
+
+        for key, value in re.findall(r"(\w+):?\s(\S+)", params, re.MULTILINE):
+            values[key.lower()] = value
+
+        return values
+
+
+class PktGen(BaseTestModule):
+    """
+    In the scope of this module, the physical interface is refered as `interface`.
+    Pktgen device (interface@anything) is refered as device.
+
+    Inspired by https://github.com/torvalds/linux/blob/master/samples/pktgen/pktgen_sample03_burst_single_flow.sh
+    """
+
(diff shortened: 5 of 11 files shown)


lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+
+    src_mac = StrParam(mandatory=False, default="90:e2:ba:ff:ff:ff")
+    dst_mac = StrParam()
+
+    src_ip = IpParam(
+        mandatory=False, default=ipaddress("192.168.1.1")
+    )  # IP doesn't really matter
+    dst_ip = IpParam()
+
+    interface = StrParam()
```
olichtne, line 71: this could be a `DeviceParam` which would give you access to both the nic name as well as the mac address and the ip address.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+    size = IntParam(mandatory=False, default=60)  # 4 bytes are added for CRC by NIC
+    frags = IntParam(mandatory=False, default=1)
+    burst = IntParam(mandatory=False, default=8)
+
+    duration = IntParam(default=60)
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        if not kmod_in_use("pktgen"):
```
olichtne, line 83: i think this shouldn't be called in the `__init__` method, the way that `BaseTestModule` classes are designed is that they create two instances every time - one on the Controller, and one on the Agent.

The `__init__` method should therefore only really take care of initializing the common parts that are required on both ends.

Theoretically speaking, the controller doesn't need the `pktgen` module at all.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+            raise TestModuleError("pktgen module is not loaded")
+
+        self._devices = []
+        self._ip6 = True
+        if isinstance(self.params.dst_ip, Ip4Address):
+            self._ip6 = False
+
+        self._res_data = {}
+        self._output_parser = PktGenResultsParser(self._devices, self._res_data)
```
olichtne, line 92: similar issue to the pktgen module - this shouldn't be instantiated in the `__init__` method but at the start of the `run()` method which is only called on the Agent side.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+
+        self._pg_ctl("reset")
```
olichtne, line 94: same here

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+
+        self._configure_generator()
```
olichtne, line 96: and here

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+    def _pg_ctl(self, cmd: str):
+        self._write_command(self.params.pgctrl, cmd)
+
+    def _pg_thread(self, thread: int, cmd: str):
+        self._write_command(f"/proc/net/pktgen/kpktgend_{thread}", cmd)
+
+    def _pg_set(self, thread: int, cmd: str):
+        self._write_command(f"/proc/net/pktgen/{self.params.interface}@{thread}", cmd)
+
+    def _configure_generator(self):
```
olichtne, line 111: can we flip the ordering of these methods? i'd like to see the following order so that it's more natural when reading the class top to bottom:

1. __init__
2. run method
3. configure_generator
4. deconfigure_generator
5. pg_ctl
6. pg thread
7. pg set
8. write command

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+    Pktgen device (interface@anything) is refered as device.
+
+    Inspired by https://github.com/torvalds/linux/blob/master/samples/pktgen/pktgen_sample03_burst_single_flow.sh
+    """
+
+    cpus = ListParam(type=IntParam())  # each CPU is 1 generator
+
+    pgctrl = StrParam(mandatory=False, default="/proc/net/pktgen/pgctrl")
+
+    src_mac = StrParam(mandatory=False, default="90:e2:ba:ff:ff:ff")
```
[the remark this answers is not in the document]
enhaut: Yeah, i expected running it in isolated lab.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,157 @@
...
+    dst_mac = StrParam()
+
+    src_ip = IpParam(
+        mandatory=False, default=ipaddress("192.168.1.1")
+    )  # IP doesn't really matter
```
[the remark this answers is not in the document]
enhaut: PktGen is connection-less and based on that i assume, the src address is not that important. I don't expect receiver to use src IP to send replies back.

But yeah, running it in non-isolated environment might cause issues.

Fixed

lnst/Tests/PktGen.py
enhaut: Nice catch, thanks.

lnst/Tests/PktGen.py
enhaut: Thanks

lnst/Tests/PktGen.py
enhaut: Thanks

lnst/Tests/PktGen.py
enhaut: Thanks

lnst/Tests/PktGen.py
enhaut: Thanks

lnst/Tests/PktGen.py
enhaut: Reordered, thanks.

enhaut:
I've also refactored `PktGenResultsParser` class a bit.


PR should be ready.

Tested again with:
```python
>>>from lnst.Common.IpAddress import ipaddress
>>>from lnst.Devices.Device import Device
>>>class Dev(Device):
    def __init__(self, hwaddr, ips, name):
        self._hwaddr = hwaddr
        self._ips = [ipaddress(addr) for addr in ips]
        self._name = name

    @property
    def hwaddr(self):
        return self._hwaddr

    @property
    def ips(self):
        return self._ips

    @property
    def name(self):
        return self._name
>>> from lnst.Tests.PktGen import PktGen as P
>>> src = Dev("b4:83:51:01:e9:b8", ["192.168.1.1", "10.0.0.1"], "eno12399")
>>> dst = Dev("b4:83:51:01:e8:cc", ["192.168.1.2", "10.0.0.2"], "eno12399")
>>> g = P(cpus=[i + 1 for i in range(0, 4)], src_if=src, dst_if=dst, duration=10)
>>> g.run()
True
>>> g._res_data
{'eno12399@1': {'params': {'count': '0', 'min_pkt_size': '60', 'max_pkt_size': '60', 'frags': '0', 'delay': '0', 'clone_skb': '0', 'ifname': 'eno12399@1', 'flows': '0', 'flowlen': '0', 'queue_map_min': '0', 'queue_map_max': '0', 'dst_min': '192.168.1.2', 'dst_max': '192.168.1.2', 'src_min': '192.168.1.1', 'src_max': '192.168.1.1', 'src_mac': 'b4:83:51:FF:FF:FF', 'dst_mac': 'b4:83:51:FF:FF:FE', 'udp_src_min': '9', 'udp_src_max': '9', 'udp_dst_min': '9', 'udp_dst_max': '9', 'src_mac_count': '0', 'dst_mac_count': '0', 'burst': '8', 'flags': 'NO_TIMESTAMP'}, 'current': {'sofar': '76146381', 'errors': '0', 'started': '13250135028us', 'stopped': '13260132321us', 'idle': '14us', 'seq_num': '76146382', 'cur_dst_mac_offset': '0', 'cur_src_mac_offset': '0', 'cur_saddr': '192.168.1.1', 'cur_daddr': '192.168.1.2', 'cur_udp_dst': '9', 'cur_udp_src': '9', 'cur_queue_map': '1', 'flows': '0'}, 'result': {'duration': 9.997292, 'packets': 76146381, 'frags': 0, 'errors': 0}}, 'eno12399@2': {'params': {'count': '0', 'min_pkt_size': '60', 'max_pkt_size': '60', 'frags': '0', 'delay': '0', 'clone_skb': '0', 'ifname': 'eno12399@2', 'flows': '0', 'flowlen': '0', 'queue_map_min': '0', 'queue_map_max': '0', 'dst_min': '192.168.1.2', 'dst_max': '192.168.1.2', 'src_min': '192.168.1.1', 'src_max': '192.168.1.1', 'src_mac': 'b4:83:51:FF:FF:FF', 'dst_mac': 'b4:83:51:FF:FF:FE', 'udp_src_min': '9', 'udp_src_max': '9', 'udp_dst_min': '9', 'udp_dst_max': '9', 'src_mac_count': '0', 'dst_mac_count': '0', 'burst': '8', 'flags': 'NO_TIMESTAMP'}, 'current': {'sofar': '147689124', 'errors': '0', 'started': '13250135034us', 'stopped': '13260132320us', 'idle': '1125us', 'seq_num': '147689125', 'cur_dst_mac_offset': '0', 'cur_src_mac_offset': '0', 'cur_saddr': '192.168.1.1', 'cur_daddr': '192.168.1.2', 'cur_udp_dst': '9', 'cur_udp_src': '9', 'cur_queue_map': '2', 'flows': '0'}, 'result': {'duration': 9.997286, 'packets': 147689124, 'frags': 0, 'errors': 0}}, 'eno12399@3': {'params': {'count': '0', 'min_pkt_size': '60', 'max_pkt_size': '60', 'frags': '0', 'delay': '0', 'clone_skb': '0', 'ifname': 'eno12399@3', 'flows': '0', 'flowlen': '0', 'queue_map_min': '0', 'queue_map_max': '0', 'dst_min': '192.168.1.2', 'dst_max': '192.168.1.2', 'src_min': '192.168.1.1', 'src_max': '192.168.1.1', 'src_mac': 'b4:83:51:FF:FF:FF', 'dst_mac': 'b4:83:51:FF:FF:FE', 'udp_src_min': '9', 'udp_src_max': '9', 'udp_dst_min': '9', 'udp_dst_max': '9', 'src_mac_count': '0', 'dst_mac_count': '0', 'burst': '8', 'flags': 'NO_TIMESTAMP'}, 'current': {'sofar': '75560659', 'errors': '0', 'started': '13250135028us', 'stopped': '13260132320us', 'idle': '12us', 'seq_num': '75560660', 'cur_dst_mac_offset': '0', 'cur_src_mac_offset': '0', 'cur_saddr': '192.168.1.1', 'cur_daddr': '192.168.1.2', 'cur_udp_dst': '9', 'cur_udp_src': '9', 'cur_queue_map': '3', 'flows': '0'}, 'result': {'duration': 9.997291, 'packets': 75560659, 'frags': 0, 'errors': 0}}, 'eno12399@4': {'params': {'count': '0', 'min_pkt_size': '60', 'max_pkt_size': '60', 'frags': '0', 'delay': '0', 'clone_skb': '0', 'ifname': 'eno12399@4', 'flows': '0', 'flowlen': '0', 'queue_map_min': '0', 'queue_map_max': '0', 'dst_min': '192.168.1.2', 'dst_max': '192.168.1.2', 'src_min': '192.168.1.1', 'src_max': '192.168.1.1', 'src_mac': 'b4:83:51:FF:FF:FF', 'dst_mac': 'b4:83:51:FF:FF:FE', 'udp_src_min': '9', 'udp_src_max': '9', 'udp_dst_min': '9', 'udp_dst_max': '9', 'src_mac_count': '0', 'dst_mac_count': '0', 'burst': '8', 'flags': 'NO_TIMESTAMP'}, 'current': {'sofar': '54989194', 'errors': '0', 'started': '13250135033us', 'stopped': '13260132323us', 'idle': '6us', 'seq_num': '54989195', 'cur_dst_mac_offset': '0', 'cur_src_mac_offset': '0', 'cur_saddr': '192.168.1.1', 'cur_daddr': '192.168.1.2', 'cur_udp_dst': '9', 'cur_udp_src': '9', 'cur_queue_map': '4', 'flows': '0'}, 'result': {'duration': 9.997289, 'packets': 54989194, 'frags': 0, 'errors': 0}}}
```

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,180 @@
...
+
+    duration = IntParam(default=60)
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        self._devices = []
+        self._ip6 = True
+        if isinstance(self.params.dst_if.ips[0], Ip4Address):
+            self._ip6 = False
```
olichtne, line 104: i'd also move this to the `configure` method... and probably make it just a local variable, this isn't needed to be an attribute of the class instance it seems.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,180 @@
...
+        self._deconfigure_generator()
+        return True
+
+    def _configure_generator(self):
+        logging.debug("Configuring generator")
+        src = f"src{6 if self._ip6 else ''}"
+        dest = f"dst{6 if self._ip6 else ''}"
+
+        src_ip = self.params.src_if.ips[0]  # xdp uses just 1 IP
+        dst_ip = self.params.dst_if.ips[0]
```
olichtne, line 138: i'm not sure this `[0]` selection will work with our enrt recipes well. in all of them we always configure both the ipv4 as well as an ipv6 address. Their order MAY (though it shouldn't) switch non deterministically... and... both will always be present, so maybe we should write this in a way that doesn't assume that the correct ip address is in the correct place all of the time.

enhaut: not relevant any more

lnst/Tests/PktGen.py
enhaut: done

lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
```diff
@@ -0,0 +1,180 @@
...
+        new_result.add_results(new_flow)
+        return new_result
+
+    @classmethod
+    def _report_flow_results(cls, recipe, flow_results):
+        generator = flow_results.generator_results
+        receiver = flow_results.receiver_results
+       
+        desc = []
+        desc.append(str(flow_results.flow))
```
olichtne, line 161: for the other measurement classes we've moved the generation of the description into the `Result` classes: https://github.com/LNST-project/lnst/blob/master/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py#L107

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
```diff
@@ -19,6 +19,7 @@
 MEASUREMENT_LOOKUP = {
     'iperf': IperfFlowMeasurement,
     'neper': NeperFlowMeasurement,
+    'xdp-bench':XDPBenchMeasurement,
```
olichtne, line 22: so since this class REQUIRES the use of an `xdp_command` it probably shouldn't be listed here, but only available in the derived class - if you include it here, then all test recipes that inherit from here gain the ability to specify the use of the xdp-bench tool as a measurement tool and since they don't define the parameter the instance init will crash.

olichtne (changes requested):

lnst/RecipeCommon/Perf/Measurements/Results/XDPBenchMeasurementResults.py
```diff
@@ -0,0 +1,30 @@
+from lnst.RecipeCommon.Perf.Results import ParallelPerfResult
+from lnst.RecipeCommon.Perf.Measurements.Results.FlowMeasurementResults import FlowMeasurementResults
+from lnst.RecipeCommon.Perf.Measurements.MeasurementError import MeasurementError
+
+
+class XDPBenchMeasurementResults(FlowMeasurementResults):
```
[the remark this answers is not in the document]
enhaut: > Could we just reuse the normal `FlowMeasurementResults` class and simply report 0 for the `cpu_stats`?

We cannot, methods that handle warm-up/down slicing and aggregating expect `cpu_stats` to be an instance of `PerfInterval` with non-zero duration (because of division and slicing).
Sure, we can put there `PerfInterval` with duration set to `1`, value `0` and timestamps to neutral values in regard to alignment/warm-up/down slicing, but I don't like the idea of having this kind of "workarounds" in production code.

> The reason being is that i think this doesn't actually contribute any new code, just... masks out the unused cpu stats.

For now, it just masks out cpu stats but in the future, there might be also another metrics e.g. for xdp forward recipe, and we'll end up with separate results anyway.

lnst/Tests/XDPBench.py
```diff
@@ -0,0 +1,139 @@
...
+    def parse_output(self):
+        stdout, stderr = self._process.communicate()
+
+        logging.debug("Stderr of xdp-bench:")
+        logging.debug(str(stderr))
+
+        self._parse_raw_results(stdout)
+
+        if not self._out:
+            raise TestModuleError(f"Could not parse xdp-bench output: {stdout}")
```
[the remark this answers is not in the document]
enhaut: Exactly, this method does not return anything. Since (almost) everything in python is stored as a pointer/reference, it doesn't need to.

`self._out` is just a reference to `XDPBench._res_data`, see https://github.com/LNST-project/lnst/pull/310/files#diff-fab965ef6bb50e158a3f0486357589d9be87461c9424e4efef2212b705108213R107

But I'll rewrite it, this might be kinda confusing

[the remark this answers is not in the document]
olichtne: ah... so the TestModule provides its own `res_data` into the parser and the parser fills it out... yeah that's a confusing way to do that i'd say.

lnst/RecipeCommon/Perf/Measurements/Results/XDPBenchMeasurementResults.py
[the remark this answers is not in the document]
enhaut: > well i'd say duration should reflect the actual duration in this case... and technically that's still correct

I don't agree, it's not correct. xdp rules are utilizing CPU, but we cannot easily measure how. Setting `receiver.cpu_stats` to `0` would imply that xdp rules are not basically do anything, which you can't conclude from results with `receiver.cpu_stats` being undefined/`None`. 

> xdp flow measurement could add support for the cpu utilization

Discussed offline, since XDP is running as part of kernel instead of a separate process, there is no simple way to measure utilization, that wouldn't significantly affect performance.

We can gather running time statistics via `bpftool prog` with `kernel.bpf_stats_enabled` enabled however, from my experiments it seems that the amount of dropped packets dropped by 30% along with CPU usage being 20% higher.

lnst/Tests/XDPBench.py
```diff
@@ -0,0 +1,139 @@
...
+    def _parse_line(self, line: str):
+        match = re.search(r"Summary\s+([\d,]+)\srx/s\s+([\d,]+)\serr/s?", line)
+
+        if not match:  # skip summary line at the end + corrupted lines
+            logging.error(f"Could not parse xdp-bench output: {line}")
+            return
+
+        rx = match.group(1).replace(",", "")
+        err = match.group(2).replace(",", "")
+        # ^^ remove thousands separators
```
jtluka, line 49: Can't this be handled in the regex somehow?

lnst/Tests/XDPBench.py
```diff
@@ -0,0 +1,139 @@
...
+        type=StrParam,
+        choices=XDP_BENCH_COMMANDS,
+        mandatory=True,
+    )
+    interface = DeviceParam(mandatory=True)
+    interface2 = DeviceParam()  # used for redirect modes
+
+    interval = IntParam(default=1)
+
+    redirect_device = DeviceParam()
```
jtluka, line 84: Isn't this the same as `interface2` as it mentions redirect?

lnst/Common/Parameters.py
```diff
@@ -51,6 +51,10 @@ def type_check(self, value):
         """
         return value
 
+class ConstParam(Param):
```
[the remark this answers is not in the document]
olichtne: https://github.com/LNST-project/lnst/pull/311 was merged so please rebase on top of that to replace the commit that adds the ConstParam.

jtluka (dismissed):
Left some comments. I'm leaving for PTO, so I can't provide more feedback now. Overall looks good.

lnst/Tests/XDPBench.py
enhaut: Well, it might be, but having it as a 2 separate params makes the implementation of bash command composing much simpler. `interface2` param is used for `xdp-bench` modes that doesn't require argument name e.g. `./xdp-bench redirect-map ingress_inf egress_inf`. However mode `redirect-cpu` requires argument `--redirect-device inf`.

I didn't want to dig into command composing based on further context, see https://github.com/LNST-project/lnst/blob/0a0afe2fe0ed24144b8101b00089234f44d95dca/lnst/Tests/XDPBench.py#L120  so i decided to include both arguments there.

lnst/Tests/XDPBench.py
enhaut: Nope, regex cannot ignore characters within matching group

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,185 @@
...
+
+        for cpu in self.params.cpus:
+            dev = f"{self.params.src_if.name}@{cpu}"
+            logging.debug(f"Adding interface {self.params.src_if.name} to cpu {cpu}")
+
+            self._pg_thread(cpu, f"add_device {dev}")
+
+            self._pg_set(cpu, f"flag QUEUE_MAP_CPU")
+            self._pg_set(cpu, f"count {self.params.count}")
+            self._pg_set(cpu, f"pkt_size {self.params.size}")
```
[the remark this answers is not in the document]
enhaut: Hmm, what do you mean by parameters?

`flag`, `count`, `pkt_size` have the same names here as in the pktgen, it won't work otherwise.

The methods are also called the same as in bash version of pktgen, see https://github.com/torvalds/linux/blob/master/samples/pktgen/functions.sh .

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,185 @@
...
+        super().__init__(**kwargs)
+
+        self._devices = []
+
+        self._res_data = {}
+        self._output_parser = None
+
+    def run(self):
+        if not kmod_in_use("pktgen"):
+            raise TestModuleError("pktgen module is not loaded")
```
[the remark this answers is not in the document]
enhaut: Actually, we can't. The `TestModule.__init__()` method is also executed on the controller, which generally doesn't have to be the same machine as the generator (which executes the `TestModule.run()` method), so if we have 3 machines setup (controller, generator, receiver) it would crash.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,185 @@
...
+    src_if = DeviceParam()
+    dst_mac = StrParam()
+
+    src_ip = IpParam()
+    dst_ip = IpParam()
+
+    count = IntParam(mandatory=False, default=0)  # 0 = no upper limit
+    size = IntParam(mandatory=False, default=60)  # 4 bytes are added for CRC by NIC
+    frags = IntParam(mandatory=False, default=1)
+    burst = IntParam(mandatory=False, default=8)
```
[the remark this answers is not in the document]
enhaut: Nice catch, thanks.

lnst/Common/Utils.py
```diff
@@ -75,10 +75,10 @@ def kmod_in_use(modulename, tries = 1):
     mod_file = "/proc/modules"
     handle = open(mod_file, "r")
     for line in handle:
-        match = re.match(r'^(\S+)\s\d+\s(\d+).*$', line)
```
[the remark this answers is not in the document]
enhaut: Oh, I see, thanks.

I've moved needed functionality to new function `kmod_loaded()` since `kmod_in_use()` might be useful.

lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
enhaut: thanks, missed that, moved.

lnst/Tests/PktGen.py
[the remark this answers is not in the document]
olichtne: i think he means the rename of `self.params.size` to `self.params.pkt_size`

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,224 @@
...
+            thread = Thread(target=self._read_dev_samples, args=(device,))
+            self._sampling_threads.append(
+                (device, thread)
+            )  # start of sampling timestamp is used for duration calculation of 1st sample
+
+    def _read_dev_samples(self, device: str):
+        self._raw_samples[device] = []
+
+        for _ in range(0, self._duration):
+            time.sleep(1)
```
olichtne, line 52: wouldn't this make more sense to be AFTER the proc file read? that way each thread immediately initializes with a raw sample which is basically "all zeros" and also stores the initial timestamp effectively removing the need for the `self._threads_start` variable and therefore also removing the conditional in the `parse_dev_outputs`?

lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
```diff
@@ -0,0 +1,189 @@
...
+            # server starts immediately, no need to wait
+            self._running_measurements.append(flow)
+
+        self._running_measurements = net_flows
+
+    def _prepare_server(self, flow: Flow):
+        params = {
+            "command": self.command,
+            "interface": flow.receiver_nic,
+            "duration": flow.duration + flow.warmup_duration * 2,
```
olichtne, line 54: curious... why are we multiplying the warmup duration by two?

lnst/Recipes/ENRT/MeasurementGenerators/XDPFlowMeasurementGenerator.py
```diff
@@ -0,0 +1,30 @@
...
+)
+
+from lnst.Tests.XDPBench import XDP_BENCH_COMMANDS
+
+
+class XDPFlowMeasurementGenerator(BaseFlowMeasurementGenerator):
+    xdp_command = ChoiceParam(type=StrParam, choices=XDP_BENCH_COMMANDS)
+
+    @property
+    def net_perf_tool_class(self):
```
olichtne, line 14: so i thought of a ... python magic thing for this:

```python
@property
def net_perf_tool_class(self):
    def XDPBenchMeasurement_partial(*args):
        return XDPBenchMeasurement(*args, self.params.xdp_command)
    return XDPBencMEasurement_partial
```

what this does is it **binds** one of the parameters of the class initializer to a specific value and returns a "new initializer" which takes one less initializer.

This is a concept from functional programming of 'partial application' : https://en.wikipedia.org/wiki/Partial_application

aaaand python actually has a thing for this: https://docs.python.org/3/library/functools.html?highlight=partial#functools.partial

so my previous example maybe can be simplified with that if it works with classes... but maybe it doesn't need to be.

why i mention this... it would basically remove the need for you to reimplement the `generate_perf_measurements_combinations` method just because of this one parameter and so you wouldn't have any issues with the super calls or breaking any sort of collaborative inheritance.

though... this would probably be best to document in a docstring for anyone else in the future looking at this.

olichtne (changes requested):
left a couple of comments, looks mostly good i think this is almost mergeable after resolving the individual discussions.

lnst/Tests/PktGen.py
enhaut: Hmm, thanks for pointing me in the right direction. That's a better solution.

And I also realized that I was actually saving **end** timestamp of sample instead of **start** one that is required due to time alignment/warm-up trimming, fixed.

Thanks.

lnst/Tests/XDPBench.py
```diff
@@ -0,0 +1,169 @@
...
+        self._raw_samples = []
+        self._capturing_start = 0
+
+    def start_sampling(self):
+        thread = Thread(target=self._capture_output)
+        thread.start()
+        self._capturing_start = time.time()
+
+    def stop_sampling(self):
+        # TODO:
```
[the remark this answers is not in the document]
enhaut: Kinda, that's just a leftover that I planned to do, however, there is no simple way of stopping python's thread so sampling is stopped by catching an exception in https://github.com/LNST-project/lnst/blob/5fbcf8ee4716447c6566c0a767bfd03cbf54bf8c/lnst/Tests/XDPBench.py#L38

Removed, thanks.

lnst/RecipeCommon/Perf/Measurements/XDPBenchMeasurement.py
enhaut: Current implementation of trimming cuts `warmup_duration` seconds from the start and `warmup_duration` seconds from the end of the results. So we need to extend measuring time.

See 
https://github.com/LNST-project/lnst/blob/9b220d8a13097e51e6b5ae3d533fa1fe7fc239e6/lnst/RecipeCommon/Perf/Recipe.py#L106
and `Recipe.time_aligned_results`

lnst/Tests/XDPBench.py
jtluka: Ok, if the tool takes this as separate arguments, then that's ok.

lnst/Tests/PktGen.py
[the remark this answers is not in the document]
jtluka: Yes, I meant rename of `self.params.size` to `self.params.pkt_size`.

lnst/Tests/PktGen.py
```diff
@@ -0,0 +1,217 @@
...
+        self._output_parser = None
+
+    def run(self):
+        if not kmod_loaded("pktgen"):
+            raise TestModuleError("pktgen module is not loaded")
+
+        self._pg_ctrl("reset")
+        self._configure_generator()
+
+        output_parser = PktGenResultsParser(self._devices, self.params.duration)
```
jtluka, line 147: I got confused by the class name here. The name `PktGenResultsParser` evokes that this should just parse the results (which it does), however the next line says `start_sampling`. Could we just rename it to `PktGenResultsSampler`?

Then it would look like:
```
sampler = PktGenResultsSampler(self._devices, self.params.duration)
sampler.start_sampling()
...
self._res_data = sampler.parse_dev_outputs()
```

The change above would make this more readable imo.

Additionally when I look at the proposed code above, can't it be simply:
```
self._res_data = sampler.dev_samples # rename + switch method to property? that would call _parse_dev_outputs() as a private method?
```

What do you think?

lnst/Tests/PktGen.py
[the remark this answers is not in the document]
enhaut: Sure, done

lnst/Recipes/ENRT/MeasurementGenerators/XDPFlowMeasurementGenerator.py
enhaut: That's a nice idea, thanks.

lnst/Tests/PktGen.py
enhaut: Sure, thanks

olichtne (approved):
looks ok to me now.

PR #311, opened 2023-06-25
Add RDMA bandwidth testing capability

Kuba314:
### Description
Adds RDMA (Remote Direct Memory Access) testing using [linux-rdma/perftest](https://github.com/linux-rdma/perftest)'s program `ib_send_bw`. In order for this measurement to work, we have to have an RDMA-capable pair of devices on two hosts. These are currently setup in software using `rdma link add`. There's also an option for hardware RDMA, which requires RDMA-capable NICs. Software RDMA can run on any NIC supporting Ethernet.

Currently, there are two software methods that enable RDMA - `rxe` (Soft-RoCE (Software RDMA over Converged Ethernet)) and `siw` (Software iWARP). One of these has to be selected in the recipe parameters.

### Tests
J:8103516

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
@@ -1,0 +1,141 @@
+from typing import Any, Optional
+import time
+
+from lnst.Controller.Job import Job
+from lnst.Controller.Recipe import BaseRecipe
+from lnst.Controller.RecipeResults import ResultType
+from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import Flow, NetworkFlowTest
+from lnst.RecipeCommon.Perf.Results import PerfInterval
+from lnst.Tests.RDMABandwidth import RDMABandwidthServer, RDMABandwidthClient
+from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import BaseFlowMeasurement
+from lnst.RecipeCommon.Perf.Measurements.Results import (
+    RDMABandwidthMeasurementResults,
+    AggregatedRDMABandwidthMeasurementResults
+)
+
+
+class RDMABandwidthMeasurement(BaseFlowMeasurement):
+    def __init__(
+        self,
+        flows: list[Flow],
+        recipe_conf: Any = None,
+    ):
+        super().__init__(recipe_conf)
+
+        self._flows = flows
+        self._endpoint_tests: list[NetworkFlowTest] = []
+
+    @property
+    def flows(self) -> list[Flow]:
+        return self._flows
+
+    def start(self) -> None:
+        self._endpoint_tests.extend(self._prepare_endpoint_tests())
+
+        for endpoint_test in self._endpoint_tests:
+            endpoint_test.server_job.start(bg=True)
+
+        time.sleep(2)
+
+        self._start_timestamp = time.time()
+        for endpoint_test in self._endpoint_tests:
+            endpoint_test.client_job.start(bg=True)
+
+    def finish(self) -> None:
+        try:
+            for endpoint_test in self._endpoint_tests:
+                timeout = endpoint_test.flow.duration + 5
+                endpoint_test.client_job.wait(timeout=timeout)
+                endpoint_test.server_job.wait(timeout=timeout)
+        finally:
+            for endpoint_test in self._endpoint_tests:
+                endpoint_test.client_job.kill()
+                endpoint_test.server_job.kill()
+
+    def collect_results(self) -> list[RDMABandwidthMeasurementResults]:
+        results: list[RDMABandwidthMeasurementResults] = []
+        for endpoint_test in self._endpoint_tests:
+            bandwidth = endpoint_test.client_job.result["bandwidth"]
+            duration = endpoint_test.flow.duration
+            result = RDMABandwidthMeasurementResults(
+                measurement=self,
+                flow=endpoint_test.flow,
+            )
+            result.bandwidth = PerfInterval(
+                # pre-multiply to get the total amount of bytes sent
+                value=bandwidth * duration,
+                duration=duration,
+                unit="MiB",
+                timestamp=self._start_timestamp,
+            )
+            results.append(result)
+        self._endpoint_tests.clear()
+        return results
+
+    def _prepare_endpoint_tests(self) -> list[NetworkFlowTest]:
+        return [
+            NetworkFlowTest(
+                flow=flow,
+                server_job=self._prepare_server_job(flow),
+                client_job=self._prepare_client_job(flow),
+            )
+            for flow in self.flows
+        ]
+
+    def _prepare_server_job(self, flow: Flow) -> Job:
+        params = {
+            "device_name": self.recipe_conf.rdma_device_name,
+            "duration": flow.duration,
+            "port": flow.receiver_port,
+            "size": flow.msg_size,
+        }
+        if flow.receiver_cpupin is not None:
+            params["cpu_bind"] = flow.receiver_cpupin
+
+        return flow.receiver.prepare_job(RDMABandwidthServer(**params))
+
+    def _prepare_client_job(self, flow: Flow) -> Job:
+        params = {
+            "device_name": self.recipe_conf.rdma_device_name,
+            "dst_ip": flow.receiver_bind,
+            "src_ip": flow.generator_bind,
+            "duration": flow.duration,
+            "port": flow.receiver_port,
+            "size": flow.msg_size,
+        }
+        if flow.generator_cpupin is not None:
+            params["cpu_bind"] = flow.generator_cpupin
+
+        if flow.warmup_duration > 0:
+            params["perform_warmup"] = True
+
+        return flow.generator.prepare_job(RDMABandwidthClient(**params))
+
+    @classmethod
+    def aggregate_results(
+        cls,
+        old: Optional[list[AggregatedRDMABandwidthMeasurementResults]],
+        new: list[RDMABandwidthMeasurementResults],
+    ) -> list[AggregatedRDMABandwidthMeasurementResults]:
+        if old is None:
+            agg_results = []
+            for result in new:
+                agg_result = AggregatedRDMABandwidthMeasurementResults(result.measurement, result.flow)
+                agg_result.add_results(result)
+                agg_results.append(agg_result)
+            return agg_results
+
+        aggregated: list[AggregatedRDMABandwidthMeasurementResults] = []
+        for old_measurements, new_measurement in zip(old, new):
+            old_measurements.add_results(new_measurement)
+            aggregated.append(old_measurements)
+        return aggregated
+
+    @classmethod
+    def report_results(
+        cls,
+        recipe: BaseRecipe,
+        aggregated_results: list[AggregatedRDMABandwidthMeasurementResults],
+    ) -> None:
+        for aggregated_result in aggregated_results:
+            recipe.add_result(ResultType.PASS, aggregated_result.describe())
--- a/lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
@@ -1,0 +1,20 @@
+from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
+from lnst.RecipeCommon.Perf.Measurements.Results.RDMABandwidthMeasurementResults import RDMABandwidthMeasurementResults
+from lnst.RecipeCommon.Perf.Results import SequentialPerfResult
+
+
+class AggregatedRDMABandwidthMeasurementResults(RDMABandwidthMeasurementResults):
+    def __init__(self, measurement: BaseMeasurement, flow: "Flow"):
+        super().__init__(measurement, flow)
+        self._individual_results = []
+
+    @property
+    def individual_results(self) -> list[RDMABandwidthMeasurementResults]:
+        return self._individual_results
+
+    @property
+    def bandwidth(self) -> SequentialPerfResult:
+        return SequentialPerfResult([result.bandwidth for result in self.individual_results])
+
+    def add_results(self, result: RDMABandwidthMeasurementResults) -> None:
+        self._individual_results.append(result)
--- a/lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
@@ -1,0 +1,50 @@
+from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
+from lnst.RecipeCommon.Perf.Measurements.Results.BaseMeasurementResults import BaseMeasurementResults
+from lnst.RecipeCommon.Perf.Results import PerfInterval, PerfResult
+
+
+class RDMABandwidthMeasurementResults(BaseMeasurementResults):
+    def __init__(self, measurement: BaseMeasurement, flow: "Flow"):
+        super().__init__(measurement)
+
+        self._flow = flow
+
+    @property
+    def flow(self):
+        return self._flow
+
+    @property
+    def bandwidth(self) -> PerfInterval:
+        return self._bandwidth
+
+    @bandwidth.setter
+    def bandwidth(self, bandwidth: PerfInterval) -> None:
+        self._bandwidth = bandwidth
+
+    @property
+    def start_timestamp(self):
+        return self._bandwidth.start_timestamp
+
+    @property
+    def end_timestamp(self):
+        return self._bandwidth.end_timestamp
+
+    def time_slice(self, start, end):
+        result_copy = RDMABandwidthMeasurementResults(self.measurement, self.flow)
+        result_copy.bandwidth = self.bandwidth.time_slice(start, end)
+        return result_copy
+
+    def describe(self) -> str:
+        bw = self.bandwidth
+        return "ib_send_bw measured bandwidth: {avg:.2f} +-{stddev:.2f}({percentage:.2f}%) MiB/s.".format(
+            avg=bw.average,
+            stddev=bw.std_deviation,
+            percentage=self._deviation_percentage(bw),
+        )
+
+    @classmethod
+    def _deviation_percentage(cls, result: PerfResult) -> float:
+        try:
+            return (result.std_deviation/result.average) * 100
+        except ZeroDivisionError:
+            return float('inf') if result.std_deviation >= 0 else float("-inf")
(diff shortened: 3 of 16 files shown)


lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
```diff
@@ -0,0 +1,135 @@
...
+        for flow_test in self._running_flow_tests:
+            flow_test.client_job.start(bg=True)
+
+    def finish(self) -> None:
+        for flow_test in self._running_flow_tests:
+            flow_test.client_job.wait()
+            flow_test.server_job.wait()
+
+        self._finished_flow_tests = self._running_flow_tests
+        self._running_flow_tests = []
```
jtluka, line 68: I don't undersrand what the `self._running_flow_tests` and `self._finished_flow_tests` are for.

Even the initialization of `self._running_flow_tests` does not quite match the name variable. The tests are not running yet.

With the current code it would be sufficient to have just `self._flow_tests`. Or rename `running_tests` to `self._flow_tests`  and have `self._finished_tests`. But still, there's no use of the variable other than tracking ALL of the flow_tests.

lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def collect_results(self) -> list[RDMABandwidthMeasurementResults]:
+        results: list[RDMABandwidthMeasurementResults] = []
+        for flow_test in self._finished_flow_tests:
+            results.append(
+                RDMABandwidthMeasurementResults(
+                    measurement=self,
+                    bandwidth=flow_test.client_job.result["bandwidth"],
```
jtluka, line 76: Maybe we should include also the server_job results here. But I'm not familiar with the tool and do not know if it generates anything.

lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
```diff
@@ -0,0 +1,21 @@
+
+
```
jtluka, line 2: Please remove these empty lines.

lnst/Recipes/ENRT/RDMARecipe.py
```diff
@@ -0,0 +1,79 @@
...
+from lnst.RecipeCommon.Perf.Measurements.RDMABandwidthMeasurement import RDMAFlow
+from lnst.RecipeCommon.Perf.Recipe import RecipeConf, Recipe as PerfRecipe
+
+
+class RDMARecipe(PerfRecipe):
+    """
+    RDMA / InfiniBand bandwidth test recipe
+    """
+
+    driver = StrParam(default="mlx5_core")
```
jtluka, line 16: Maybe change this to not have default and have `mandatory=True` to force user to specify it.

lnst/Recipes/ENRT/RDMARecipe.py
```diff
@@ -0,0 +1,79 @@
...
+
+    host2 = HostReq()
+    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))
+
+    def test(self) -> None:
+        config = self.test_wide_configuration()
+        try:
+            results = self.perf_test(config)
+            self.perf_report(results)
+            # self.perf_evaluate(results)
```
jtluka, line 32: Do you need to keep this commented out?

lnst/Recipes/ENRT/RDMARecipe.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_wide_configuration(self) -> RecipeConf:
+        host1, host2 = self.matched.host1, self.matched.host2
+
+        ipv4_addr_pool = interface_addresses(self.params.net_ipv4)
+        for host in [host1, host2]:
+            host.eth0.ip_add(next(ipv4_addr_pool))
+            host.eth0.up()
+
+            # setup RoCE (RDMA over Converged Ethernet), emulating InfiniBand on Ethernet
+            host.run("ip a")
```
jtluka, line 45: Debugging statement?

lnst/Recipes/ENRT/RDMARecipe.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+    def test_wide_deconfiguration(self, config: RecipeConf) -> None:
+        host1, host2 = self.matched.host1, self.matched.host2
+
+        for host in [host1, host2]:
+            host.run(f"rdma link delete ib0")
+
+    def apply_perf_test_tweak(self, config: RecipeConf) -> None:
+        pass
```
jtluka, line 73: Does these tweak bypasses need to be specified?

jtluka (changes requested):
Otherwise looks ok. I've left some comments and questions.

lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
Kuba314: It was because the list doesn't get cleared on more perf iterations. I clear the list now and moved to using a single list.

lnst/RecipeCommon/Perf/Measurements/RDMABandwidthMeasurement.py
Kuba314: server result == client result == single float bandwidth `MiB/s` result.

lnst/Recipes/ENRT/RDMARecipe.py
Kuba314: Now inheriting from `SimpleNetworkRecipe`. This is not relevant anymore.

lnst/Recipes/ENRT/RDMARecipe.py
Kuba314: Yes. Removed it.

lnst/Recipes/ENRT/RDMARecipe.py
Kuba314: Not sure, but now I'm inheriting from `SimpleNetworkRecipe`. This is not relevant anymore.

lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
Kuba314: Fixed.

lnst/Recipes/ENRT/RDMARecipe.py
Kuba314: Not relevant anymore.

lnst/Controller/Recipe.py
```diff
@@ -106,6 +106,8 @@ def __init__(self, **kwargs):
         params = ((name, val) for name, val in attrs.items() if isinstance(val, Param))
         for name, val in params:
             if name in kwargs:
+                if isinstance(val, ConstParam):
+                    raise RecipeError(f"Value for constant parameter {name} was provided")
```
olichtne, line 110: imo this exception should be a `ParamError` raised from the `validate` method of the `ConstParam` class and should check if the provided value matches the required constant, otherwise raise the exception.

this would:
1. keep responsibility for ConstParam inside the ConstParam class instead of extending the Recipe class
2. make it easier to use in automation where we simply expand `**params` dynamically to generate the recipe classes - we wouldn't have to implement a `if "param" in params: del params["param"]` for situations where the values are actually ok.

Kuba314: 1. sure, makes sense, but
2. I thought you wanted to raise an exception if a `ConstParam` argument was provided?

lnst/Tests/RDMABandwidth.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def run(self) -> bool:
+        self._res_data = {}
+        if not is_installed("ib_send_bw"):
+            self._res_data["msg"] = "perftest is not installed on this machine!"
+            logging.error(self._res_data["msg"])
+            return False
+
+        command = self._compose_cmd()
+        print(command)
```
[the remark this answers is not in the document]
Kuba314: Oops, forgot I put that there. Do we want to keep it as a `logging.debug` call? I guess why won't we want that, right?

lnst/Recipes/ENRT/MeasurementGenerators/RDMABandwidthMeasurementGenerator.py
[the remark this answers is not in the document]
Kuba314: This obviously shouldn't be here. I'll remove it.

lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
```diff
@@ -0,0 +1,31 @@
+from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
+from lnst.RecipeCommon.Perf.Measurements.Results.BaseMeasurementResults import BaseMeasurementResults
+from lnst.RecipeCommon.Perf.Results import PerfResult
+
+
+class RDMABandwidthMeasurementResults(BaseMeasurementResults):
+    def __init__(self, measurement: BaseMeasurement, flow, bandwidth: PerfResult):
```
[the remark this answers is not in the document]
Kuba314: Yes, because I couldn't import it because of a cyclic dependency. I could work around this the intended way by doing the following:
```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # import flow
```
But I feel like this would be pointless, because we currently don't do type-checking. Should I include it?

lnst/Tests/RDMABandwidth.py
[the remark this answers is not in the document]
Kuba314: Changed it to be `logging.debug` instead.

lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
```diff
@@ -0,0 +1,31 @@
...
+    @property
+    def start_timestamp(self):
+        return self._bandwidth.start_timestamp
+
+    @property
+    def end_timestamp(self):
+        return self._bandwidth.end_timestamp
+
+    def time_slice(self, start, end):
+        return self
```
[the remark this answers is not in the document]
Kuba314: Ah, right. Of course. Fixed.

lnst/Recipes/ENRT/MeasurementGenerators/RDMABandwidthMeasurementGenerator.py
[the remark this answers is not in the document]
Kuba314: Done.

lnst/Controller/Recipe.py
olichtne: i think this is fine:

```python
class MyRecipe():
    p = ConstParam(value=123)

MyRecipe(p=123) # this is ok
MyRecipe(p=321) # this raises an exception that the constant is being changed
```

lnst/Recipes/ENRT/SoftwareRDMARecipe.py
```diff
@@ -0,0 +1,47 @@
...
+    RDMA / InfiniBand bandwidth test recipe
+    """
+
+    # rxe driver / siw driver
+    software_rdma_type = ChoiceParam(type=StrParam, choices={"rxe", "siw"}, mandatory=True)
+
+    # override BaseFlowMeasurementGenerator's incompatible params
+    perf_tests = ConstParam(value=("rdma_stream",))
+    net_perf_tool = ConstParam(value="rdma-measurement")
+    ip_versions = ConstParam(value=("ipv4",))
```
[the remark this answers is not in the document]
olichtne: hmmm, ok probably a todo item then and let's check with an expert

lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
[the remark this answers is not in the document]
olichtne: i think i'd prefer at least to use the string version: `flow: "Flow"` just so we're somewhat consistent within the function parameters.

Kuba314:
> additionally, can you add some beaker job test ids?

J:8089873. I reduced it and also baselines are missing. I'm also not sure if I'm reporting the results correctly, because currently I'm just doing every individual result, but I should be probably doing something _aggregated_. See [RDMABandwidthMeasurement.report_results()](https://github.com/LNST-project/lnst/pull/311/files#diff-4255eedb3be82575407d19fe74c0b94ae1a8a0817a8f21f88d55196fe0388a91R139).

olichtne:
> J:8089873. I reduced it and also baselines are missing. I'm also not sure if I'm reporting the results correctly, because currently I'm just doing every individual result, but I should be probably doing something _aggregated_. See [RDMABandwidthMeasurement.report_results()](https://github.com/LNST-project/lnst/pull/311/files#diff-4255eedb3be82575407d19fe74c0b94ae1a8a0817a8f21f88d55196fe0388a91R139).

missing baselines are ok, this will also require changes on the RHExtensions side internally - since you implemented a new Result type it will need a new Translator class that imports/exports data to/from the catalog database.

Wrt. reporting, yeah imo you should just report the aggregated **average +-3*std_deviation  unit** type of string

olichtne:
> I've also experienced with `ib_send_bw`'s `--bidirectional --report_both` flags to get rx and tx bandwidth, but they seem almost identical and I'm not sure what we'd exactly test with that. Also the format of the output of that is not really nice to work with... so I'm sticking to a unidirectional measurement for now.

well they seem identical NOW, but the point of the test is to look for issues, so of course NOW when everything is working it seems like, why do i need to look... but i imagine there can be a bug that only affects one of the directions.

Probably ok to skip for now and leave as a next step, and should be discussed with the subject area expert, but just providing some context.

Kuba314:
I improved the reporting, see J:8103516.
> ib_send_bw measured bandwidth: 582.56 +-15.28(2.62%) MiB/s.

olichtne:
> ib_send_bw measured bandwidth: 582.56 +-15.28(2.62%) MiB/s.

that throughput seems a bit low, the normal simple network recipe is reaching line rate which is... 9.3gbps and that should be around 1125 MiB/s can we check with the area expert if this is expected?

Kuba314:
Area expert replied that this is expected. `rxe` (and I think `siw` too) is usually used as an "enabler for users that don't have the RDMA hardware" and they're not sure it was ever tested for performance.

lnst/Controller/Recipe.py
Kuba314: Fixed, please resolve.

lnst/RecipeCommon/Perf/Measurements/Results/RDMABandwidthMeasurementResults.py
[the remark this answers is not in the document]
Kuba314: Fixed, please resolve.

lnst/Recipes/ENRT/SoftwareRDMARecipe.py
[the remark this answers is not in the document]
Kuba314: We couldn't really figure this out. I'm getting
```
Address family for hostname not supported for <ipv6_address>:18515
Problem in resolving basic address and port
Unable to open file descriptor for socket connection Unable to init the socket connection
```
Let's skip IPv6 testing.

lnst/Tests/RDMABandwidth.py
```diff
@@ -0,0 +1,85 @@
...
+    device_name = StrParam(mandatory=True)
+    duration = IntParam()
+    port = IntParam()
+    cpu_bind = ListParam(type=IntParam())
+    size = IntParam(default=65536)
+
+    def run(self) -> bool:
+        self._res_data = {}
+        if not is_installed("ib_send_bw"):
+            self._res_data["msg"] = "perftest is not installed on this machine!"
```
olichtne, line 23: this mentions `perftest` not installed?

olichtne:
these are my final minor questions, i think we can merge after this.

lnst/Tests/RDMABandwidth.py
Kuba314: `perftest` is the package that provides `ib_send_bw` program and some others. See [linux-rdma/perftest](https://github.com/linux-rdma/perftest). Should I change to message to something like `"perftest package (provides ib_send_bw) ..."`?

lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
```diff
@@ -0,0 +1,32 @@
...
+        return SequentialPerfResult([result.bandwidth for result in self.individual_results])
+
+    @property
+    def individual_results(self) -> list[RDMABandwidthMeasurementResults]:
+        return self._individual_results
+
+    def add_results(self, result: RDMABandwidthMeasurementResults) -> None:
+        self._individual_results.append(result)
+
+    def describe(self) -> str:
```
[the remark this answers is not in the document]
Kuba314: Yee there's an issue with typing (and logic), because the classes are not the same. When we force inheritance like this, we force the non aggregated result to be possibly not initialized from the constructor. I think it would clean up the code if we used two separate classes AND a protocol connecting them.

```py
from dataclasses import dataclass, field
from typing import Protocol


class Result(Protocol):
    @property
    def bandwidth(self) -> float:
        ...

@dataclass
class SingleResult(Result):  # inheritance optional
    bandwidth: float
    

class AggregatedResult(Result):  # inheritance optional
    results: list[Result] = field(default_factory=list)
    
    @property
    def bandwidth(self) -> float:
        return sum(result.bandwidth for result in self.results) / len(self.results)
```
This should for sure be its own MR though.

For now, should I make `AggregatedRDMABandwidthMeasurementResults` inehrit from `RDMABandwidthMeasurementResults` and make `RDMABandwidthMeasurementResults` possible to be instantiated without any actual bandwidth, possibly making it raise an exception if any other methods were called on it, making it the same as all other results/aggregated results in LNST?

lnst/Tests/RDMABandwidth.py
olichtne: ah, i see, makes sense, sure, please change the message, that should be a bit more helpful.

lnst/RecipeCommon/Perf/Measurements/Results/AggregatedRDMABandwidthMeasurementResults.py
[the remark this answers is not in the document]
olichtne: > For now, should I make AggregatedRDMABandwidthMeasurementResults inehrit from RDMABandwidthMeasurementResults

yeah, let's keep it consistent for now.

i agree with a separate MR for a more complex refactor of the `MeasurementResults` classes, i also mentioned some "proposal" for this on our internal slack, but let's have that as a separate discussion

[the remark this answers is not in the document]
Kuba314: fixed.

Kuba314:
Note that some commits should be fixupped before merging this.

olichtne (approved):
looks ok to me now.

jtluka:
Looks ok.

jtluka (approved):

PR #312, opened 2023-06-28
RecipeCommon.Perf.MeasurementResults: refactor result description generation

olichtne:
### Description
Previously the descriptions were generated directly or almost directly as part of the Measurement classmethod "report_results" which also creates a RecipeResult object that is added to the current RecipeRun instance.

This meant that it wasn't possible to do custom descriptions or custom result objects without calling this "report_results" method which would also create the recipe result object.

This refactors the code a bit so that the MeasurementResult description is generated directly by the MeasurementResult object and the "report_results" method simply calls this when needed.

Unfortunatelly this means that the format for the CPUMeasurementResults has to change a bit as these are not grouped by "host" anymore and each individual result will contain the host id.

### Tests
CI tests should be ok for the normal functionality/syntax checks.

I did test this locally with container deployment and ran this many times with a local database as well to get a bunch of data. I'll proceed with running some internal tests on real hardware next after I reorganize my internal code updates.

### Reviews

Closes: #

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
@@ -103,3 +103,36 @@ class FlowMeasurementResults(BaseMeasurementResults):
         result_copy.receiver_results = self.receiver_results.time_slice(start, end)
 
         return result_copy
+
+    def describe(self):
+        generator = self.generator_results
+        generator_cpu = self.generator_cpu_stats
+        receiver = self.receiver_results
+        receiver_cpu = self.receiver_cpu_stats
+        desc = []
+        desc.append("Generator measured throughput: {tput:.2f} +-{deviation:.2f}({percentage:.2f}%) {unit} per second."
+                .format(tput=generator.average,
+                        deviation=generator.std_deviation,
+                        percentage=self._deviation_percentage(generator),
+                        unit=generator.unit))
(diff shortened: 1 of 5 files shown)


lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
```diff
@@ -103,3 +103,36 @@ def time_slice(self, start, end):
...
+                        deviation=receiver.std_deviation,
+                        percentage=self._deviation_percentage(receiver),
+                        unit=receiver.unit))
+        desc.append("Receiver process CPU data: {cpu:.2f} +-{cpu_deviation:.2f} {cpu_unit} per second."
+                .format(cpu=receiver_cpu.average,
+                        cpu_deviation=receiver_cpu.std_deviation,
+                        cpu_unit=receiver_cpu.unit))
+        return "\n".join(desc)
+
+    @classmethod
```
jtluka, line 133: Does this need to be a class method? Not a big deal, just trying to understand why.

jtluka (approved):
Looks ok.

lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
olichtne: ah! good catch, that's an artifact from copy pasting from the previous location, this can probably be removed now... since the `cls` argument isn't used anywhere anyway.

enhaut (approved):
lgtm

PR #314, opened 2023-07-07
ENRT/PerfReversibleFlowMixin: fix _create_perf_flow()

jtluka:
### Description
The method in BaseFlowMeasurementGenerator has been extended with client_port parameter. The reversible mixin has not been updated to match this change and so any recipe using this mixin would crash because of different number of parameters used.

Fixes commit b0dc4521b0844d6a6999fc3bd3312e1dfe8996fe

### Tests

Internal job id 8042835

### Reviews

Hotfix, will merge without review once tested.


jtluka:
Internal test passed.

Issue #210, opened 2021-07-22
Ignore IPv6 addresses obtained by autoconf


naggety:
Bond and Team recipes had ended with "unexpected error", printing this backtrace:
```
Traceback (most recent call last):
  File "/home/ihuguet/Dev/lnst/./thunderx-nicvf-tests.py", line 49, in <module>
    ctl.run(team)
  File "/home/ihuguet/Dev/lnst/lnst/Controller/Controller.py", line 161, in run
    recipe.test()
  File "/home/ihuguet/Dev/lnst/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 164, in test
    self.do_tests(recipe_config)
  File "/home/ihuguet/Dev/lnst/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 285, in do_tests
    self.do_ping_tests(recipe_config)
  File "/home/ihuguet/Dev/lnst/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 295, in do_ping_tests
    for ping_configs in self.generate_ping_configurations(recipe_config):
  File "/home/ihuguet/Dev/lnst/lnst/Recipes/ENRT/BaseEnrtRecipe.py", line 342, in generate_ping_configurations
    raise LnstError("Source/destination ip lists are of different size.")
lnst.Common.LnstError.LnstError: Source/destination ip lists are of different size.
```
Adding some log messages to BaseEnrtRecipe.py, generate_ping_configurations, I see this IPv6 lists:
```
TeamDevice(machine=host1, id=team0, name=t_team0, ifindex=22)
[Ip6Address(fc00:0:0:1::1/64), Ip6Address(fd00::1e1b:dff:fe9d:b048/64)]
Device(machine=host2, id=eth0, name=enp4s0f4d1, ifindex=7)
[Ip6Address(fc00:0:0:1::2/64)]
```

In the slave, I confirm that info:
```
# ip addr show t_team0
22: t_team0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 1c:1b:0d:9d:b0:48 brd ff:ff:ff:ff:ff:ff
    inet 192.168.10.1/24 scope global t_team0
       valid_lft forever preferred_lft forever
    inet6 fd00::1e1b:dff:fe9d:b048/64 scope global dynamic mngtmpaddr
       valid_lft 86394sec preferred_lft 14394sec
[...]

PR #318, opened 2023-08-03
Add better endpoints definitions

Kuba314:
### Description
Make `generate_perf_endpoints` return resolved IP addresses instead of just devices.

### Tests
`kernel-full` test group in J:8289303

### Reviews
@LNST-project/lnst-developers 

Closes: #210

### DIFF
--- a/lnst/RecipeCommon/endpoints.py
+++ b/lnst/RecipeCommon/endpoints.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+from dataclasses import dataclass
+from collections.abc import Iterator
+from typing import Generic, TypeVar
+
+from lnst.Common.IpAddress import BaseIpAddress
+from lnst.Controller.Namespace import Namespace
+from lnst.Devices import RemoteDevice
+
+
+# marked as covariant to allow saving a value of type `IPEndpoint[Ip4Address]`
+# to a variable of type `IPEndpoint[BaseIpAddress]`
+TIpAddress_co = TypeVar("TIpAddress_co", bound=BaseIpAddress, covariant=True)
+
+
+@dataclass
+class Endpoint:
+    """Basic endpoint object.
+
+    The host and device name are derived from a provided device.
+    """
+
+    device: RemoteDevice
+
+    @property
+    def host(self) -> Namespace:
+        return self.device.netns
+
+    @property
+    def device_name(self) -> str:
+        return self.device.name
+
+
+@dataclass
+class IPEndpoint(Endpoint, Generic[TIpAddress_co]):
+    """IP endpoint object"""
+
+    address: TIpAddress_co
+
+
+@dataclass
+class IPPortEndpoint(Generic[TIpAddress_co], IPEndpoint[TIpAddress_co]):
+    port: int
+
+
+# marked as covariant to allow saving a value of type `EndpointPair[IPEndpoint]`
+# to a variable of type `EndpointPair[Endpoint]`
+TEndpoint_co = TypeVar("TEndpoint_co", bound=Endpoint, covariant=True)
+
+
+@dataclass
+class EndpointPair(Generic[TEndpoint_co]):
+    """Represents a pair of endpoints."""
+
+    first: TEndpoint_co
+    second: TEndpoint_co
+
+    def __iter__(self) -> Iterator[TEndpoint_co]:
+        yield self.first
+        yield self.second
--- a/lnst/Recipes/ENRT/BaseEnrtRecipe.py
+++ b/lnst/Recipes/ENRT/BaseEnrtRecipe.py
@@ -1,6 +1,7 @@
 import pprint
 import copy
 from contextlib import contextmanager
+from typing import Literal, Optional
 
 from lnst.Common.LnstError import LnstError
 from lnst.Common.Parameters import (
@@ -10,8 +11,9 @@ from lnst.Common.Parameters import (
     BoolParam,
     FloatParam,
 )
-from lnst.Common.IpAddress import AF_INET, AF_INET6
+from lnst.Common.IpAddress import AF_INET, AF_INET6, BaseIpAddress
 from lnst.Controller.RecipeResults import ResultType
+from lnst.Devices import RemoteDevice
 from lnst.Recipes.ENRT.ConfigMixins.BaseSubConfigMixin import BaseSubConfigMixin
 from lnst.Recipes.ENRT.MeasurementGenerators.BaseMeasurementGenerator import (
     BaseMeasurementGenerator,
@@ -25,14 +27,43 @@ from lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement import BaseFlowMeas
 from lnst.RecipeCommon.Perf.Evaluators import NonzeroFlowEvaluator
 from lnst.RecipeCommon.Ping.Evaluators import RatePingEvaluator
 
-class EnrtConfiguration(object):
+
+class EnrtConfiguration:
     """Container object for configuration
 
-    Intentionally left empty as it is intended to be used as a container to
-    store any values relevant to configuration being applied during the lifetime
-    of the Recipe.
+    Stores configured devices and IPs configured on them. Can also be
+    used to store any values relevant to configuration being applied
+    during the lifetime of the Recipe.
     """
-    pass
+    def __init__(self):
+        self._device_ips = {}
+
+    @property
+    def configured_devices(self) -> list[RemoteDevice]:
+        return list(self._device_ips.keys())
+
+    def ips_for_device(
+        self, device: RemoteDevice, family: Optional[Literal[AF_INET, AF_INET6]] = None
+    ) -> list[BaseIpAddress]:
+        if family is None:
+            return self._device_ips[device]
+        return [ip for ip in self._device_ips[device] if ip.family == family]
+
+    def track_device(self, device: RemoteDevice) -> None:
+        self._device_ips.setdefault(device, [])
+
+    def untrack_device(self, device: RemoteDevice) -> None:
+        del self._device_ips[device]
+
+    def configure_and_track_ip(
+        self,
+        device: RemoteDevice,
+        ip_address: BaseIpAddress,
+        peer: Optional[BaseIpAddress] = None,
+    ) -> None:
+        """Configure IP for device and"""
+        device.ip_add(ip_address, peer=peer)
+        self._device_ips.setdefault(device, []).append(ip_address)
 
 
 class BaseEnrtRecipe(
--- a/lnst/Recipes/ENRT/BaseEnrtRecipe.py
+++ b/lnst/Recipes/ENRT/BaseEnrtRecipe.py
@@ -281,7 +281,7 @@ class BaseEnrtRecipe(
                         pprint.pformat(self.params._to_dict())))
         self.add_result(ResultType.PASS, "\n".join(description))
 
-    def generate_test_wide_description(self, config):
+    def generate_test_wide_description(self, config: EnrtConfiguration):
         """Generates the test wide configuration description
 
         Another class inteded to be used with the collaborative version of the
--- a/lnst/Recipes/ENRT/BaseTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/BaseTunnelRecipe.py
@@ -1,3 +1,4 @@
+from lnst.Devices import RemoteDevice
 from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
 from lnst.RecipeCommon.PacketAssert import PacketAssertTestAndEvaluate
 from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
@@ -25,7 +26,7 @@ class BaseTunnelRecipe(
     * :meth:`get_packet_assert_config`
     """
 
-    def test_wide_configuration(self):
+    def test_wide_configuration(self) -> EnrtConfiguration:
         """
         The base class defines common steps to create the test wide
         configuration of a tunnel recipe.
@@ -35,17 +36,13 @@ class BaseTunnelRecipe(
         then the tunnel is created between the specified endpoints in
         :meth:`create_tunnel`.
         """
+        config = super().test_wide_configuration()
+        config.tunnel_endpoints = self.configure_underlying_network(config)
+        config.tunnel_devices = self.create_tunnel(config, config.tunnel_endpoints)
+        self.wait_tentative_ips(config.configured_devices)
+        return config
 
-        configuration = super().test_wide_configuration()
-        configuration.tunnel_endpoints = None
-        configuration.tunnel_devices = []
-
-        self.configure_underlying_network(configuration)
-        self.create_tunnel(configuration)
-
-        return configuration
-
-    def configure_underlying_network(self, configuration):
+    def configure_underlying_network(self, config: EnrtConfiguration) -> tuple[RemoteDevice, RemoteDevice]:
         """
         This method must be implemented by the child class.
 
@@ -57,29 +54,41 @@ class BaseTunnelRecipe(
         etc.
 
         The method must also update the ``tunnel_endpoints`` attribute of the
-        ``configuration`` object with a tuple of two Device objects that will
+        ``config`` object with a tuple of two Device objects that will
         be used as tunnel endpoints in :meth:`create_tunnel`
 
-        :param configuration:
-            An EnrtConfiguration instance that will be populated with the
-            configured network devices and tunnel endpoints.
-        :type configuration: :any:`EnrtConfiguration`
+        :param config:
+            Used to configure IP addresses and store them inside the config
+        :type config: `EnrtConfiguration`
+
+        :return: returns a tuple of two Device objects that will
+            be used as tunnel endpoints in :meth:`create_tunnel`
+        :rtype: `tuple[RemoteDevice, RemoteDevice]`
         """
         raise NotImplementedError
 
-    def create_tunnel(self, configuration):
+    def create_tunnel(
+        self,
+        config: EnrtConfiguration,
+        tunnel_endpoints: tuple[RemoteDevice, RemoteDevice],
+    ) -> tuple[RemoteDevice, RemoteDevice]:
         """
         This method must be implemented by the child class.
 
-        The child class should create a network tunnel between endpoints
-        provided in the ``configuration``. That usually includes creating
-        tunnel devices such as :any:`GreDevice`, :any:`VxlanDevice`, etc.
-        and configuring the IP addresses.
+        The child class should create a network tunnel between the provided
+        endpoints. That usually includes creating tunnel devices such as
+        :any:`GreDevice`, :any:`VxlanDevice`, etc. and configuring the IP addresses.
 
-        :param configuration:
-            An EnrtConfiguration instance that should have ``tunnel_endpoints``
-            attribute populated with the tunnel endpoints.
-        :type configuration: :any:`EnrtConfiguration`
+        :param config:
+            Configuration object useful for assigning IP addresses to devices
+        :type config: `EnrtConfiguration`
+
+        :param tunnel_endpoints:
+            A tuple of the tunnel endpoints.
+        :type tunnel_endpoints: `tuple[RemoteDevice, RemoteDevice]`
+
+        :return: A tuple of the configured tunnel devices.
+        :rtype: `tuple[RemoteDevice, RemoteDevice]
         """
         raise NotImplementedError
 
@@ -103,8 +112,6 @@ class BaseTunnelRecipe(
 
     def test_wide_deconfiguration(self, config):
         ""  # overriding the parent docstring
-        del config.tunnel_devices
-
         super().test_wide_deconfiguration(config)
 
     def ping_test(self, ping_configs):
@@ -137,7 +144,7 @@ class BaseTunnelRecipe(
         """
         raise NotImplementedError
 
-    def generate_perf_endpoints(self, config):
+    def generate_perf_endpoints(self, config: EnrtConfiguration):
         """
         The perf endpoints for recipes derived from this class are usually
         the tunnel endpoints. The derived class can override the endpoints
--- a/lnst/Recipes/ENRT/BaseTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/BaseTunnelRecipe.py
@@ -1,5 +1,6 @@
 from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
 from lnst.RecipeCommon.PacketAssert import PacketAssertTestAndEvaluate
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
 
 
 class BaseTunnelRecipe(
@@ -36,7 +37,6 @@ class BaseTunnelRecipe(
         """
 
         configuration = super().test_wide_configuration()
-        configuration.test_wide_devices = []
         configuration.tunnel_endpoints = None
         configuration.tunnel_devices = []
 
@@ -83,7 +83,7 @@ class BaseTunnelRecipe(
         """
         raise NotImplementedError
 
-    def generate_test_wide_description(self, config):
+    def generate_test_wide_description(self, config: EnrtConfiguration):
         """
         Test wide description is extended with the configured addresses
         of the underlying network devices and the configured tunnel devices
@@ -91,7 +91,7 @@ class BaseTunnelRecipe(
         desc = super().generate_test_wide_description(config)
         desc += [
             "Configured {}.{}.ips = {}".format(dev.host.hostid, dev.name, dev.ips)
-            for dev in config.test_wide_devices
+            for dev in config.configured_devices
         ]
         desc += [
             "Configured tunnel endpoint {}.{}.ips = {}".format(
@@ -104,7 +104,6 @@ class BaseTunnelRecipe(
     def test_wide_deconfiguration(self, config):
         ""  # overriding the parent docstring
         del config.tunnel_devices
-        del config.test_wide_devices
 
         super().test_wide_deconfiguration(config)
 
--- a/lnst/Recipes/ENRT/BaseTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/BaseTunnelRecipe.py
@@ -1,4 +1,8 @@
+from collections.abc import Collection
+
 from lnst.Devices import RemoteDevice
+from lnst.RecipeCommon.endpoints import EndpointPair, IPEndpoint
+from lnst.Recipes.ENRT.helpers import ip_endpoint_pairs
 from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
 from lnst.RecipeCommon.PacketAssert import PacketAssertTestAndEvaluate
 from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
@@ -140,10 +144,11 @@ class BaseTunnelRecipe(
         """
         raise NotImplementedError
 
-    def generate_perf_endpoints(self, config: EnrtConfiguration):
+    def generate_perf_endpoints(self, config: EnrtConfiguration) -> list[Collection[EndpointPair[IPEndpoint]]]:
         """
         The perf endpoints for recipes derived from this class are usually
         the tunnel endpoints. The derived class can override the endpoints
         if needed.
         """
-        return [tuple(config.tunnel_devices)]
+        dev1, dev2 = config.tunnel_devices
+        return [ip_endpoint_pairs(config, (dev1, dev2))]
--- a/lnst/Recipes/ENRT/BaseTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/BaseTunnelRecipe.py
@@ -110,10 +110,6 @@ class BaseTunnelRecipe(
         ]
         return desc
 
-    def test_wide_deconfiguration(self, config):
-        ""  # overriding the parent docstring
-        super().test_wide_deconfiguration(config)
-
     def ping_test(self, ping_configs):
         pa_config = self.get_packet_assert_config(ping_configs[0])
         self.packet_assert_test_start(pa_config)
--- a/lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
@@ -104,11 +104,11 @@ class GeneveOvsTunnelRecipe(
         m1 = endpoint1.netns
         m2 = endpoint2.netns
 
-        for i, (host, endpoint) in enumerate([(m1, endpoint2), (m2, endpoint1)]):
+        for i, (host, endpoint) in enumerate([(m1, endpoint2), (m2, endpoint1)], 1):
             host.br0 = OvsBridgeDevice()
             host.int0 = host.br0.port_add(interface_options={"type": "internal"})
-            config.configure_and_track_ip(host.int0, ipaddress("192.168.200." + str(i + 1) + "/24"))
-            config.configure_and_track_ip(host.int0, ipaddress("fc00::" + str(i + 1) + "/64"))
+            config.configure_and_track_ip(host.int0, ipaddress(f"192.168.200.{i}/24"))
+            config.configure_and_track_ip(host.int0, ipaddress(f"fc00::{i}/64"))
 
             remote_ip = config.ips_for_device(endpoint)[0]
             host.br0.tunnel_add(
--- a/lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
@@ -12,8 +12,9 @@ from lnst.Common.Parameters import (
 )
 from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
 from lnst.RecipeCommon.PacketAssert import PacketAssertConf
-from lnst.Devices import OvsBridgeDevice
+from lnst.Devices import OvsBridgeDevice, RemoteDevice
 from lnst.Recipes.ENRT.BaseTunnelRecipe import BaseTunnelRecipe
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
 from lnst.Recipes.ENRT.ConfigMixins.OffloadSubConfigMixin import (
     OffloadSubConfigMixin,
 )
@@ -73,7 +74,7 @@ class GeneveOvsTunnelRecipe(
 
     net_ipv4 = IPv4NetworkParam(default="192.168.101.0/24")
 
-    def configure_underlying_network(self, configuration):
+    def configure_underlying_network(self, config: EnrtConfiguration) -> tuple[RemoteDevice, RemoteDevice]:
         """
         The underlying network for the tunnel consists of the Ethernet
         devices on the matched hosts.
@@ -81,14 +82,16 @@ class GeneveOvsTunnelRecipe(
         host1, host2 = self.matched.host1, self.matched.host2
         ipv4_addr = interface_addresses(self.params.net_ipv4)
         for device in [host1.eth0, host2.eth0]:
-            device.ip_add(next(ipv4_addr))
+            config.configure_and_track_ip(device, next(ipv4_addr))
             device.up()
-            configuration.test_wide_devices.append(device)
 
-        self.wait_tentative_ips(configuration.test_wide_devices)
-        configuration.tunnel_endpoints = (host1.eth0, host2.eth0)
+        return (host1.eth0, host2.eth0)
 
-    def create_tunnel(self, configuration):
+    def create_tunnel(
+        self,
+        config: EnrtConfiguration,
+        tunnel_endpoints: tuple[RemoteDevice, RemoteDevice],
+    ) -> tuple[RemoteDevice, RemoteDevice]:
         """
         OvS bridges are created on each of the matched hosts with two ports.
         One port as an integration port and another port of type Geneve acting
@@ -97,19 +100,17 @@ class GeneveOvsTunnelRecipe(
         Integration ports are configured with IPv4 and IPv6 addresses
         of the tunneled networks.
         """
-        endpoint1, endpoint2 = configuration.tunnel_endpoints
+        endpoint1, endpoint2 = tunnel_endpoints
         m1 = endpoint1.netns
         m2 = endpoint2.netns
-        ip_filter = {"family": AF_INET}
 
         for i, (host, endpoint) in enumerate([(m1, endpoint2), (m2, endpoint1)]):
-            remote_ip = endpoint.ips_filter(**ip_filter)[0]
             host.br0 = OvsBridgeDevice()
             host.int0 = host.br0.port_add(interface_options={"type": "internal"})
-            configuration.tunnel_devices.append(host.int0)
-            host.int0.ip_add(ipaddress("192.168.200." + str(i + 1) + "/24"))
-            host.int0.ip_add(ipaddress("fc00::" + str(i + 1) + "/64"))
+            config.configure_and_track_ip(host.int0, ipaddress("192.168.200." + str(i + 1) + "/24"))
+            config.configure_and_track_ip(host.int0, ipaddress("fc00::" + str(i + 1) + "/64"))
 
+            remote_ip = config.ips_for_device(endpoint)[0]
             host.br0.tunnel_add(
                 "geneve", {"options:remote_ip": remote_ip, "options:key": 1234}
             )
@@ -117,7 +118,7 @@ class GeneveOvsTunnelRecipe(
             host.br0.up()
             host.int0.up()
 
-        self.wait_tentative_ips(configuration.tunnel_devices)
+        return (m1.int0, m2.int0)
 
     def generate_ping_endpoints(self, config):
         """
--- a/lnst/Recipes/ENRT/GeneveTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/GeneveTunnelRecipe.py
@@ -6,7 +6,7 @@ from lnst.Common.IpAddress import (
     Ip6Address,
     interface_addresses,
 )
-from lnst.Devices import GeneveDevice
+from lnst.Devices import GeneveDevice, RemoteDevice
 from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
 from lnst.RecipeCommon.PacketAssert import PacketAssertConf
 from lnst.Common.Parameters import (
@@ -17,6 +17,7 @@ from lnst.Common.Parameters import (
     IPv6NetworkParam,
 )
 from lnst.Recipes.ENRT.BaseTunnelRecipe import BaseTunnelRecipe
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
 from lnst.Recipes.ENRT.ConfigMixins.OffloadSubConfigMixin import (
     OffloadSubConfigMixin,
 )
@@ -82,7 +83,7 @@ class GeneveTunnelRecipe(
     net_ipv4 = IPv4NetworkParam(default="192.168.101.0/24")
     net_ipv6 = IPv6NetworkParam(default="fc00::/64")
 
-    def configure_underlying_network(self, configuration):
+    def configure_underlying_network(self, config: EnrtConfiguration) -> tuple[RemoteDevice, RemoteDevice]:
         """
         The underlying network for the tunnel consists of the Ethernet
         devices on the matched hosts.
@@ -92,29 +93,27 @@ class GeneveTunnelRecipe(
         ipv6_addr = interface_addresses(self.params.net_ipv6)
         for device in [host1.eth0, host2.eth0]:
             if self.params.carrier_ipversion == "ipv4":
-                device.ip_add(next(ipv4_addr))
+                config.configure_and_track_ip(device, next(ipv4_addr))
             else:
-                device.ip_add(next(ipv6_addr))
+                config.configure_and_track_ip(device, next(ipv6_addr))
             device.up_and_wait()
-            configuration.test_wide_devices.append(device)
 
-        self.wait_tentative_ips(configuration.test_wide_devices)
-        configuration.tunnel_endpoints = (host1.eth0, host2.eth0)
+        return (host1.eth0, host2.eth0)
 
-    def create_tunnel(self, configuration):
+    def create_tunnel(
+        self,
+        config: EnrtConfiguration,
+        tunnel_endpoints: tuple[RemoteDevice, RemoteDevice],
+    ) -> tuple[RemoteDevice, RemoteDevice]:
         """
         The Geneve tunnel devices are configured with IPv4 and IPv6 addresses.
         """
-        endpoint1, endpoint2 = configuration.tunnel_endpoints
+        endpoint1, endpoint2 = tunnel_endpoints
         m1 = endpoint1.netns
         m2 = endpoint2.netns
-        if self.params.carrier_ipversion == "ipv4":
-            ip_filter = {"family": AF_INET}
-        else:
-            ip_filter = {"family": AF_INET6, "is_link_local": False}
 
-        endpoint1_ip = endpoint1.ips_filter(**ip_filter)[0]
-        endpoint2_ip = endpoint2.ips_filter(**ip_filter)[0]
+        endpoint1_ip = config.ips_for_device(endpoint1)[0]
+        endpoint2_ip = config.ips_for_device(endpoint2)[0]
 
         a_ip4 = Ip4Address("20.0.0.10/8")
         b_ip4 = Ip4Address("20.0.0.20/8")
@@ -128,17 +127,16 @@ class GeneveTunnelRecipe(
         # A
         m1.gnv_tunnel.mtu = 1400
         m1.gnv_tunnel.up_and_wait()
-        m1.gnv_tunnel.ip_add(a_ip4)
-        m1.gnv_tunnel.ip_add(a_ip6)
+        config.configure_and_track_ip(m1.gnv_tunnel, a_ip4)
+        config.configure_and_track_ip(m1.gnv_tunnel, a_ip6)
 
         # B
         m2.gnv_tunnel.mtu = 1400
         m2.gnv_tunnel.up_and_wait()
-        m2.gnv_tunnel.ip_add(b_ip4)
-        m2.gnv_tunnel.ip_add(b_ip6)
+        config.configure_and_track_ip(m2.gnv_tunnel, b_ip4)
+        config.configure_and_track_ip(m2.gnv_tunnel, b_ip6)
 
-        configuration.tunnel_devices.extend([m1.gnv_tunnel, m2.gnv_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
+        return (m1.gnv_tunnel, m2.gnv_tunnel)
 
     def generate_ping_endpoints(self, config):
         """
--- a/lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
@@ -104,11 +104,11 @@ class GreOvsTunnelRecipe(
         m1 = endpoint1.netns
         m2 = endpoint2.netns
 
-        for i, (host, endpoint) in enumerate([(m1, endpoint2), (m2, endpoint1)]):
+        for i, (host, endpoint) in enumerate([(m1, endpoint2), (m2, endpoint1)], 1):
             host.br0 = OvsBridgeDevice()
             host.int0 = host.br0.port_add(interface_options={"type": "internal"})
-            config.configure_and_track_ip(host.int0, ipaddress("192.168.200." + str(i + 1) + "/24"))
-            config.configure_and_track_ip(host.int0, ipaddress("fc00::" + str(i + 1) + "/64"))
+            config.configure_and_track_ip(host.int0, ipaddress(f"192.168.200.{i}/24"))
+            config.configure_and_track_ip(host.int0, ipaddress(f"fc00::{i}/64"))
 
             remote_ip = config.ips_for_device(endpoint)[0]
             host.br0.tunnel_add("gre", {"options:remote_ip": remote_ip})
--- a/lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
+++ b/lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
@@ -12,8 +12,9 @@ from lnst.Common.Parameters import (
 )
 from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
 from lnst.RecipeCommon.PacketAssert import PacketAssertConf
-from lnst.Devices import OvsBridgeDevice
+from lnst.Devices import OvsBridgeDevice, RemoteDevice
 from lnst.Recipes.ENRT.BaseTunnelRecipe import BaseTunnelRecipe
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
 from lnst.Recipes.ENRT.ConfigMixins.OffloadSubConfigMixin import (
     OffloadSubConfigMixin,
 )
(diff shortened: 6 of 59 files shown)


lnst/RecipeCommon/endpoints.py
```diff
@@ -0,0 +1,87 @@
...
+    @property
+    def host(self) -> Namespace:
+        ...
+
+    @property
+    def device_name(self) -> str:
+        ...
+
+
+TIpAddress_co = TypeVar("TIpAddress_co", bound=BaseIpAddress, covariant=True)
```
olichtne, line 20: the typing in this module could probably use some docstring explanations or examples so that it's easier to understand what usecases are expected and attempted to be represented with the types.

lnst/RecipeCommon/endpoints.py
```diff
@@ -85,3 +85,17 @@ def from_devices_with_ips(
...
+
+def sequential_ip_endpoints(
+    *device_pairs: tuple[Device, Device]
+) -> list[list[EndpointPair[IPEndpointLike]]]:
+    """Helper function for use in generate_perf_endpoints method.
+
+    Generates a sequential endpoint list for each input device pair.
+    """
+    return [
+        [EndpointPair.from_devices_with_ips(dev1, dev2, ip_type=ip_type)]
```
olichtne, line 98: this inner list is here due to the internal semantics of the `generate_perf_endpoints` right? if so then i think it should be removed from here as this is a more "common/generic" module and instead the `generate_perf_endpoints` method should then do an additional transformation which adds the use case specific semantics.

lnst/Devices/Device.py
```diff
@@ -532,6 +533,13 @@ def ips(self):
         """
         return self._ip_addrs
 
+    @property
+    def lnst_ips(self) -> list[BaseIpAddress]:
```
olichtne, line 537: i get what the goal is here, but i'm not 100% sure about placing this property into the `Device` class.

My main reason is that the "abstract idea" behind the Device class is to provide a "get and push" interface for the device configuration on the system. These two operations being separated but always representing the "current/immediate" state - a push operation immediately sends the configuration request, and a get operation retrieves the current most up to date information.

This `lnst_ips` property however is... static and represents an "expected" state during some time of the test recipe. There can be situations where this is not the TRUE state of the system. An example of such a situation being if there's some configuration action that happens outside of the scope of the `Device` class. For example if a recipe calls some kind of a `/sys/` path configuration that for some reason initiates a reset of the device configuration. In this case this property stays static, and that's GOOD - when we generate endpoints I want that to be static based on what we THINK we configured. However, if this property is placed in the `Device` class I think it may be confusing as everything else in the class represents the immediate state.

And at the same time I think that this implementation of the `lnst_ips` may not be fully sufficient for all of our test recipes. For example, some recipes step outside of the `Device` classes for their configuration, such as the DPDK recipes, because in these cases there aren't really any "configured" ip addresses since the device driver module is unloaded, however there are "imagined" configured ip addresses that should be used for the Endpoint generation.

lnst/RecipeCommon/endpoints.py
```diff
@@ -0,0 +1,87 @@
...
+            Endpoint(dev2),
+        )
+
+    @classmethod
+    def from_devices_with_ips(
+        cls, dev1: Device, dev2: Device, ip_type: type[TIpAddress]
+    ) -> EndpointPair[IPEndpoint[TIpAddress]]:
+        selected_ips = []
+        for dev in [dev1, dev2]:
+            dev_ips = [ip for ip in dev.lnst_ips if isinstance(ip, ip_type)]
```
olichtne, line 78: not sure if this method should have filtering capability. we're already going to be applying the filtering where we use the results of this method.

IMO it's ok to have this method just always generate all of the device-ip endpoint configurations, and let the filtering be done by the user of the result afterwards.

olichtne (changes requested):

olichtne:
overall i like this first version of the implementation, i think there are some conceptual discussions that i tried to point to with my comments that may change the design a bit but i think this is very much a step in the correct direction and a very good start.

lnst/RecipeCommon/endpoints.py
Kuba314: So, you're proposing changing each recipe's `generate_perf_endpoints` like this:
```py
# before
return sequential_ip_endpoints(
    (self.matched.host1.eth0, self.matched.host2.eth0),
)

# after
dev1, dev2 = self.matched.host1.eth0, self.matched.host2.eth0
return [
    [EndpointPair.from_devices_with_ips(dev1, dev2, ip_type=ip_type)]
    for ip_type in [Ip4Address, Ip6Address]
]
```
Can't we just move the function to a different module instead?

lnst/RecipeCommon/endpoints.py
Kuba314: Changed the name to `all_with_ips` and made it yield all ip endpoint pairs. Let me know if you're happy with how I implemented it.

lnst/Devices/Device.py
Kuba314: Right, could we maybe change all recipes to manually store each configured IP in `recipe_conf`, or make `Device` wrapped in something, BUT AFAICS no DPDK recipe inherits from `BaseFlowMeasurementGenerator` and thus doesn't define `generate_perf_endpoints`. The `OvSDPDKBondRecipe` *does* define this method, but I think that's a mistake, because it's probably not called anywhere - the return value also doesn't make any sense, because it returns `DummyRecipeConfig`s, which don't define `ips_filter`, which is used to filter the IPs from endpoints currently.

As for `Device` being an immediate state... I'm also against mixing it, but at the same time I would expect each recipe to have access to its configured NICs and each NIC have access to its configured IPs. Maybe we should rename the property to something like `configured_lnst_ips` so its clear we expect them to be *configured* as opposed to actually being there I guess. I'm not sure how else to tackle this. I wouldn't really want to create a wrapper or a different structure for just IPs.

lnst/RecipeCommon/endpoints.py
Kuba314: Added docstrings to the classes. Let me know if they make more sense now.

lnst/RecipeCommon/endpoints.py
olichtne: well my "internal semantics of `generate_perf_endpoints`" i meant the method or how it is used - we can do your proposal or move the filter into the function that USES the result of the `generate_perf_endpoints` so imo this would be fine as well:

```python
return sequential_ip_endpoints((self.matched.host1.eth0, self.matched.host2.eth0))
```

and then the user of the method:
```python
useful_endpoints = apply_some_filters(self.generate_perf_endpoints)
```

Kuba314: That's kind of how this is done now. The filtering is done in `BaseFlowMeasurementGenerator`'s `generate_flow_combinations()`

lnst/Devices/Device.py
Kuba314: Not relevant anymore.

lnst/RecipeCommon/endpoints.py
Kuba314: Not relevant anymore.

lnst/RecipeCommon/endpoints.py
Kuba314: Changed it so this method returns a simple list of generated endpoints. Since it's now strongly tied to `BaseEnrtRecipe`, we should maybe move it there, maybe even make it a method of `EnrtConfiguration`?

lnst/RecipeCommon/endpoints.py
```diff
@@ -81,3 +83,27 @@ def from_devices(cls, dev1: Device, dev2: Device) -> EndpointPair[Endpoint]:
             Endpoint(dev1),
             Endpoint(dev2),
         )
+
+
+def ip_endpoints(
+    config: EnrtConfiguration, *device_pairs: tuple[Device, Device]
```
olichtne, line 89: i don't really like that this `lnst.RecipeCommon.endpoints` function is importing an `lnst.Recipes.ENRT.BaseEnrtRecipe` class... the recipe common package should be a more "common" supporting library that is used in the Recipes package. this is in reverse, creating a sort of cycle... could actually easily lead to cyclical dependencies imo.

i think it would be safer:
1. move this method into the enrt package - if it's a helper method that we strictly use in the enrt package, and it depends on using classes/types from the enrt package... it should be in the enrt package
2. if it  is a more common use case method, it should use more common use case arguments/types. and so the method signature should be more like `ip_endpoints(*device_pairs: tuple[tuple[Device, list[IpAddress]], tuple[Device, list[IpAddress]]])`

lnst/Recipes/ENRT/BaseTunnelRecipe.py
```diff
@@ -141,10 +143,11 @@ def get_packet_assert_config(self, ping_config):
...
-    def generate_perf_endpoints(self, config: TunnelConfiguration):
+    def generate_perf_endpoints(self, config: TunnelConfiguration) -> list[Collection[EndpointPair[IPEndpointLike]]]:
         """
         The perf endpoints for recipes derived from this class are usually
         the tunnel endpoints. The derived class can override the endpoints
         if needed.
         """
-        return [tuple(config.tunnel_devices)]
+        dev1, dev2 = config.tunnel_devices
+        return ip_endpoints(config, (dev1, dev2))
```
olichtne, line 149: this one is missing the top level list wrapper initializer - all of the other classes in this commit are changed to return:
```
[ip_endpoints(...)]
```
is this intentional here? The return type of the method is still the same.

olichtne (changes requested):
i like the current state, just some minor comments to discuss/resolve.

lnst/RecipeCommon/endpoints.py
olichtne: ah i see you've made the same comment i have independently while reviewing this. i'll copy that comment here and close the other thread so that we can keep the context in just a single thread:



i don't really like that this lnst.RecipeCommon.endpoints function is importing an lnst.Recipes.ENRT.BaseEnrtRecipe class... the recipe common package should be a more "common" supporting library that is used in the Recipes package. this is in reverse, creating a sort of cycle... could actually easily lead to cyclical dependencies imo.

i think it would be safer:

    move this method into the enrt package - if it's a helper method that we strictly use in the enrt package, and it depends on using classes/types from the enrt package... it should be in the enrt package
    if it is a more common use case method, it should use more common use case arguments/types. and so the method signature should be more like ip_endpoints(*device_pairs: tuple[tuple[Device, list[IpAddress]], tuple[Device, list[IpAddress]]])

lnst/RecipeCommon/endpoints.py
olichtne: closing in favour of the older thread.

lnst/Recipes/ENRT/SimpleMacsecRecipe.py
```diff
@@ -42,7 +42,7 @@ def test_wide_configuration(self):
 
         return configuration
 
-    def generate_test_wide_description(self, config):
+    def generate_sub_configuration_description(self, config):
         host1, host2 = self.matched.host1, self.matched.host2
         desc = super().generate_test_wide_description(config)
```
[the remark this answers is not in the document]
Kuba314: Do you mean to move it after the `generate_ping_configurations` method? I don't see any other testwide things than `test_wide_configuration`, which is already before that.

lnst/RecipeCommon/endpoints.py
Kuba314: Do you want me to explain what covariant means? Or why did I use it maybe? I used it because I want to be able to assign a value of type `EndpointPair[IPEndpoint]` to a variable of type `EndpointPair[Endpoint]`. If the typevar wasn't covariant, this wouldn't work.

Should I put this in a comment there?

lnst/RecipeCommon/endpoints.py
Kuba314: I'll move it elsewhere. I don't really want to make it more universal, because each recipe would have to do this:
```py
# before
ip_endpoints(config, (dev1, dev2))

# after
ip_endpoints(((dev1, config.ips_for_device(dev1)), (dev2, config.ips_for_device(dev2))))
```

lnst/Recipes/ENRT/BaseTunnelRecipe.py
```diff
@@ -1,8 +1,20 @@
+from typing import Optional, Union
+
+from lnst.Devices import Device, RemoteDevice
 from lnst.Recipes.ENRT.BaremetalEnrtRecipe import BaremetalEnrtRecipe
 from lnst.RecipeCommon.PacketAssert import PacketAssertTestAndEvaluate
 from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
 
 
+AnyDevice = Union[Device, RemoteDevice]
```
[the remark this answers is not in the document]
Kuba314: When creating software devices (such as `GreTunnelRecipe`'s `GreDevice`), they inherit from `Device`, not from `RemoteDevice`. These devices are then put into `tunnel_devices`. What should I do in this case?

lnst/Recipes/ENRT/BaseEnrtRecipe.py
```diff
@@ -25,14 +27,34 @@
 from lnst.RecipeCommon.Perf.Evaluators import NonzeroFlowEvaluator
 from lnst.RecipeCommon.Ping.Evaluators import RatePingEvaluator
 
-class EnrtConfiguration(object):
+
+AnyDevice = Union[Device, RemoteDevice]
```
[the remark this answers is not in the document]
Kuba314: Ahh, right. This Union is not needed here, I think it's only needed in `BaseTunnelRecipe`. See the other discussion. I'll remove it from BaseEnrtRecipe.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
```diff
@@ -25,14 +27,34 @@
...
+    def configured_devices(self) -> list[AnyDevice]:
+        return list(self._device_ips.keys())
+
+    def track_device(self, device: AnyDevice) -> None:
+        self._device_ips.setdefault(device, [])
+
+    def untrack_device(self, device: AnyDevice) -> None:
+        del self._device_ips[device]
+
+    def register_device_ip(self, device: AnyDevice, ip_address: BaseIpAddress) -> None:
```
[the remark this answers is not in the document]
Kuba314: I think it's a better name than `add_device_ip` or something like that. I'm not sure if there's a better name. Do you have some ideas? Maybe something like `configure_and_register_ip()` or maybe `configure_ip()` is enough?

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
Kuba314: Wait, should all endpoints classes use `RemoteDevice` instead of `Device` then?

lnst/Recipes/ENRT/BaseTunnelRecipe.py
```diff
@@ -35,17 +47,12 @@ def test_wide_configuration(self):
         then the tunnel is created between the specified endpoints in
         :meth:`create_tunnel`.
         """
+        config = self.configure_underlying_network()
+        self.create_tunnel(config)
 
-        configuration = super().test_wide_configuration()
```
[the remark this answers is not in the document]
Kuba314: Right, I get what you mean, but... tunnel recipe config is split into two methods, setup of network devices in `configure_underlying_network()` and the tunnel creation on top of these devices in `create_tunnel()`. If you look at this from a typing perspective, how do we know that the `tunnel_endpoints` property/attribute of the configuration is set when we call `create_tunnel`?

Previously this was not checked. I now changed it so that `configure_underlying_network()` actually creates the configuration object by supplying it with the configured endpoints so that we're sure we have them. If the configuration was created beforehand we wouldn't be able to ensure that we actually properly set these (and we'd be getting a ton of typing errors. I know there's a ton already, but I'm trying to think futureproof).

I guess what we could do is remove `TunnelConfiguration` and make `configure_underlying_network()` return a tuple of configured endpoints make `create_tunnel()` return a tuple of configured tunnel devices. Both of these tuples would then be set in `BaseTunnelRecipe.test_wide_configuration` to the `EnrtConfiguration` so that we don't break the inheritance. Sounds good?

lnst/Recipes/ENRT/BaseTunnelRecipe.py
Kuba314: Definitely not, nice catch!

lnst/Recipes/ENRT/SimpleMacsecRecipe.py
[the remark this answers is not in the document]
Kuba314: Ah, I get what you mean - move it where is should be (after config, but before deconfig). Will fix this.

lnst/RecipeCommon/endpoints.py
Kuba314: Moved to `lnst.Recipes.ENRT.helpers`.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
olichtne: `configure_and_register_ip()` is probably explicit enough.

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
olichtne: you mean this? https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/GeneveLwtTunnelRecipe.py#L134

right well... there's a bit of python forbidden magic at play here. The way that device classes work in lnst is that the constructs:

```python
Host.new_dev_var = SomeDevice()
```

actually creates a `RemoteDevice(SomeDevice)` instance that is assigned to the `Host` object, and on the assignment call (so `setattr`) we detect a RemoteDevice assignment and translate it into a configuration command sent to the agent where the actual `SomeDevice` instance is created.

so the order of operations is:
1. import SomeDevice from `lnst.Devices` instead of `lnst.Devices.SomeDevice` e.g. - https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/GeneveLwtTunnelRecipe.py#L10 imports GeneveDevice from `lnst.Device`
2. https://github.com/LNST-project/lnst/blob/master/lnst/Devices/__init__.py#L48 `lnst.Device.GeneveDevice` is actually wrapped as a `RemoteDevice(Geneve) here
3. https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/GeneveLwtTunnelRecipe.py#L134 creates a `RemoteDevice(GeneveDevice)` instance
4. the assignment goes to https://github.com/LNST-project/lnst/blob/master/lnst/Controller/Namespace.py#L157 where we detect a RemoteDevice and send a create device command to the agent.

so on the controller, you're never actually using the Device class instances.... YOU MAY however use them for instance checks:
```python
from lnst.Device import RemoteGeneveDevice
from lnst.Device.GeneveDevice import GeneveDevice

m1.geneve_device = RemoteGeneveDevice(...)

isinstance(m1.geneve_device, GeneveDevice) == True
isinstance(m1.geneve_device, RemoteGeneveDevice) == True
```

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
olichtne: > Wait, should all endpoints classes use RemoteDevice instead of Device then?

if they're only used on the Controller then i think that's correct.

lnst/RecipeCommon/endpoints.py
olichtne: >  I used it because I want to be able to assign a value of type EndpointPair[IPEndpoint] to a variable of type EndpointPair[Endpoint]. If the typevar wasn't covariant, this wouldn't work.

i think this should be the comment in the code then.

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
Kuba314: Right... this is confusing and very type unsafe resulting in a lot of errors like
```
Missing positional argument "ifmanager" in call to "GeneveDevice"
```
I wish this was handled better but it seems we're stuck like this, at least for some time.

I'll use only `RemoteDevice` typehints then.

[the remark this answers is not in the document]
olichtne: > Missing positional argument "ifmanager" in call to "GeneveDevice"

this can only happen if the type checker is using the incorrect thing... we're not actually importing `GeneveDevice` in the recipe... we're importing the decorated `RemoteDevice` and that doesn't take the ifmanager parameter. So if the type checker correctly used the right objects this error shouldn't appear.

But that's a separate discussion, let's have that at some coffee time or something instead of here...

lnst/Recipes/ENRT/SimpleMacsecRecipe.py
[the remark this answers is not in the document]
Kuba314: fixed.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
Kuba314: fixed.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
Kuba314: renamed. resolving

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
Kuba314: true, but I don't think type-checkers can enforce this when we're editing `globals()` using strings. Anyway, I removed the Union and used `RemoteDevice` instead.

lnst/RecipeCommon/endpoints.py
Kuba314: Added the comment there.

lnst/RecipeCommon/endpoints.py
olichtne: looks ok

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
Kuba314: Sadly this does not work. The line
```py
config: TunnelConfiguration = super().test_wide_configuration()
```
will display an error because the `EnrtConfiguration` object returned from `BaseEnrtRecipe.test_wide_configuration()` will not conform to the `TunnelConfiguration` protocol. Also I tested this locally and even after setting the tunnel properties to the config object, the object still doesn't conform to the `TunnelConfiguration` protocol. You can try this yourself:
```py
from typing import Protocol


class TunnelConfig(Protocol):
    @property
    def tunnel_endpoints(self) -> tuple:
        ...


class EnrtConfig:
    pass


tunnel_config: TunnelConfig

enrt_config = EnrtConfig()
tunnel_config = enrt_config  # incompatible types error

enrt_config.tunnel_endpoints = (1, 2)  # no attribute error
tunnel_config = enrt_config  # incompatible types error
```
On top of that, this doesn't solve the issue of `create_tunnel` not knowing whether `configure_underlying_network()` actually defined the endpoints in the config.

I'm fairly strongly in favor of the following solution:
```py
class BaseTunnelRecipe(...):
    def test_wide_configuration(self) -> EnrtConfiguration:
        config.tunnel_endpoints = self.configure_underlying_network(config)
        config.tunnel_devices = self.create_tunnel(config.tunnel_endpoints)
        self.wait_tentative_ips(config.tunnel_devices)
        return config

    def configure_underlying_network(self, config: EnrtConfiguration) -> tuple[RemoteDevice, RemoteDevice]:
        ...
        
    def create_tunnel(self, tunnel_endpoints: tuple[RemoteDevice, RemoteDevice]) -> tuple[RemoteDevice, RemoteDevice]:
        ...
```
Note that for some reason `create_tunnel()` adds IPs to tunnel devices, but we never tracked those AFAICS.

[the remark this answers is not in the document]
olichtne: > Sadly this does not work.

ok, was just a guess... good to know. I guess it makes sense, typing (including this Protocol typing intended for ducktyping) is intended for STATIC analysis... and what my "proposal" would require would be a DYNAMIC type check analysis as the duck typing simply changes over the course of execution.

---

i agree with your final proposal, i think that looks good.

---

if we ever get back to this again i think the proper solution to introducing "recipe specific config class" would have to look like this:

```python
class EnrtConfig(): ...

class BaseEnrtRecipe():
    config_class = EnrtConfig

    @contextmanager
      def _test_wide_context(self):
          config = self.config_class
          self.test_wide_configuration(config)
          self.describe_test_wide_configuration(config)
          try:
              yield config
          finally:
              self.test_wide_deconfiguration(config)


---

class TunnerConfig(EnrtConfig): ...

class TunnelRecipe(...):
    config_class = TunnelConfig
```

this should work and at this point it should be possible to do a correct typing static analysis as well.

it will however require us to update all of our `test_wide_configuration` methods to include `config` as a parameter... but i think that may be a good thing if we want to introduce stricter typing into this.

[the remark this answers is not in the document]
Kuba314: All tunnel recipes have been edited to follow the structure I proposed.

Kuba314:
The test job in the description finished with a few failed installs thanks to poetry issues it seems. I downloaded the test logs of tests that actually ran and I see no `Exception` in them, but that doesn't mean that this PR is safe. What should I do next? (also marking as ready)

olichtne:
> The test job in the description finished with a few failed installs thanks to poetry issues it seems. I downloaded the test logs of tests that actually ran and I see no `Exception` in them, but that doesn't mean that this PR is safe. What should I do next? (also marking as ready)

can you provide the jobid?

what i would like to be checked is:
1. that there are no exceptions/tracebacks
2. there are no weird ERRORs... that's a bit undefined for us as we do have "normal errors" wrt. some config options but probably some sort of a `grep ERROR | sort | uniq` to check if there's anything weird is enough
3. that all of the tests that ran correctly found baselines and evaluated... this patch shouldn't impact performance so the actual pass/fail result of this is unlikely to be problematic... but a check that everything works is recommended.

Kuba314:
> can you provide the jobid?

@olichtne `J:8255628`, it's in this PR's description.

> that there are no exceptions/tracebacks

grep 'Exception' returns nothing.

> there are no weird ERRORs...

There appear to be some errors, I'll investigate.
```
$ grep -r ERROR | cut -d' ' -f 3- | sort -u | grep -v ethtool
         (guest1)        -   ERROR: 
         (guest1)        -   ERROR: errors reported by iperf
          (host1)        -   ERROR: 
          (host1)        -   ERROR: Command "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
          (host1)        -   ERROR: Command "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
          (host1)        -   ERROR: [Errno 3] No such process
          (host1)        -   ERROR: errors reported by iperf
          (host1)        -   ERROR: Iperf provided incomplete json data
          (host2)        -   ERROR: Command "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
          (host2)        -   ERROR: Command "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
          (host2)        -   ERROR: Iperf provided incomplete json data
      (localhost)        -   ERROR: Recipe execution terminated by unexpected exception
      (localhost)        -   ERROR: Result time alignment impossible, falling back to unaligned results
```

> that all of the tests that ran correctly found baselines and evaluated

Any idea how to check for this?

olichtne:
>          (guest1)        -   ERROR: errors reported by iperf
>           (host1)        -   ERROR: [Errno 3] No such process
>           (host1)        -   ERROR: errors reported by iperf
>           (host1)        -   ERROR: Iperf provided incomplete json data
>           (host2)        -   ERROR: Iperf provided incomplete json data

these iperf ones are imo unexpected and may be realted to the changes here...
this can happen when iperf fails to connect or bind to a nic, which may be
related to the endpoints.


>       (localhost)        -   ERROR: Result time alignment impossible, falling back to unaligned results

this one i'm not sure about, probably double check...

>       (localhost)        -   ERROR: Recipe execution terminated by unexpected exception

This suggests that some exceptions have appeared. Maybe also try to grep with
`-i` to ignore case... and just double check this specific test-runner task to
see if anything more can be figured out.

>           (host1)        -   ERROR: Command "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
>           (host1)        -   ERROR: Command "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)
The no_turbo error lines are ok, those appear regularly.

> > that all of the tests that ran correctly found baselines and evaluated
> 
> Any idea how to check for this?

I'd grep for "no baseline found" and for "no threshold found"... not sure about
the exact wording but that's where i'd start.

If either of those is found, i'd get a list of uuids - so do a `grep -l` to get
the file names and then grep UUID from just those files.

And then I'd check probably the dashboard to see if those tests normally work
or if they also don't have baselines even when running in a production
pipeline.

Kuba314:
> `errors reported by iperf`

(probably expected?) `warning: UDP block size 1400 exceeds TCP MSS 1378, may result in fragmentation / drops`

> `Command "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" execution failed (exited with 1)`

This appears in a ton of recipe runs, probably expected.

> `Iperf provided incomplete json data`

only in `MPTCPRecipe`s, will have to investigate.

> `Recipe execution terminated by unexpected exception`

Ah, apparently there are exceptions, should've searched with `grep -i`... I will look into this...

> `Result time alignment impossible, falling back to unaligned results`

Is this related to neper and expected?

Kuba314:
Reran full test-set again with the errors fixed at J:8264579.

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
jtluka: One question - why you had to change and needed to explicitly assign the result of these two methods:
```
        config.tunnel_endpoints = self.configure_underlying_network(config)
        config.tunnel_devices = self.create_tunnel(config, config.tunnel_endpoints)
```

Why this assignment can't be done directly in the methods?

lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
```diff
@@ -117,7 +120,7 @@ def create_tunnel(self, configuration):
             host.br0.up()
             host.int0.up()
 
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 120: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GeneveTunnelRecipe.py
```diff
@@ -128,17 +132,16 @@ def create_tunnel(self, configuration):
...
         # B
         m2.gnv_tunnel.mtu = 1400
         m2.gnv_tunnel.up_and_wait()
-        m2.gnv_tunnel.ip_add(b_ip4)
-        m2.gnv_tunnel.ip_add(b_ip6)
+        config.configure_and_register_ip(m2.gnv_tunnel, b_ip4)
+        config.configure_and_register_ip(m2.gnv_tunnel, b_ip6)
 
-        configuration.tunnel_devices.extend([m1.gnv_tunnel, m2.gnv_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 141: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
```diff
@@ -97,25 +101,24 @@ def create_tunnel(self, configuration):
...
-            host.int0.ip_add(ipaddress("fc00::" + str(i + 1) + "/64"))
+            config.configure_and_register_ip(host.int0, ipaddress(f"192.168.200.{i}/24"))
+            config.configure_and_register_ip(host.int0, ipaddress(f"fc00::{i}/64"))
 
             host.br0.tunnel_add("gre", {"options:remote_ip": remote_ip})
 
             host.br0.up()
             host.int0.up()
 
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 118: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GreTunnelOverBondRecipe.py
```diff
@@ -161,20 +165,19 @@ def create_tunnel(self, configuration):
...
         m2.gre_tunnel.up()
-        m2.gre_tunnel.ip_add(b_ip4)
-        m2.gre_tunnel.ip_add(b_ip6)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip4)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip6)
         m2.run("ip -4 route add {} dev {}".format(a_net4, m2.gre_tunnel.name))
         m2.run("ip -6 route add {} dev {}".format(a_net6, m2.gre_tunnel.name))
 
-        configuration.tunnel_devices.extend([m1.gre_tunnel, m2.gre_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 177: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GreTunnelOverMacvlanRecipe.py
```diff
@@ -130,20 +134,19 @@ def create_tunnel(self, configuration):
...
         m2.gre_tunnel.up()
-        m2.gre_tunnel.ip_add(b_ip4)
-        m2.gre_tunnel.ip_add(b_ip6)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip4)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip6)
         m2.run("ip -4 route add {} dev {}".format(a_net4, m2.gre_tunnel.name))
         m2.run("ip -6 route add {} dev {}".format(a_net6, m2.gre_tunnel.name))
 
-        configuration.tunnel_devices.extend([m1.gre_tunnel, m2.gre_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 146: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
```diff
@@ -134,20 +138,19 @@ def create_tunnel(self, configuration):
...
         m2.gre_tunnel.up()
-        m2.gre_tunnel.ip_add(b_ip4)
-        m2.gre_tunnel.ip_add(b_ip6)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip4)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip6)
         m2.run("ip -4 route add {} dev {}".format(a_net4, m2.gre_tunnel.name))
         m2.run("ip -6 route add {} dev {}".format(a_net6, m2.gre_tunnel.name))
 
-        configuration.tunnel_devices.extend([m1.gre_tunnel, m2.gre_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 150: You removed this from the code, so we will now won't wait for tentative flag.

lnst/Recipes/ENRT/GreTunnelRecipe.py
```diff
@@ -114,20 +118,19 @@ def create_tunnel(self, configuration):
...
         m2.gre_tunnel.up_and_wait()
-        m2.gre_tunnel.ip_add(b_ip4)
-        m2.gre_tunnel.ip_add(b_ip6)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip4)
+        config.configure_and_register_ip(m2.gre_tunnel, b_ip6)
         m2.run("ip -4 route add {} dev {}".format(a_net4, m2.gre_tunnel.name))
         m2.run("ip -6 route add {} dev {}".format(a_net6, m2.gre_tunnel.name))
 
-        configuration.tunnel_devices.extend([m1.gre_tunnel, m2.gre_tunnel])
-        self.wait_tentative_ips(configuration.tunnel_devices)
```
jtluka, line 130: You removed this from the code, so we will now won't wait for tentative flag.

Kuba314:
There were some other failures, I'll mark this as draft again until these are investigated.

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
```diff
@@ -101,66 +104,64 @@ def generate_flow_combinations(self, config):
...
-        if ipv == "ipv4":
-            ip_filter.update(family=AF_INET)
-        elif ipv == "ipv6":
-            ip_filter.update(family=AF_INET6)
-            ip_filter.update(is_link_local=False)
-
-        client_ips = client_nic.ips_filter(**ip_filter)
-        server_ips = server_nic.ips_filter(**ip_filter)
+    ) -> list[PerfFlow]:
+        port_iter = itertools.count(12000)
```
jtluka, line 144: I think we should initialize the iterator for each endpoint pair. So, move it before:
```
for i in range(self.params.perf_parallel_processes):
```

lnst/Recipes/ENRT/SimpleMacsecRecipe.py
```diff
@@ -98,6 +93,7 @@ def generate_sub_configuration_description(self, config):
     def remove_sub_configuration(self, config):
         host1, host2 = config.host1, config.host2
         for host in (host1, host2):
+            config.untrack_device(host.msec0)
```
jtluka, line 96: Why is this required?

lnst/Recipes/ENRT/VirtualBridgeVlanInGuestMirroredRecipe.py
```diff
@@ -60,18 +64,16 @@ def test_wide_configuration(self):
...
-        configuration.test_wide_devices = [guest1.vlan0, guest2.vlan0,
-            host1.br0, host2.br0]
+        config = super().test_wide_configuration()
 
         ipv4_addr = interface_addresses(self.params.net_ipv4)
         ipv6_addr = interface_addresses(self.params.net_ipv6, default_start="fc00:0:0:1::3/64")
-        host1.br0.ip_add(next(ipv4_addr))
-        host2.br0.ip_add(next(ipv4_addr))
+        config.configure_and_register_ip(host1.br0, next(ipv4_addr))
+        config.configure_and_register_ip(host2.br0, next(ipv6_addr))
```
jtluka, line 72: I think this is a typo. Originally, the test configures ipv4 address on host1.br0 and host2.br0 (the other host). Now you configure ipv4 address on host1 and ipv6 address on other host.

lnst/Recipes/ENRT/VirtualOvsBridgeVlanInGuestRecipe.py
```diff
@@ -49,62 +53,58 @@ def test_wide_configuration(self):
...
-        configuration.test_wide_devices = [host2.vlan0, guest1.vlan0]
+        config = super().test_wide_configuration()
 
         ipv4_addr = interface_addresses(self.params.net_ipv4, default_start="192.168.10.2/24")
         ipv6_addr = interface_addresses(self.params.net_ipv6, default_start="fc00:0:0:1::2/64")
         for machine in [host2, guest1]:
+            config.configure_and_register_ip(host2.vlan0, next(ipv4_addr))
+            config.configure_and_register_ip(guest1.vlan0, next(ipv6_addr))
             machine.vlan0.ip_add(next(ipv4_addr))
             machine.vlan0.ip_add(next(ipv6_addr))
```
jtluka, line 64: This code block does not look right. If this is to replace the `ip_add()` methods, they're still kept here. Also the new `configure_and_register_ip()` is run twice on the same parameters (`host2.vlan0` and `guest1,vlan0`), but should be on `machine.vlan0`.

lnst/Recipes/ENRT/VlansOverBondRecipe.py
```diff
@@ -125,12 +129,8 @@ def test_wide_configuration(self):
...
         host2.vlan2 = VlanDevice(realdev=host2.eth0, vlan_id=self.params.vlan2_id)
 
-        configuration = super().test_wide_configuration()
-        configuration.test_wide_devices = []
-        for host in [host1, host2]:
-            configuration.test_wide_devices.extend([host.vlan0,
-                host.vlan1, host.vlan2])
-        configuration.test_wide_devices.append(host1.bond0)
+        config = super().test_wide_configuration()
+        config.track_device(host1.bond0)
```
jtluka, line 133: So, my understanding of track/untrack_device methods is to simply get a device into `configured_devices` so that it can be processed in `generate_test_wide_description()` if needed. However, in particular in this recipe, the description method directly references the device where required. So, it's not needed to track it here.

Similarly in other recipes you can simply do self.configured_devices + [self.matched.host1.some_device].

I'm not quite sure if this method is useful at all. Some explanation would be good here.

lnst/Recipes/ENRT/VxlanNetnsTunnelRecipe.py
```diff
@@ -117,6 +118,7 @@ def configure_underlying_network(self, configuration):
         host2.bridge.slave_add(host2.veth0)
         host2.bridge.slave_add(host2.eth0)
 
+
```
jtluka, line 121: Extra line?

lnst/Recipes/ENRT/VlansOverBondRecipe.py
jtluka: Ondrej explained the concept to me offline. However, I have one rename proposal. Can't we unify the `track/untrack_device()` and `configure_and_register()` methods to either:
* `register/unregister_device()` and `configure_and_register()` or
* `track/untrack_device()` and `configure_and_track()`

Simply to know that these methods are related.

@olichtne wdyt?

lnst/Recipes/ENRT/SimpleMacsecRecipe.py
jtluka: After offline discussion, I understand why this is required. Simply to avoid corruption in multiple sub configuration runs.

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
Kuba314: @jtluka I could, but then we wouldn't be able to enforce that something was actually assigned. Also, it's best to keep functions pure. This function's purpose is to accept a pair of endpoints and create a tunnel on top of them. For me it's a *creation* function and thus should return the created result. Consider the following the functions:
```py
def add(a, b):
    return a + b

def add(obj, a, b):  # replace obj with self, same thing
    obj.attr = a + b
```
I'm for option 1, every time.

lnst/Recipes/ENRT/VlansOverBondRecipe.py
Kuba314: > my understanding of track/untrack_device methods is to simply get a device into `configured_devices` so that it can be processed in generate_test_wide_description() if needed.

Your understanding is correct. Some recipes added some devices to `self.test_wide_devices` even though the devices in question didn't get any IP assigned. I mirrored this behaviour to the `track_device()` method. However, I think this could be eventually unified. Ondrej shared his thoughts that he'd like to keep a track of every configured device in there, but I'm not sure how we would do that yet.

I'd definitely leave this out of this MR though, so I'll just rename the ip method to `configure_and_track_ip()` for now. Later I think, most recipes should be cleaned up. There's a lot of unnecessary duplication.

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
Kuba314: The generated `PerfFlow` instances are ran in parallel (see description of return value of `generate_flow_combinations()`. Therefore I think these can't have the same ports.

lnst/Recipes/ENRT/GeneveOvsTunnelRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GeneveTunnelRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GreOvsTunnelRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GreTunnelOverBondRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GreTunnelOverMacvlanRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GreTunnelOverVlanRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/GreTunnelRecipe.py
Kuba314: Accidentally removed, I returned it back there.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
```diff
@@ -25,14 +26,32 @@
...
     """
-    pass
+    def __init__(self):
+        self._device_ips = {}
+
+    @property
+    def configured_devices(self) -> list[RemoteDevice]:
+        return list(self._device_ips.keys())
+
+    def track_device(self, device: RemoteDevice) -> None:
```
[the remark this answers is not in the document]
Kuba314: I wouldn't really like to add a doc-string to all the methods so instead I edited the class doc-string and mentioned the purpose of storing configured devices and IPs. Is this ok with you?

lnst/Recipes/ENRT/VirtualBridgeVlanInGuestMirroredRecipe.py
Kuba314: A very good catch. Thanks, fixed.

lnst/Recipes/ENRT/VirtualOvsBridgeVlanInGuestRecipe.py
Kuba314: A very good catch. Thanks, fixed.

lnst/Recipes/ENRT/VxlanNetnsTunnelRecipe.py
Kuba314: Fixed.

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
jtluka: Typically the enpoints have different IP addresses, so using the same port should be ok. Btw, are there endpoints that can define the ports themselves. If so, the generator should use them.

lnst/Recipes/ENRT/BaseTunnelRecipe.py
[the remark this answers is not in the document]
jtluka: Ok, let's keep it as is.

lnst/Recipes/ENRT/BaseEnrtRecipe.py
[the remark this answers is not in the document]
jtluka: Sure, that works for me.

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
olichtne: having a different port number should be fine regardless of if the ip addresses of the endpoints are the same or different, the space for port numbers is big enough that we likely can't exhaust it with a test like this....

on the other hand if the endpoint ips are the same then having the same port number would be an issue...

i think the decision for this is what keeps the closes backwards compatibility - if we apply this version, will the tests that utilize multiple parallel flows be all different or will they stay the same? Does this change anything wrt. looking for baselines in catalog?

---

i aggree with the "if an endpoint defines a port it should be used" instead of overwriting it with a new value though...

lnst/Recipes/ENRT/VlansOverBondRecipe.py
olichtne: ack to `configure_and_track_ip()`

lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py
Kuba314: > if we apply this version, will the tests that utilize multiple parallel flows be all different or will they stay the same?

I think everything stays the same until some recipe defines multiple IPs of the same type (let's say IPv4) on both endpoints defined in `generate_perf_endpoints()`. Currently, we do a product on all defined IPs of the same type and run these in *parallel*. See the definition of `ip_endpoints`. Maybe we should change that to be sequential?

Previously if A and B each defined multiple IPs, only the first pair would be picked I think. Currently, we pick all of them *at once*.
```
previously:
A.a <-> B.a

now:
A.a <-> B.a
A.a <-> B.b
A.b <-> B.a
A.b <-> B.b

File: lnst/Recipes/ENRT/VlansOverBondRecipe.py (hunk 23 above)
Comment (Kuba314) [reply lnst/Recipes/ENRT/VlansOverBondRecipe.py:133]: They have been renamed and this can be resolved.

File: lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py (hunk 19 above)
Comment (jtluka) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: > I think everything stays the same until some recipe defines multiple IPs of the same type

Yes, I think this should not be an issue. I'm fine with different ports based on your arguments.

Comment (jtluka) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: > i aggree with the "if an endpoint defines a port it should be used" instead of overwriting it with a new value though...

the question is how this should work if there are multiple processes. because there would be a conflict if the same port is reused for all measurements.

Comment (olichtne) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: > > i aggree with the "if an endpoint defines a port it should be used" instead of overwriting it with a new value though...
> 
> the question is how this should work if there are multiple processes. because there would be a conflict if the same port is reused for all measurements.

IMO.... if the tester requests specific ports with specific endpoints... then these should be used exactly as requested... if this doesn't work because there is a port conflict... that's either intended and expected by the tester... or it's a bug caused by the tester and should be fixed... we shouldn't attempt to fix his or disallow these scenarios from the tester.

----

> Previously if A and B each defined multiple IPs, only the first pair would be picked I think. Currently, we pick all of them at once.

i'm ok with this, especially since we don't have a test configuration that would be affected by this at this moment. The behaviour imo makes sense, I just want to make sure we're not unknowingly changing something that previously worked one way to work another way without proper checks...

**Review by olichtne — CHANGES_REQUESTED** [review]
i'm ok with merging this after the test failures are resolved and the threads from @jtluka are resolved as well.

File: lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py (hunk 19 above)
Comment (Kuba314) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: So... the `generate_perf_endpoints()` returns a list of lists of `EndpointPair[IPEndpointLike]`. You basically want to make sure that if someone _does_ return `IPPortEndpoint` instead of `IPEndpoint`, you want to respect the port. I don't think adding a `isinstance` check to `generate_flow_combinations()` where the ports are used is a good idea.

Maybe we should change `generate_perf_endpoints()` return value to be `IPPortEndpoint` and let the recipe author pick the ports themselves. I believe this is the cleanest solution. Do you agree? @olichtne @jtluka

Comment (olichtne) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: that would mean that the logic of "generate 100 parallel flows" would move into the specific recipes where the `generate_perf_endpoints` would have to be the generator that integrates the port number generation correct?

it sounds... correct, but i think that means that it also will generate a bunch of duplication that we will want somehow move into a shared/common function or at least a helper....

soooo... does that just end up being moved to the same level where the current `generate_flow_combinations` method is? because if so... we're back to the current generator solution?

Comment (Kuba314) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: Hmm... right, I forgot about the `perf_parallel_processes` param... What should we do in a scenario where a tester requests `perf_parallel_processes=1000`, but also returns endpoints with specific ports? We can't follow the ports then. Do we just error out?

Comment (Kuba314) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: This would mean we *will* need to `isinstance` check the returned endpoint pairs... Or we make `IPPortEndpoint` class have `port: Optional[int]`?

Personally, I would _not_ support custom port specification _for now_. Currently, we do not need it. I would like to finally merge this. We can leave this for a different PR.

Comment (olichtne) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: yeah i think we can postpone this. open an issue so we have a reminder about it. @jtluka  do  you agree?

Comment (jtluka) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: Sure, let's followup on this in separate issue.

However, I want to make sure that the parallel iperf test scenarios work fine after this change. Looking at the results of the linked job it seems it's not.

Original:
```
2023-08-31 10:36:18       (localhost)        -    INFO: 
    CPU Baseline average evaluation for Host host1:
    PASS: CPU cpu: 1514_utilization 6.87% lower than baseline. Allowed difference: 7.94%
2023-08-31 10:36:18       (localhost)        -    INFO: Result: PASS, What:
2023-08-31 10:36:18       (localhost)        -    INFO: 
    PASS: CPU cpu: 1515_utilization 1.75% lower than baseline. Allowed difference: 4.71%
```

New:
```
2023-09-04 19:49:26       (localhost)        -    INFO: 
    CPU Baseline average evaluation for Host host1:
    FAIL: CPU cpu: 1512_utilization no threshold found
2023-09-04 19:49:26       (localhost)        -    INFO: Result: FAIL, What:
2023-09-04 19:49:26       (localhost)        -    INFO: 
    FAIL: CPU cpu: 1513_utilization no threshold found
```

Seems that indexes got broken.

Comment (jtluka) [reply lnst/Recipes/ENRT/MeasurementGenerators/BaseFlowMeasurementGenerator.py:144]: It looks like that two calls to `ips_filter()` were removed:
```
2023-08-31 10:29:57       (localhost)        -    INFO: Result: PASS, What: Calling Device method host1.eth0.ips_filter(family=<AddressFamily.AF_INET: 2>)
2023-08-31 10:29:57       (localhost)        -    INFO: Result: PASS, What: Calling Device method host2.eth0.ips_filter(family=<AddressFamily.AF_INET: 2>)
```

I assume this is expected, just mentioning here so that we can prepare for downstream changes. I will check if this impacts single stream (basically all tests).

jtluka: Single stream are also impacted. I'll close this thread as it's unrelated to the original issue discussed here.

Kuba314:
Status update: `J:8350199` should finally have fixed thresholds! Once I verify that, I'll create a downstream MR that updates them. Once that's done I'll let you know here as we have to merge both at the same time.

olichtne:
checked the tests and compared the new comparisons generated by github, all seems ok, will do one more full check of the commits just to be sure and we should be ok to merge tomorrow.

olichtne (approved):

PR #319, opened 2023-08-04
container_files/install.sh: fix install path for neper

olichtne:
### Description
Neper module by default expects that neper is installed in the path '/root/neper' which means that using the container for running tests that use neper in the ENRT package is currently impossible.

Moving the neper installation into the correct location fixes the issue.

### Tests
used the following test configuration to run in the containers:

```python
recipe = ShortLivedConnectionsRecipe(
    driver="virtio",
    perf_duration=10,
    perf_iterations=2,
    perf_tool_cpu=[0],
    dev_intr_cpu=[0],
    adaptive_tx_coalescing=True,
    adaptive_rx_coalescing=True,
    ip_versions=["ipv4"],
    perf_tests=["tcp_rr"],
    perf_msg_sizes=[100],
    rx_pause_frames=False,
    tx_pause_frames=False,
    disable_turboboost=True,
    minimal_idlestates_latency=0,
    drop_caches=True,
    perf_warmup_duration=3,
)
```

### Reviews

Closes: #


container_files/install.sh
```diff
@@ -15,6 +15,8 @@ git checkout mptcp_udp_retry_cherrypick
 echo "================== Successfully installed iperf  ======================"
 
 echo "================== Installing neper  ======================"
+mkdir -p /root/
+cd /root/
 git clone -b lnst-production https://github.com/LNST-project/neper.git
```
enhaut, line 20: `git clone`'s last argument is destination path (if provided). So it would be easier just to run:
```
git clone -b lnst-production https://github.com/LNST-project/neper.git /root/neper
```
It also creates all the missing parent directories along the path.

Anyway, not a big deal, so feel free to resolve and merge.

olichtne: > It also creates all the missing parent directories along the path.

ah, didn't try that... i'll test and fix.

olichtne: yup, that worked. thanks, resolving.

PR #321, opened 2023-08-09
Add CPU pinning to TrafficControlRecipe

Kuba314:
### Description
Previously, the `TrafficControlRecipe` didn't use `StatCPUMeasurement` and didn't allow users to pin CPUs. This now works even with a CPU pinning policy (similar to perf pinning).

### Tests
J:8172306

### Reviews
@jtluka @olichtne

### DIFF
--- a/lnst/Tests/TrafficControl.py
+++ b/lnst/Tests/TrafficControl.py
@@ -8,7 +8,7 @@ from lnst.Tests.BaseTestModule import BaseTestModule
 
 
 class TrafficControlRunner(BaseTestModule):
-    batchfiles = ListParam(type=StrParam(), required=True)
+    batchfiles = ListParam(type=StrParam(), mandatory=True)
 
     def run(self) -> bool:
         self._res_data = {}
--- a/lnst/Tests/TrafficControl.py
+++ b/lnst/Tests/TrafficControl.py
@@ -1,14 +1,18 @@
+import itertools
 import logging
 import shutil
 import asyncio
 import time
+from typing import Iterator, Optional
 
-from lnst.Common.Parameters import StrParam, ListParam
+from lnst.Common.Parameters import ChoiceParam, IntParam, StrParam, ListParam
 from lnst.Tests.BaseTestModule import BaseTestModule
 
 
 class TrafficControlRunner(BaseTestModule):
     batchfiles = ListParam(type=StrParam(), mandatory=True)
+    cpu_bind = ListParam(type=IntParam())
+    cpu_bind_policy = ChoiceParam(type=StrParam, choices={"all", "round-robin"}, default="round-robin")
 
     def run(self) -> bool:
         self._res_data = {}
@@ -31,18 +35,35 @@ class TrafficControlRunner(BaseTestModule):
 
     async def run_instances(self) -> list[dict]:
         tc_exec = shutil.which("tc")
+        cpu_bind_gen = self._get_cpu_bind_generator()
+
         instances = [
-            self.run_tc(tc_exec, bf)
+            self.run_tc(tc_exec, bf, cpu_bind=next(cpu_bind_gen))
             for bf in self.params.batchfiles
         ]
         results = await asyncio.gather(*instances)
         return results
 
-    async def run_tc(self, tc_exec: str, batchfile: str) -> dict[str]:
+    async def run_tc(
+        self,
+        tc_exec: str,
+        batchfile: str,
+        cpu_bind: Optional[list[int]] = None,
+    ) -> dict[str]:
+        args = [tc_exec, "-b", batchfile]
+        if cpu_bind is not None:
+            args = [
+                "/usr/bin/taskset",
+                "-c",
+                ",".join(map(str, cpu_bind)),
+                *args,
+            ]
+
         start_timestamp = time.time()
         start_time = time.perf_counter()
+        logging.debug(f"Running TC: {args}")
         proc = await asyncio.create_subprocess_exec(
-            tc_exec, "-b", batchfile,
+            *args,
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
         )
@@ -57,3 +78,13 @@ class TrafficControlRunner(BaseTestModule):
             stderr=stderr.decode(),
             batchfile=batchfile,
         )
+
+    def _get_cpu_bind_generator(self) -> Iterator[Optional[list[int]]]:
+        cpu_bind = self.params.get("cpu_bind")
+        if not cpu_bind:
+            return itertools.repeat(None)
+
+        policy = self.params.cpu_bind_policy
+        if policy == "round-robin":
+            return ([cpu] for cpu in itertools.cycle(cpu_bind))
+        return itertools.repeat(cpu_bind)
--- a/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
@@ -24,16 +24,7 @@ class TcRunInstance:
         self._num_rules = num_rules
         self._instance_id = instance_id
         self._batchfile_path: PathLike = None
-        self._instance_job: Job = None
         self.validate()
-
-    @property
-    def instance_job(self):
-        return self._instance_job
-
-    @instance_job.setter
-    def instance_job(self, job: Job):
-        self._instance_job = job
 
     @property
     def batchfile_path(self):
--- a/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
@@ -114,6 +114,8 @@ class TcRunMeasurement(BaseMeasurement):
             num_instances: int,
             rules_per_instance: int,
             timeout: int = 120,
+            cpu_bind: Optional[list[int]] = None,
+            cpu_bind_policy: str = "round-robin",
             parent_recipe_conf=None,
     ):
         super().__init__(recipe_conf=parent_recipe_conf)
@@ -123,6 +125,8 @@ class TcRunMeasurement(BaseMeasurement):
         self._running_jobs: list[Job] = []
         self._finished_jobs: dict[Job] = []
         self._timeout = timeout
+        self._cpu_bind = cpu_bind
+        self._cpu_bind_policy = cpu_bind_policy
         self.instance_configs = self._make_instances_cfgs()
 
     @property
@@ -160,8 +164,15 @@ class TcRunMeasurement(BaseMeasurement):
 
     def _prepare_jobs(self) -> list[Job]:
         batchfiles = [i.generate_batchfile() for i in self.instance_configs]
+        params: dict = {
+            "batchfiles": batchfiles,
+        }
+        if self._cpu_bind is not None:
+            params["cpu_bind"] = self._cpu_bind
+            params["cpu_bind_policy"] = self._cpu_bind_policy
+
         job = self.host.prepare_job(
-            TrafficControlRunner(batchfiles=batchfiles),
+            TrafficControlRunner(**params),
             job_level=ResultLevel.NORMAL,
         )
         return [job]
--- a/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/TcRunMeasurement.py
@@ -1,5 +1,4 @@
 import logging
-from os import PathLike
 from tempfile import NamedTemporaryFile
 from typing import Optional
 
@@ -19,20 +18,20 @@ from lnst.RecipeCommon.Perf.Measurements.BaseMeasurement import BaseMeasurement
 class TcRunInstance:
     MAX_ID = 0xFFFF
 
-    def __init__(self, device: Device, num_rules: int, instance_id:int):
+    def __init__(self, device: Device, num_rules: int, instance_id: int, batchfile_path: Optional[str] = None):
         self._device = device
         self._num_rules = num_rules
         self._instance_id = instance_id
-        self._batchfile_path: PathLike = None
+
+        if batchfile_path is None:
+            batchfile_path = self._generate_batchfile()
+        self._batchfile_path = batchfile_path
+
         self.validate()
 
     @property
     def batchfile_path(self):
         return self._batchfile_path
-
-    @batchfile_path.setter
-    def batchfile_path(self, p: PathLike):
-        self._batchfile_path = p
 
     def validate(self):
         if self._instance_id > self.MAX_ID:
@@ -68,7 +67,7 @@ class TcRunInstance:
     def end_mac(self) -> str:
         return f"{self.pool_oui}:ff:ff:ff"
 
-    def generate_batchfile(self) -> str:
+    def _generate_batchfile(self) -> str:
         logging.info(f"Generating tc batchfile for {self.device.name} instance {self._instance_id} num_rules={self.num_rules}")
         rules = self.generate_rules()
         with NamedTemporaryFile(
@@ -78,7 +77,6 @@ class TcRunInstance:
                 f.write(r)
 
             remote_path = self.host.copy_file_to_machine(f.name)
-        self.batchfile_path = remote_path
         logging.info(f"tc batchfile written to {self.host.hostname}:{remote_path}")
         return remote_path
 
@@ -154,9 +152,8 @@ class TcRunMeasurement(BaseMeasurement):
             self._running_jobs.append(job)
 
     def _prepare_jobs(self) -> list[Job]:
-        batchfiles = [i.generate_batchfile() for i in self.instance_configs]
         params: dict = {
-            "batchfiles": batchfiles,
+            "batchfiles": [i.batchfile_path for i in self.instance_configs],
         }
         if self._cpu_bind is not None:
             params["cpu_bind"] = self._cpu_bind
--- a/lnst/Recipes/ENRT/TrafficControlRecipe.py
+++ b/lnst/Recipes/ENRT/TrafficControlRecipe.py
@@ -3,7 +3,7 @@ import time
 from contextlib import contextmanager
 
 from lnst.Common.LnstError import LnstError
-from lnst.Common.Parameters import StrParam, IntParam, ChoiceParam
+from lnst.Common.Parameters import ListParam, StrParam, IntParam, ChoiceParam
 from lnst.Controller import HostReq, DeviceReq, RecipeParam
 from lnst.Controller.Namespace import Namespace
 from lnst.RecipeCommon import BaseResultEvaluator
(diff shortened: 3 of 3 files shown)


lnst/Tests/TrafficControl.py
```diff
@@ -31,18 +35,44 @@ def run(self) -> bool:
 
     async def run_instances(self) -> list[dict]:
         tc_exec = shutil.which("tc")
+        cpu_bind = self.params.get("cpu_bind")
+        cpu_bind_gen: Iterator[Optional[list[int]]]
+        if cpu_bind is not None:
```
jtluka, line 40: `is not None` is imo not needed. If the list is empty, the result would be empty generator, or am I wrong?

lnst/Tests/TrafficControl.py
```diff
@@ -31,18 +35,44 @@ def run(self) -> bool:
...
 
-    async def run_tc(self, tc_exec: str, batchfile: str) -> dict[str]:
+    async def run_tc(
+        self,
+        tc_exec: str,
+        batchfile: str,
+        cpu_bind: Optional[list[int]] = None,
+    ) -> dict[str]:
+        args = [tc_exec, "-b", batchfile]
+        if cpu_bind is not None:
```
jtluka, line 63: `is not None` is imo not needed. If the list is empty, the result would be empty generator, or am I wrong?

lnst/Tests/TrafficControl.py
```diff
@@ -31,18 +35,44 @@ def run(self) -> bool:
...
+            args = [
+                "/usr/bin/taskset",
+                "-c",
+                ",".join(map(str, cpu_bind)),
+                *args,
+            ]
+
         start_timestamp = time.time()
         start_time = time.perf_counter()
+        logging.debug(args)
```
jtluka, line 73: Can we make this into something more descriptive. Such debug log entry would be quite hard to find in the code.

jtluka (changes requested):
Otherwise looks good.

jtluka:
I took a look at the beaker job, there are some cpus other than those specified in `cpu_bind` parameter utilized, so wondering where this load comes from. Most likely unrelated to the changes.

jtluka:
> I took a look at the beaker job, there are some cpus other than those specified in `cpu_bind` parameter utilized, so wondering where this load comes from. Most likely unrelated to the changes.

Can't this be from generation of the rules batch files?

jtluka:
> > I took a look at the beaker job, there are some cpus other than those specified in `cpu_bind` parameter utilized, so wondering where this load comes from. Most likely unrelated to the changes.
> 
> Can't this be from generation of the rules batch files?

Checked the code, this is done in `_prepare_jobs()`, that runs before measurements get started.

olichtne:
i don't have any other comments than what jan already mentioned, the pr looks ok to me other than that.

I did look at the beaker job and what I would add is that the TC test measurement is actually very short - about 2 to 5 seconds. When working with measurements this short I think we have to be very careful about the alignment of the additional StatCPUMeasurement as there may be other stray processes that use some cycles here and there and skew the measurements.

As such I think we actually don't really want to actively force the evaluation of the CPU Measurements immediately and maybe just try to monitor it for a bit because it may just be inherently unstable...


On that note, it may help us to implement or work on some solution to get a much more precise startup execution of measurements and calculating their alignment but that's a secondary project...

lnst/Tests/TrafficControl.py
Kuba314: You're right, fixed and also moved to a separate method.

Kuba314:
> Can't this be from generation of the rules batch files?

Yes, definitely. The generation almost takes longer than the actual measurement. I wonder if we can somehow mitigate this. Maybe pin the generation to a separate core? Because I don't think we can signal to the `StatCPUMeasurement` that it shouldn't measure until we're actually running `tc`.

lnst/Tests/TrafficControl.py
Kuba314: Fixed (hopefully).

Kuba314:
Also a note about CPU pinning, `tc` appears to be single threaded, so setting a policy of `all` and pinning cpus to let's say `[0, 1, 2, 3]` will result in all parallel `tc` processes being ran on core `0`. See first `test-runner` of job J:8175517. So I'm not sure if we should even include this policy setting. @jtluka thoughts?

jtluka:
> Also a note about CPU pinning, tc appears to be single threaded ...

The way how LNST wraps it is that it [creates a subprocess](https://github.com/LNST-project/lnst/blob/944170e44973f9d229a451653e365ad7b3409dcb/lnst/Tests/TrafficControl.py#L44) for each batch file defined. So I think this is not an issue. Each sub process can be handled by a separate CPU. I'm not sure why not all cpus are utilized, but I'd keep both policies.

jtluka:
> Yes, definitely. The generation almost takes longer than the actual measurement. ...

Ah, I missed that the StatCPUMeasurement start before the TcRunMeasurement:
```
2023-08-10 13:40:05       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host1), netns(None), CPUStatMonitor(interval=1000)
2023-08-10 13:40:05       (localhost)        -    INFO: Generating tc batchfile for ens5f0 instance 0 num_rules=25000
2023-08-10 13:40:05       (localhost)        -    INFO: tc batchfile written to wsfd-advnetlab55.anl.lab.eng.bos.redhat.com:/tmp/tmpdqjm67yu
2023-08-10 13:40:05       (localhost)        -    INFO: Generating tc batchfile for ens5f0 instance 1 num_rules=25000
2023-08-10 13:40:05       (localhost)        -    INFO: tc batchfile written to wsfd-advnetlab55.anl.lab.eng.bos.redhat.com:/tmp/tmp4zydb6ft
2023-08-10 13:40:05       (localhost)        -    INFO: Generating tc batchfile for ens5f0 instance 2 num_rules=25000
2023-08-10 13:40:06       (localhost)        -    INFO: tc batchfile written to wsfd-advnetlab55.anl.lab.eng.bos.redhat.com:/tmp/tmpl11fi9c4
2023-08-10 13:40:06       (localhost)        -    INFO: Generating tc batchfile for ens5f0 instance 3 num_rules=25000
2023-08-10 13:40:06       (localhost)        -    INFO: tc batchfile written to wsfd-advnetlab55.anl.lab.eng.bos.redhat.com:/tmp/tmp8cc9y0gh
2023-08-10 13:40:06       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host1), netns(None), TrafficControlRunner(batchfiles=['/tmp/tmpdqjm67yu', '/tmp/tmp4zydb6ft', '/tmp/tmpl11fi9c4', '/tmp/tmp8cc9y0gh'], cpu_bind=[0, 1, 2, 3], cpu_bind_policy='all')
```

That is a bit unfortunate. I think the generation of the batch files should be done separately then and passed as argument to the measurement.

@olichtne would you agree?

olichtne:
> @olichtne would you agree?

yes, the file generation isn't an interesting thing to measure in this case.... so it shouldn't be part of the measurement.

The separation can be done simply by doing the file generation during the `init` method of the Measurement imo. Optionally it could be parametrized via the `init` method to either take files pre generated in the recipe, or to generate its own files as well...

jtluka:
There still seem to be some CPU utilization (sometimes ~10% which seems a lot) on CPUs other than expected. Have you checked what could be causing this? For example tasks: T:164623224, T:164623223, T:164623222

Kuba314:
The weird CPU util on other cores seem's to not happen in beaker. See J:8189330.

jtluka:
> The weird CPU util on other cores seem's to not happen in beaker. See J:8189330.

Two notes, this job is a test run where we intentionally increased number of rules by 10x.

I'd like to check the cpu utilzation samples as I think the load is simply spread over time. So, I expect there is a peak at the beginning of the measurement and then it drops to 0%.

jtluka:
> > The weird CPU util on other cores seem's to not happen in beaker. See J:8189330.
> 
> Two notes, this job is a test run where we intentionally increased number of rules by 10x.
> 
> I'd like to check the cpu utilzation samples as I think the load is simply spread over time. So, I expect there is a peak at the beginning of the measurement and then it drops to 0%.

So. here's my findings. I've extracted the raw samples from lrc files of the first two runs in Beaker job and turned them into utilization samples.

For T:164661753, I've checked the samples for 'cpu20', in beaker log it reports 1.18% utilization:

user | nice | system | idle | iowait | irq | sotfirq | steal | guest | guest_nice |   | 100 – idle |   | SUM ALL (should be 100)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 0 | 2 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 97
0 | 0 | 0 | 101 | 0 | 0 | 0 | 0 | 0 | 0 |   | -1 |   | 101
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 1 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 98
0 | 0 | 0 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 99
0 | 0 | 3 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 96
0 | 0 | 8 | 91 | 0 | 0 | 0 | 0 | 0 | 0 |   | 9 |   | 99
0 | 0 | 0 | 98 | 0 | 0 | 0 | 0 | 0 | 0 |   | 2 |   | 98
0 | 0 | 2 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 96
0 | 0 | 0 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 99
0 | 0 | 0 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 99
0 | 0 | 0 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 99
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 98 | 0 | 0 | 0 | 0 | 0 | 0 |   | 2 |   | 98
0 | 0 | 0 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 99
0 | 0 | 1 | 99 | 0 | 0 | 0 | 0 | 0 | 0 |   | 1 |   | 100
0 | 0 | 4 | 90 | 0 | 0 | 0 | 0 | 0 | 0 |   | 10 |   | 94
0 | 0 | 5 | 91 | 0 | 0 | 0 | 0 | 0 | 0 |   | 9 |   | 96
0 | 0 | 3 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 96
0 | 0 | 3 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 97
0 | 0 | 3 | 95 | 0 | 1 | 0 | 0 | 0 | 0 |   | 5 |   | 99
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 3 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 98
0 | 0 | 2 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 96
0 | 0 | 1 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 96
0 | 0 | 2 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 97
0 | 0 | 3 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 97
0 | 0 | 2 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 95
0 | 0 | 3 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 96
0 | 0 | 3 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 96
0 | 0 | 2 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 99
0 | 0 | 3 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 97
0 | 0 | 0 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 97
0 | 0 | 0 | 94 | 0 | 0 | 0 | 0 | 0 | 0 |   | 6 |   | 94
0 | 0 | 2 | 91 | 0 | 0 | 0 | 0 | 0 | 0 |   | 9 |   | 93
0 | 0 | 0 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 97
0 | 0 | 0 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 96
0 | 0 | 1 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 98
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 101 | 0 | 0 | 0 | 0 | 0 | 0 |   | -1 |   | 101
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100

For T:164661754 I checked cpu14 (in beaker log it reports 2.32% utilization:

user | nice | system | idle | iowait | irq | sotfirq | steal | guest | guest_nice |   | 100 – idle |   | SUM ALL (should be 100)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
0 | 0 | 9 | 86 | 0 | 1 | 0 | 0 | 0 | 0 |   | 14 |   | 96
0 | 0 | 1 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 97
0 | 0 | 6 | 90 | 0 | 0 | 0 | 0 | 0 | 0 |   | 10 |   | 96
0 | 0 | 4 | 92 | 0 | 0 | 0 | 0 | 0 | 0 |   | 8 |   | 96
0 | 0 | 2 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 95
0 | 0 | 3 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 96
0 | 0 | 3 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 99
0 | 0 | 1 | 98 | 0 | 0 | 0 | 0 | 0 | 0 |   | 2 |   | 99
0 | 0 | 6 | 85 | 0 | 0 | 0 | 0 | 0 | 0 |   | 15 |   | 91
0 | 0 | 3 | 90 | 0 | 0 | 0 | 0 | 0 | 0 |   | 10 |   | 93
0 | 0 | 3 | 91 | 0 | 0 | 0 | 0 | 0 | 0 |   | 9 |   | 94
0 | 0 | 4 | 89 | 0 | 0 | 1 | 0 | 0 | 0 |   | 11 |   | 94
0 | 0 | 3 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 98
0 | 0 | 3 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 98
0 | 0 | 1 | 93 | 0 | 0 | 0 | 0 | 0 | 0 |   | 7 |   | 94
0 | 0 | 0 | 95 | 0 | 1 | 0 | 0 | 0 | 0 |   | 5 |   | 96
0 | 0 | 3 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 99
0 | 0 | 2 | 92 | 0 | 0 | 0 | 0 | 0 | 0 |   | 8 |   | 94
0 | 0 | 0 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 96
0 | 0 | 0 | 101 | 0 | 0 | 0 | 0 | 0 | 0 |   | -1 |   | 101
0 | 0 | 0 | 95 | 0 | 0 | 0 | 0 | 0 | 0 |   | 5 |   | 95
0 | 0 | 1 | 91 | 0 | 0 | 0 | 0 | 0 | 0 |   | 9 |   | 92
0 | 0 | 2 | 96 | 0 | 0 | 0 | 0 | 0 | 0 |   | 4 |   | 98
0 | 0 | 1 | 98 | 0 | 0 | 0 | 0 | 0 | 0 |   | 2 |   | 99
0 | 0 | 1 | 92 | 0 | 0 | 0 | 0 | 0 | 0 |   | 8 |   | 93
0 | 0 | 0 | 88 | 0 | 0 | 0 | 0 | 0 | 0 |   | 12 |   | 88
0 | 0 | 0 | 97 | 0 | 0 | 0 | 0 | 0 | 0 |   | 3 |   | 97
0 | 0 | 0 | 100 | 0 | 0 | 0 | 0 | 0 | 0 |   | 0 |   | 100

In both cases there are significant increases in CPU utilization ('system' samples).

When I compare this to SimpleNetworkRecipe iperf performance test, the utilization of unused CPUs is typically 0.15 %.

Kuba314:
It appears that `kernel` is doing some work. I saw some `kworker` processes in `top`. Unfortunately, I don't think we can `taskset` kernel workers.

jtluka (approved):

olichtne (approved):

PR #324, opened 2023-08-23
docs/source/conf.py: fix html_theme setting

olichtne:
### Description
This seems to have changed on the readthedocs side and is now I think required to be set.

### Tests

### Reviews

Closes: #


olichtne:
this seems to work and fixes the readthedocs builds. i'll proceed with merging this in.

PR #325, opened 2023-08-23
docs/requirements.txt: add sphinx_rtd_theme

olichtne:
### Description
This may be now a required dependency so that the theme is selected by default when building in readthedocs.

### Tests

### Reviews

Closes: #


olichtne:
this doesn't work... closing.

PR #328, opened 2023-09-05
Remove driver param defaults

Kuba314:
### Description
This allows us to leave out the driver and make mapper match NICs with any driver param.

Keeping driver default in `TrafficControlRecipe` makes sense, because not every NIC supports steering.

### Tests
Not tested.

### Reviews
@LNST-project/lnst-developers


olichtne (approved):
this should be harmless... the test should still be able to run without the parameter defined as we've tried out locally, it is simply ignored from the machine mapper.

this is therefore ok to merge.

PR #332, opened 2023-09-25
Dependencies linux

### LINKED ISSUE #331: Do not import pyroute2 and ethtool until they are required

Axonis:
### Description
Using LNST as dependency when you only require PerfResult structure / importing LRC files it is not required to be installing and using ethtool / pyroute2 dependencies, which are netlink specific - meaning linux specific. 

There are multiple LNST users and developers that are not actively using Linux distros and it would simplify working with LNST as dependency.

### Tests
8351283

### Reviews
@LNST-project/lnst-developers 

Closes: #331

### DIFF
--- a/lnst/Common/NetUtils.py
+++ b/lnst/Common/NetUtils.py
@@ -13,7 +13,6 @@ rpazdera@redhat.com (Radek Pazdera)
 import re
 import socket
 import subprocess
-from pyroute2 import IPRoute
 
 
 def normalize_hwaddr(hwaddr):
@@ -24,6 +23,7 @@ def normalize_hwaddr(hwaddr):
 
 
 def scan_netdevs():
+    from pyroute2 import IPRoute
     scan = []
 
     with IPRoute() as ipr:
--- a/lnst/Devices/Device.py
+++ b/lnst/Devices/Device.py
@@ -12,8 +12,6 @@ olichtne@redhat.com (Ondrej Lichtner)
 """
 
 import re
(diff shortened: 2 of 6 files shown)


enhaut (commented):
Have you tried platform-specific imports e.g. https://stackoverflow.com/a/41097342 ? 

I guess it broke static analysis?

Axonis:
> Have you tried platform-specific imports e.g. https://stackoverflow.com/a/41097342 ?
> 
> I guess it broke static analysis?

I thought about it, but there are no alternatives on darwin kernel that would be just "plug-and-play". 

I am just trying to make LNST work in minimal sense so that darwin kernel is able to load LRC files mostly, not to make it fully functional as that would require quite a lot of effort for minimal to no gain as LNST is currently only being utilized by Linux specific server setups. 

Also I don't see this happening on Windows NT kernel as there are just way too many references to underlaying system (being linux based), which would break and had to be rewritten to support multiple platforms. Instead I am leveraging FreeBSD properties of Darwin, which allow us to get it somewhat functional. 

Finally, we already added somewhat of a lazy loading of imports/dependencies at the time of their need/utilization, instead of being always present and this goes along with that idea.

enhaut (approved):

Axonis:
Tests are passing, SimpleNetwork, L2TP and MPTCP all seem to work after the changes.

olichtne (changes requested):
minor comments

PR #333, opened 2023-10-06
JsonRunSummaryFormatter: format all results

Kuba314:
### Description
We can format all results, because all have a failure value and description that's json serializable. The data just isn't known if the result type is not known so it can't be formatted.

### DIFF
--- a/lnst/Controller/RunSummaryFormatters/JsonRunSummaryFormatter.py
+++ b/lnst/Controller/RunSummaryFormatters/JsonRunSummaryFormatter.py
@@ -4,6 +4,9 @@ import json
 
 from lnst.Controller.Recipe import RecipeRun
 from lnst.Controller.RecipeResults import (
+    DeviceAttrSetResult,
+    DeviceConfigResult,
+    DeviceCreateResult,
     DeviceMethodCallResult,
     JobResult,
     JobStartResult,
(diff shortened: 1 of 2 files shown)


enhaut (approved):

lnst/Controller/RunSummaryFormatters/JsonRunSummaryFormatter.py
```diff
@@ -85,23 +91,20 @@ def _transform_result(self, result: Result) -> Optional[dict]:
...
-                return ret | {
-                    "type": "unknown",
-                    "description": result.description,
-                }
-            else:
-                logging.warning(f"unhandled recipe result type: {repr(result)} with data of type {type(result.data)}")
         else:
-            logging.warning(f"unhandled recipe result type: {repr(result)}")
-            return None
+            if result.data is not None:
```
olichtne, line 132: i think i'd prefer this to have an explicit branch that checks `isinstance(BaseResult)`, ideally that can then include `BaseResult.result` as well.

and i'd keep the `else` branch as well which imo should simply throw an exception to show OBVIOUS issues as this method should never receive an object that doesn't inherit from `BaseResult`.

does this make sense or do you think this is too restrictive?

Kuba314:
I have added a commit including `"measurement_type"` to comparisons of `BaselineEvaluationResult`. I know this is not necessarily related to this MR, but it's needed in downstream baseline handling since this property was taken directly from the LRC file.

lnst/Controller/RunSummaryFormatters/JsonRunSummaryFormatter.py
Kuba314: This contract of "only pass in `BaseResult` objects" is specified in the method signature. This is the caller's responsibility and we shouldn't check for this dynamically.

> ideally that can then include BaseResult.result as well.

This is already done at the start of the method, we always include `str(result.result)` (`"PASS"` or `"FAIL"`) in the result dictionary.

As for the explicitness, I didn't include the isinstance check, because the contract specifies that the variable is most definitely some inheritee of `BaseResult`. Most probably a `Result` instance. The only instance of us creating a `Result` with some data is at https://github.com/LNST-project/lnst/blob/master/lnst/RecipeCommon/PacketAssert.py#L93. All other instances use just the description field as some generic message that should be included as a result.

An explicit `isinstance(result, Result)` branch would be identical to the current `else` branch so I don't really see a point. The only point I see would be exhaustiveness checking done by a type-checker which would warn us about an unhandled result type and the `logging.warning()` call could be removed. [`assert_never`](https://docs.python.org/3/library/typing.html#typing.assert_never) is available only from python3.11 though...

olichtne: > This contract of "only pass in BaseResult objects" is specified in the method signature. This is the caller's responsibility and we shouldn't check for this dynamically.

this is a good argument, and i agree, the only reason i'm thinking of the explicit **Exception** is because we don't have a type check ci and we don't require anyone to typecheck their lnst code at the moment. Meaning that we can break the contract without anything actually breaking. The exception would ensure that this breaks the scenario that introduces this issue.

It would be a very unlikely scenario though so I think I'm ok with keeping it as is.

---

> An explicit isinstance(result, Result) branch

this isn't about the `Result` class... i'm thinking about this in terms of how you usually structure a `switch` statement and applying that to a type hierarchy:

```
switch:
    derived type1:
    derived_type2:
    ....
    base_type: default behaviour
    default: crash because this is something that doesn't fit the type tree
```

and the point of the `base type` here is that it provides the default functionality for any newly added derived types which will be added in the future which we don't want to crash.

and the final else/default clause makes sure that any incompatible types crash.

i feel like this is safer, though i admit that the likelihood of this happening is low.

---

> This is already done at the start of the method, we always include str(result.result) ("PASS" or "FAIL") in the result dictionary.

makes sense, was just looking at the diff and didn't notice this.

Kuba314: Everything inheriting from `BaseResult` will still have `result` and `description`. I hope noone will be deleting these down the hierarchy. The default case is catches even newly created types that inherit from `BaseRecipe` and thus they automatically get formatted. If the developer wanted to format them specially, they could do that. A warning that the result contained `data` that were not formatted is IMO good enough.

Ofc the other approach would be prohibiting this and raising an Exception upon any unrecognized result. I don't have a preference here, but in that case we shouldn't have a default branch. I'm resolving this. Feel free to unresolve if you don't agree.

PR #334, opened 2023-10-10
ovs dpdk fixes for mellanox

jtluka:
### Description

Current implementation of Ovs_DPDK_PvP recipe works only for ixgbe/intel nics. Using the same setup with Mellanox cards requires sligh adjustments that include:
* no `driverctl set-override`
* use `--no-ofed-check` for TRex, the ofed utils are not required for this

### Tests

Tested on i40e and mlx5_core NICs.


olichtne (approved):
looks ok

PR #336, opened 2023-10-16
XDPMeasurementResults: fixed describe method return value type

enhaut:
### Description

All other `MeasurementResults.describe()` methods returns already formatted string, so our internal tooling expects string not the list of strings, and therefore it crashes when treating list as a string.

### Tests
Before: `J:8418399`  
After: `J:8433522`

### Reviews
@olichtne @jtluka


olichtne (approved):
looks ok

PR #340, opened 2023-10-24
BaseLACPRecipe.test_wide_conf(): call grand parent's test_wide_conf() method

enhaut:
### Description  
BaseLACPRecipe reimplements same behaviour as DoubleBondRecipe but adds multiple IPs. Because direct parent also implements bond device setup we need to bypass direct parent's `test_wide_conf()` method to prevent LNST crashing on calling bond device setup twice.

### Tests
Excluding changes: `J:8469701`  
Including changes: `J:8470332`

### DIFF
--- a/lnst/Recipes/ENRT/BaseLACPRecipe.py
+++ b/lnst/Recipes/ENRT/BaseLACPRecipe.py
@@ -39,7 +39,7 @@ class BaseLACPRecipe(DoubleBondRecipe):
         it calls switch configuration method.
         """
         host1, host2 = self.matched.host1, self.matched.host2
-        config = super().test_wide_configuration()
+        config = super(DoubleBondRecipe, self).test_wide_configuration()
 
         ipv4_addr = interface_addresses(self.params.net_ipv4)
         ipv6_addr = interface_addresses(self.params.net_ipv6)


Kuba314 (approved):

olichtne:
shouldn't this recipe inheritance be flipped around? shouldn't it be:

* `BaseLACPRecipe(baremetal enrt and the other generic bases)`

* `DoubleBondRecipe(BaseLACPRecipe)`

* `DellLACPRecipe(BaseLACPRecipe)`

?

Basically if the only thing that is "changed" between the DoubleBond and the BaseLACP recipe is the test wide configuration parts relating to the mode then IMO this should just be a shared base class that reuses the code.

IMO having two ip addresses configured by default in the base recipe shouldn't be a problem for the `DoubleBondRecipe` as long as we don't change the ping/perf endpoints for it.... (though i guess it will shift the threshold indexes)

enhaut:
> shouldn't this recipe inheritance be flipped around? shouldn't it be:
> 
> 
> 
> * `BaseLACPRecipe(baremetal enrt and the other generic bases)`
> 
> 
> 
> * `DoubleBondRecipe(BaseLACPRecipe)`
> 
> 
> 
> * `DellLACPRecipe(BaseLACPRecipe)`
> 
> 
> 
> ?
> 

No, LACP is just a type of bond. There are couple of other bonds, e.g. `active-backup`, `balance-rr`,... 


> IMO having two ip addresses configured by default in the base recipe shouldn't be a problem for the `DoubleBondRecipe` as long as we don't change the ping/perf endpoints for it.... (though i guess it will shift the threshold indexes)

Yes, I wanted to avoid shifting result indexes, so I just left it as is.

olichtne:
> No, LACP is just a type of bond. There are couple of other bonds, e.g. active-backup, balance-rr,...

yes... it's a `mode` which is a shared parameter, so technically the same recipe could work for all modes...

anyway, since this will shift threshold indexes, let's ignore this for now. let's refactor this once we get rid of the threshold indexes.

Issue #345, opened 2023-11-06
create an API for the switchdev VFs


jtluka:
We have some tests that are testing VFs in https://github.com/LNST-project/lnst/blob/be9f9a0462b5192382aaceb8f5bb338a4ca03ae8/lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py#L111

The code for creating the VFs can be made into an API, so that we can interact with the VFs as ordinary `Device`s.

The overall idea I currently have is:
```
host1.eth0 = DeviceReq()

sd = SwitchDevice(host1.eth0) # turns the device into 'mode switchdev', otherwise is just a Device
vf_rep_pairs = sd.create_vfs(cnt=2) # does echo $cnt > /sys/class/net/.../sriov_numvfs, returns list of (vf, vf_representor) pairs

vf_rep_pair = vf_rep_pairs[0]

type(vf_rep_pair.vf)
Device

type(vf_rep_pair.vf_rep)
Device

# move vf to netns
host1.newns = NetNamespace("lnst")
host1.newns.vf = vf_rep_pair.vf
```

Axonis:
Makes sense, I had similar idea but wanted to go even more abstract, where we would also have e.g. SRIOVManager, that would be able to create TC rules above VF/VF-Repr/PF devices. 

Also is Device class sufficient for this use case?

jtluka:
For all of this to work we need to modify the InterfaceManager to handle the netlink messages related to VFs.

Here's the analysis of the netlink msgs when 2 vf devices are created:
* for vf device, the message contains:
```
    1st message                                                                 
        ('IFLA_PHYS_PORT_NAME', 'pf0vf0'),                                      
        ('IFLA_PHYS_SWITCH_ID', '10:f7:d5:fe:ff:1a:3d:e4'),                     
        ('IFLA_PARENT_DEV_BUS_NAME', 'pci')                                     
        ('IFLA_PARENT_DEV_NAME', '0000:01:00.0') same as PF device              
                                                                                
    2nd message                                                                 
        ('IFLA_PHYS_PORT_NAME', 'pf0vf0'),                                      
        ('IFLA_PROP_LIST', {'attrs': [('IFLA_ALT_IFNAME', 'enp1s0f0npf0vf0')]}, 32768),
        ('IFLA_PARENT_DEV_NAME', '0000:01:00.0'),                               
                                                                                
    2nd message for 2nd vf                                                      
[...]

Issue #344, opened 2023-11-02
SRIOVNetnsTcRecipes: device needs to be up to create VFs for specific drivers 


jtluka:
For bnxt_en devices, the PF device needs to be up.

The SRIOV base recipe class should be updated and bring the device up. We should check if this impacts other drivers such as mlx5 or ice.

jtluka:
Note that having the device up is not an issue for Meallanox NICs.

PR #346, opened 2023-11-15
VF/Switchdev abstraction

jtluka:
### Description

This MR extends the Device class with methods and properties required to configure and use vfs of an ethernet device in switchdev mode.

User would typically configure the device as follows (see also the change to BaseSRIOVNetnsTcRecipe):
```python
class SwitchdevRecipe(BaseRecipe):
    host1 = HostReq()
    host1.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))

    host2 = HostReq()
    host2.eth0 = DeviceReq(label="net1", driver=RecipeParam("driver"))

    def run(self):

        # create virtual functions
        for host in [host1, host2]:
            host.eth0.eswitch_mode = "switchdev"
            host.eth0.create_vfs(1)
            vf_device = host.eth0.vf_devices[0]
            vf_rep_device = host.eth0.vf_rep_devices[0]

            host.map_device("vf_representor_eth0", {"ifname": vf_rep_device.name})

            host.run(f"ethtool -K {host.vf_representor_eth0.name} hw-tc-offload on")
            host.run(f"ethtool -K {host.eth0.name} hw-tc-offload on")
```

Known issues:
* for i40e NICs, the code does not work as there are no representors by the NIC design
* for bnxt_en NICs, the code does not work on RHEL distros due to missing patch `[PATCH net-next v2] bnxt_en: Allow to set switchdev mode without existing VFs`

### Tests

J:8572626 J:8572625 J:8572624 J:8572623

### Reviews

@olichtne @Axonis 

Closes: #345  #344

### DIFF
--- a/lnst/Devices/Device.py
+++ b/lnst/Devices/Device.py
@@ -57,6 +57,21 @@ def sriov_capable(func):
 
     return wrapper
 
+def switchdev_capable(func):
+    """
+    Decorator for Device class methods that checks whether the Device can be
+    used in switchdev mode.
+    """
+    def wrapper(self, *args, **kwargs):
+        try:
+            _ = self.eswitch_mode
+        except DeviceFeatureNotSupported:
+            raise
+
+        return func(self, *args, **kwargs)
+
+    return wrapper
+
 class Device(object, metaclass=DeviceMeta):
     """The base Device class
 
@@ -904,6 +919,38 @@ class Device(object, metaclass=DeviceMeta):
         if (rx_val, tx_val) != (None, None):
             self._write_pause_frames(rx_val, tx_val)
 
+    @property
+    def eswitch_mode(self):
+        try:
+            # TODO: do this through device._devlink?
+            stdout, _ = exec_cmd(f"devlink dev eswitch show pci/{self.bus_info}")
+        except ExecCmdFail as e:
+            if e.get_stderr().find("Operation not supported") > -1 or e.get_stderr().find("No such device"):
+                raise DeviceFeatureNotSupported(f"Device {self.name} not compatible with switchdev")
+            else:
+                raise DeviceError(
+                    f"Error while querying device {self.name} eswitch mode:\n{e.get_stderr()}"
+                )
+
+        m = re.search("mode (legacy|switchdev)", stdout)
+        try:
+            return m.group(1)
+        except AttributeError:
+            raise DeviceError(f"Could not parse device {self.name} eswitch mode")
+
+    @eswitch_mode.setter
+    def eswitch_mode(self, mode):
+        if self.eswitch_mode == mode:
+            return
+
+        try:
+            # TODO: do this through device._devlink?
+            exec_cmd(f"devlink dev eswitch set pci/{self.bus_info} mode {mode}")
+        except ExecCmdFail as e:
+            raise DeviceError(
+                f"Error while setting device {self.name} eswitch mode:\n{e.get_stderr()}"
+            )
+
     @sriov_capable
     def create_vfs(self, number_of_vfs=1):
         try:
@@ -920,7 +967,23 @@ class Device(object, metaclass=DeviceMeta):
         for vf_dev in (vf_devices := self._get_vf_devices()):
             vf_dev._enable()
 
-        return vf_devices
+        try:
+            eswitch_mode = self.eswitch_mode
+        except DeviceFeatureNotSupported:
+            return vf_devices, None
+
+        if eswitch_mode == "switchdev":
+            # wait for representor devices to appear on the netlink
+            if not self._wait_for_vf_rep_devices(number_of_vfs, 10):
+                raise DeviceError(f"Error while waiting for vf_reps for {self.name}")
+
+            for vf_rep_dev in (vf_rep_devices := self._get_vf_rep_devices()):
+                vf_rep_dev._enable()
+
+            return vf_devices, vf_rep_devices
+
+        # for any other cases, just return vf devices
+        return vf_devices, None
 
     def _wait_for_vf_devices(self, vfs_count, timeout):
         if timeout > 0:
@@ -942,6 +1005,29 @@ class Device(object, metaclass=DeviceMeta):
             raise
 
         logging.info(f"vfs on PF {self.name} successfully created")
+
+        return True
+
+    def _wait_for_vf_rep_devices(self, vfs_count, timeout):
+        if timeout > 0:
+            logging.info(f"Waiting for vf_rep(s) creation on PF {self.name} for {timeout} seconds")
+        elif timeout == 0:
+            logging.info(f"Waiting for vf_rep(s) creation on PF {self.name}")
+
+        def condition():
+            return all(
+                [
+                    self._vf_rep_is_ready(vf_index) for vf_index in range(0, vfs_count)
+                ]
+            )
+
+        try:
+            wait_for_condition(condition, timeout=timeout)
+        except TimeoutError:
+            logging.info(f"Timeout while waiting for vf_reps creation on PF {self.name}")
+            raise
+
+        logging.info(f"vf_reps on PF {self.name} successfully created")
 
         return True
 
@@ -967,10 +1053,31 @@ class Device(object, metaclass=DeviceMeta):
 
         return True
 
+    def _vf_rep_is_ready(self, vf_index):
+        logging.debug(f"Waiting for {self.name} vf_representor with index {vf_index}")
+
+        # search for vf representor
+        vf_rep_device = self._get_vf_representor(vf_index)
+
+        if not vf_rep_device:
+            logging.debug(f"vf representor {vf_index} for {self.name} not ready")
+            return False
+
+        logging.debug(f"vf representor {vf_index} for {self.name} ready")
+
+        return True
+
     @sriov_capable
     def _get_vf_devices(self):
         return [
             self._get_vf(vf_index)
+            for vf_index in range(self._get_vf_count())
+        ]
+
+    @switchdev_capable
+    def _get_vf_rep_devices(self):
+        return [
+            self._get_vf_representor(vf_index)
             for vf_index in range(self._get_vf_count())
         ]
 
@@ -979,6 +1086,38 @@ class Device(object, metaclass=DeviceMeta):
         stdout, _ = exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs")
 
         return int(stdout)
+
+    def _get_vf_representor(self, vf_index: int):
+        pf_number = int(self.phys_port_name[1:])
+
+        vf_rep_device_phys_port_name = f"pf{pf_number}vf{vf_index}"
+        all_devices = self._if_manager.get_devices()
+
+        """
+        netlink message content for vf representor devices
+        1st message
+        ('IFLA_PHYS_PORT_NAME', 'pf0vf0'),
+        ('IFLA_PHYS_SWITCH_ID', '10:f7:d5:fe:ff:1a:3d:e4'),
+        ('IFLA_PARENT_DEV_BUS_NAME', 'pci')
+        ('IFLA_PARENT_DEV_NAME', '0000:01:00.0') is same as PF device
+
+        2nd message, contains alternative device name derived from the PF device
+        ('IFLA_PHYS_PORT_NAME', 'pf0vf0'),
+        ('IFLA_PROP_LIST', {'attrs': [('IFLA_ALT_IFNAME', 'enp1s0f0npf0vf0')]}, 32768),
+        ('IFLA_PARENT_DEV_NAME', '0000:01:00.0'),
+        """
+        vf_reps_match = [
+            dev for dev in all_devices if (
+                dev.phys_port_name == vf_rep_device_phys_port_name and
+                dev.parent_dev_bus_name == self.parent_dev_bus_name and
+                dev.parent_dev_name == self.parent_dev_name
+            )
+        ]
+
+        try:
+            return vf_reps_match[0]
+        except IndexError:
+            return None
 
     def _get_vf(self, vf_index: int):
         all_devices = self._if_manager.get_devices()
@@ -1007,6 +1146,18 @@ class Device(object, metaclass=DeviceMeta):
             return vfs_match[0]
         except IndexError:
             return None
+
+    @property
+    def phys_port_name(self):
+        return self._nl_msg.get("IFLA_PHYS_PORT_NAME")
+
+    @property
+    def parent_dev_bus_name(self):
+        return self._nl_msg.get("IFLA_PARENT_DEV_BUS_NAME")
+
+    @property
+    def parent_dev_name(self):
+        return self._nl_msg.get("IFLA_PARENT_DEV_NAME")
 
     @property
     def alt_if_names(self):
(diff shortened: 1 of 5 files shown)


jtluka:
Tests done through beaker job ran without issues.

I'd hold the MR until I get a feedback from Ivan whether the use case when VFs are created before setting the switchdev mode is valid and expected. If it is valid this will need to be reworked.

jtluka:
Just a note that here's the RFC for unfication of PF/VF/VF_REP names: https://patchwork.ozlabs.org/project/netdev/patch/20170725051344.2040-1-jakub.kicinski@netronome.com/

jtluka:
> I'd hold the MR until I get a feedback from Ivan whether the use case when VFs are created before setting the switchdev mode is valid and expected. If it is valid this will need to be reworked.

Apparently it is possible to create VFs without setting the switchdev mode - this is typical scenario where tuntap is substituted with virtual function devices.

The MR should be updated and create_vfs should be updated to support both non-switchdev and switchdev scenarios.

jtluka:
Note, for mlx5 it is not possible to create vfs and then set the switchdev mode:
```
[root@wsfd-advnetlab57 ~]# echo 1 > /sys/class/net/ens5f0/device/sriov_numvfs 
[root@wsfd-advnetlab57 ~]# cat /sys/class/net/ens5f0/device/sriov_numvfs 
1
[root@wsfd-advnetlab57 ~]# devlink dev eswitch set pci/0000:81:00.0 mode switchdev
Error: mlx5_core: Inline mode is different between vports.
kernel answers: Invalid argument
[root@wsfd-advnetlab57 ~]# dmesg | tail
[  165.786118] mlx5_core 0000:81:00.2 ens5f0v0: renamed from eth0
[  165.810195] IPv6: ADDRCONF(NETDEV_UP): ens5f0v0: link is not ready
[  165.955773] mlx5_core 0000:81:00.2 ens5f0v0: Link up
[  165.962261] IPv6: ADDRCONF(NETDEV_UP): ens5f0v0: link is not ready
[  165.968699] IPv6: ADDRCONF(NETDEV_CHANGE): ens5f0v0: link becomes ready
[  264.102362] mlx5_core 0000:81:00.0: E-Switch: Disable: mode(LEGACY), nvfs(1), active vports(2)
[  264.121696] mlx5_core 0000:81:00.2 ens5f0v0: Link down
[  265.831026] mlx5_core 0000:81:00.0: mlx5_cmd_out_err:806:(pid 2226): CREATE_FLOW_TABLE(0x930) op_mod(0x0) failed, status bad resource state(0x9), syndrome (0x98afbb), err(-22)
[  265.846675] mlx5_core 0000:81:00.0: mlx5_cmd_dr_create_flow_table:86:(pid 2226): Failed creating dr flow_table
[  265.856671] mlx5_core 0000:81:00.0: E-Switch: Failed to create slow path FDB Table err -22
```

Axonis:
With all the differences between drivers/NICs, I think it makes sense to split all VF/SRIOV related actions to separate commands. What do you think?

jtluka:
> With all the differences between drivers/NICs, I think it makes sense to split all VF/SRIOV related actions to separate commands. What do you think?

The only difference is the bnxt_en driver. The fact that you can't set switchdevice mode unless you have virtual functions created is considered a bug that has been already fixed in upstream kernel. I asked Ivan to update RHEL kernels.

I could expose more Device methods, but you would have to be more specific. Eventually these methods can be exposed with followup patches.

jtluka:
> > With all the differences between drivers/NICs, I think it makes sense to split all VF/SRIOV related actions to separate commands. What do you think?
> 
> The only difference is the bnxt_en driver. The fact that you can't set switchdevice mode unless you have virtual functions created is considered a bug that has been already fixed in upstream kernel. I asked Ivan to update RHEL kernels.
> 

I've checked with Ivan and there's already a MR for RHEL9 kernel to update the bnxt_en driver, so there's no other difference across the drivers.

Axonis:
> > With all the differences between drivers/NICs, I think it makes sense to split all VF/SRIOV related actions to separate commands. What do you think?
> 
> The only difference is the bnxt_en driver. The fact that you can't set switchdevice mode unless you have virtual functions created is considered a bug that has been already fixed in upstream kernel. I asked Ivan to update RHEL kernels.
> 
> I could expose more Device methods, but you would have to be more specific. Eventually these methods can be exposed with followup patches.
> 
> The example in the MR description was a bit out of date, maybe that would fit what you're asking for. I updated it, so please check.

I'm fine with current implementation, makes sense.

lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py
```diff
@@ -108,19 +104,15 @@ def test_wide_configuration(self):
...
-            time.sleep(2)
-            host.run(f"echo 1 > /sys/class/net/{host.eth0.name}/device/sriov_numvfs")
-            time.sleep(3)
-
-            vf_ifname = dict(ifname=f"{host.eth0.name}{self.params.vf_suffix}")
-            host.map_device("vf_eth0", vf_ifname)
+            host.eth0.eswitch_mode = "switchdev"
+            host.eth0.create_vfs(1)
+            vf_device = host.eth0.vf_devices[0]
+            vf_rep_device = host.eth0.vf_rep_devices[0]
```
Axonis, line 110: What about using dataclass to hold vf device + vf representor device (Optional)

```
from dataclasses import dataclass
from typing import Optional

@dataclass
class SRIOVDevices:
    virtual_function: Device
    virtual_function_representor: Optional[Device] = None

vf_only = SRIOVDevices(virtual_function=vf)
device_pair = SRIOVDevices(virtual_function=vf, virtual_function_representor=vf_repr)
```

This object be returned in a list if there are multiple VF pairs / VFs created.
Also this allows us to implement or move methods that can be used for checks/creation of VFs, depending how you implementing it.

jtluka:
Based on a discussion with LNST team I plan to update the code.

The main issue detected was with `vf_devices` property in following use case:
```
host1.phys_dev.create_vfs()
vf_device = phys_dev.vf_devices[0]
host1.newns = NetNamespace("lnst") 
host.newns.vf_eth0 = vf_device
...
vf_device = phys_dev.vf_devices[0]
```

The last statement will not work because `phys_dev` is in different namespace than the vf device and would not be able to fetch it from netlink messages during rescan of available devices.

We came to conclusion that `vf_devices` and `vf_rep_devices` properties will be removed and instead the `create_vfs()` method will return the vf/vf_rep devices directly.

lnst/Devices/Device.py
```diff
@@ -878,6 +901,125 @@ def restore_pause_frames(self):
...
+            raise
+
+        logging.info(f"vfs on PF {self.name} successfully created")
+
+        return True
+
+    @sriov_capable
+    def delete_vfs(self):
+        if not self.state != "up":
+            self.up_and_wait()
```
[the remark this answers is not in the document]
jtluka: So, for most of the drivers device does not have to be up for creating/deleting virtual functions. Only one driver (I think it's bnxt_en) requires the device to be up. Instead of dealing with this in recipes I included this in the device method.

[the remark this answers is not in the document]
olichtne: i think that is even more of an argument for this to be outside of the Device class.

Basically this makes it a sort of implicit change to a Test Recipe - if you run it with driver X it does some actions, if you run it with driver Y it automatically does a bunch more functions....

Using your example from the start of the PR, the main problem being this:

```pytthon
    def run(self):

        # create virtual functions
        for host in [host1, host2]:
            host.eth0.eswitch_mode = "switchdev"
            host.eth0.create_vfs(1)
            # some action after create_vfs
```

when running with driver X, you will have this "some other action" run with the device in state `down` and with driver Y, the same action will run with the device in state `up` and when developing the recipe you don't immediately see this as an "non deterministic action" that can change the device state.

So... depending on what this "other action" is, you will want to ENSURE that the device is in a predictable state anyway and  you'll be including these `up/down` calls as necessary for these actions anyway:

```pytthon
    def run(self):

        # create virtual functions
        for host in [host1, host2]:
            host.eth0.eswitch_mode = "switchdev"
            host.eth0.create_vfs(1)
            host.eth0.up()
            # some action that requires the device be UP

            host.eth0.down()
            # some action that requires the device be DOWN
```

and so if you're already doing this for every action... you may as well do this for the `create_vfs` action as well. This helps keep consistency, but also... it simplifies the thinking process of "is this action changing the device state in more than one way? and what is the current state of the device?"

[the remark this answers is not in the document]
jtluka: Pushed changes that remove the `up_and_wait()` from `create_vfs()` and `delete_vfs()`.

Updated `cleanup()` with explicit `up()` call.

lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py
```diff
@@ -106,36 +131,36 @@ def test_wide_configuration(self):
         host1, host2 = self.matched.host1, self.matched.host2
         config = super().test_wide_configuration()
 
+        # TODO: ideally we would store the sriov_devices in EnrtConfiguration
+        # instance, but the devices are used also in overriden perf_test()
+        # method that does not have reference to EnrtConfiguration.
+        # Instead we will store as class instance attribute
```
olichtne, line 137: this comment isn't accurate btw... `perf_test` receives a `lnst.RecipeCommon.Perf.Recipe.RecipeConf` which has an attribute of `parent_recipe_config`.

The `RecipeConf` object is created here: https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/BaseEnrtRecipe.py#L449 and contains a copy of the **CURRENT** `EnrtConfiguration`

the solution to store the instance attribute of `sriov_devices` is still ok IMO, just a note that the comment itself isn't accurate.

lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py
olichtne: the location makes sense to me, it is right now used just for the `BaseSRIOVNetnsTcRecipe` hierarchy... we can move it once it becomes more generically used. In case we want to make some generalization adjustments for that...

olichtne (approved):
looks ok to me.

lnst/Recipes/ENRT/SRIOVNetnsTcRecipe.py
```diff
@@ -37,102 +37,106 @@ def add_tc_filter_rules(self, config):
...
                   f"protocol arp ingress flower "
-                  f"src_mac {host2.newns.vf_eth0.hwaddr} "
+                  f"src_mac {host2_vf_dev.hwaddr} "
                   f"dst_mac FF:FF:FF:FF:FF:FF "
                   f"action mirred egress redirect dev {host2.eth0.name}")
 
     @property
     def pause_frames_dev_list(self):
-        return [self.matched.host1.newns.vf_eth0, self.matched.host2.newns.vf_eth0]
+        return [self.sriov_devices[self.matched.host1].vfs[0], self.sriov_devices[self.matched.host2].vfs[0]]
```
Axonis, line 122: Thinking about this, also in the future... does it make sense to have `SRIOV devices > Host > PF/VF/VF_REPR` instead of `Host > SRIOV devices > PF/VF/VF_REPR`? 

Only possible upside I see is being able to create algorithm form Tc rules creation.

lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py
jtluka: Thanks, I'll remove the comment. I was wondering if this is hidden somewhere.

lnst/Devices/Device.py
```diff
@@ -380,10 +411,20 @@ def cleanup(self):
         if self.master:
             self.master.cleanup()
             self.master = None
+
+        try:
+            if not self.state != "up":
```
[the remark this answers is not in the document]
jtluka: Thanks, it indeed looks weird, I will take a look and most likely change it.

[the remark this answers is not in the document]
jtluka: And actually it is a bug ... I will remove `not`.

lnst/Devices/Device.py
```diff
@@ -41,6 +42,36 @@ def __instancecheck__(self, other):
...
             return super(DeviceMeta, self).__instancecheck__(other)
 
+def sriov_capable(func):
+    """
+    Decorator for Device class methods that checks whether the Device is
+    capable of SR-IOV.
+    """
+    def wrapper(self, *args, **kwargs):
+        try:
+            exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs")
```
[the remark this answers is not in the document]
jtluka: > But on the other hand I understand that you basically only work with /sys/class/net... so it is probably fine

I'm not aware of other way how to configure vfs on a sriov capable device, so checking this file is not based on assumption, but simply a pre-requisite. If this file does not exist, the device can't create vfs. Sure, it may also mean that the device does not have the feature enabled - but in that case, it simply is not capable of using the sriov features, iow, is not sriov capable from the current device status point of view.

>  It could possibly make more sense to check lspci or similar bus info to see whether SRIOV is supported.

Not quite sure how I could fetch this info. Also, what options do we have if the device can have the SRIOV feature turned on but is disabled. We would have to enable such feature either in BIOS or using a driver specific tools which is imo out of scope.

Maybe I'm missing something, please clarify if that's the case.

lnst/Recipes/ENRT/SRIOVNetnsTcRecipe.py
jtluka: Do you mean to change this so that we can use following?
```
        return [self.matched.host1.sriov_devices.vfs[0], self.matched.host2.sriov_devices.vfs[0]
```

lnst/Recipes/ENRT/BaseSRIOVNetnsTcRecipe.py
jtluka: Removed.

lnst/Devices/Device.py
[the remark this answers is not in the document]
Axonis: I can't find my comments from issues where I was verifying this, but I remember using `lspci -vv` on the given bus info to see whether it had "single root" capabilities. This ensured that BIOS settings for the card were enabled and that the presense of the `sriov_numvfs`.

However just the last assumption and existence should be enough. Disregard this.

lnst/Recipes/ENRT/SRIOVNetnsTcRecipe.py
jtluka: Hmm, this actually makes sense. I've pushed a change where the devices are stored directly in Host instances, see https://github.com/LNST-project/lnst/pull/346/commits/59aa3d04038f21ea8700fe0abf15861c042882f4

@Axonis @olichtne please take a look.

olichtne: yeah, this looks better.

Axonis: yeah, ack.

jtluka:
New test jobs: J:8640504 J:8640505 J:8640506 J:8640507

jtluka:
> New test jobs: J:8640504 J:8640505 J:8640506 J:8640507

These tests passed. Note that bnxt_en driver contains bug where the vfs need to be created before moving to switchdev mode.

Issue #201, opened 2020-11-04
Tests/Iperf.py: res_data['msg'] is assigned twice discarding the previous content


jtluka:
https://github.com/LNST-project/lnst/blob/66f0767c32f1b07c9326c700cde82716f756731c/lnst/Tests/Iperf.py#L45

olichtne:
Yeah, the first assignment at https://github.com/LNST-project/lnst/blob/66f0767c32f1b07c9326c700cde82716f756731c/lnst/Tests/Iperf.py#L40 can probably be safely removed. The stderr value is reported in the result, I don't think there's a need to add a message to it...

PR #349, opened 2023-11-21
Reorganize Iperf error reporting

Kuba314:
### Description
Improve iperf reporting and reorganize the code a bit.

### Tests
`J:8583736` - second recipeset contains debug logs.

### Reviews
@jtluka 

Closes: #201

### DIFF
--- a/lnst/Tests/Iperf.py
+++ b/lnst/Tests/Iperf.py
@@ -1,4 +1,5 @@
 import logging
+import signal
 import subprocess
 import json
 from json.decoder import JSONDecodeError
@@ -34,57 +35,52 @@ class IperfBase(BaseTestModule):
 
         try:
             stdout, stderr = server.communicate()
-            stdout = stdout.decode()
-            stderr = stderr.decode()
         except KeyboardInterrupt:
-            pass
+            server.send_signal(signal.SIGINT)
+        finally:
+            stdout, stderr = server.communicate()
+            stdout = stdout.decode().strip()
+            stderr = stderr.decode().strip()
+
+        if server.returncode > 0:
+            msg = f"iperf {self._role} returncode = {server.returncode}"
+            logging.error(msg)
+            logging.error(f"iperf stderr: {stderr}")
+            logging.debug(f"iperf stdout: {stdout}")
+            self._res_data["msg"] = msg
+            self._res_data["stderr"] = stderr
+            return False
 
         try:
             self._res_data["data"] = json.loads(stdout)
         except JSONDecodeError:
-            self._res_data["msg"] = "Error while parsing the iperf json output"
-            self._res_data["data"] = stdout
+            msg = "Error while parsing the iperf json output"
+            logging.error(msg)
+            logging.error(f"iperf stderr: {stderr}")
+            logging.debug(f"iperf stdout: {stdout}")
+            self._res_data["msg"] = msg
             self._res_data["stderr"] = stderr
-            logging.error(self._res_data["msg"])
             return False
 
-        try:
-            self._check_json_sanity()
-        except:
-            self._res_data["msg"] = "Iperf provided incomplete json data"
-            self._res_data["data"] = stdout
+        if not self._is_json_complete(self._res_data["data"]):
+            msg = "Iperf provided incomplete json data"
+            logging.error(msg)
+            logging.error(f"iperf stderr: {stderr}")
+            logging.debug(f"iperf stdout: {stdout}")
+            self._res_data["msg"] = msg
             self._res_data["stderr"] = stderr
-            logging.error(self._res_data["msg"])
-            return False
-
-        self._res_data["stderr"] = stderr
-
-        if stderr != "":
-            self._res_data["msg"] = "errors reported by iperf"
-            logging.error(self._res_data["msg"])
-            logging.error(self._res_data["stderr"])
-
-        if server.returncode > 0:
-            self._res_data["msg"] = "{} returncode = {}".format(
-                    self._role, server.returncode)
-            logging.error(self._res_data["msg"])
             return False
 
         return True
 
-    def _check_json_sanity(self):
-        data = self._res_data["data"]
-        if "start" not in data:
-            raise Exception()
-
-        if "end" not in data:
-            raise Exception()
-
-        if len(data["intervals"]) == 0:
-            raise Exception()
-
-        if "streams" not in data["end"]:
-            raise Exception
+    @staticmethod
+    def _is_json_complete(data: dict) -> bool:
+        return (
+            "start" in data
+            and "end" in data
+            and len(data["intervals"]) > 0
+            and "streams" in data["end"]
+        )
 
 
 class IperfServer(IperfBase):
--- a/lnst/Agent/Job.py
+++ b/lnst/Agent/Job.py
@@ -96,7 +96,7 @@ class Job(object):
 
         os.setpgrp()
         signal.signal(signal.SIGHUP, signal.SIG_DFL)
-        signal.signal(signal.SIGINT, signal.SIG_DFL)
+        signal.signal(signal.SIGINT, signal.default_int_handler)
         signal.signal(signal.SIGTERM, signal.SIG_DFL)
 
         self._log_ctl.disable_logging()


lnst/Tests/Iperf.py
```diff
@@ -32,59 +32,47 @@ def run(self):
         server = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
                                   stderr=subprocess.PIPE, close_fds=True)
 
-        try:
-            stdout, stderr = server.communicate()
-            stdout = stdout.decode()
-            stderr = stderr.decode()
-        except KeyboardInterrupt:
```
olichtne, line 39: i think this `KeyboardInterrupt` should be kept for at least the server side (though i don't see why it couldn't be the same for client side as well) of iperf as there may be situations where there may be a usecase to do `server_job.kill(signal.SIGINT)` to gracefully exit the process.

yes, we don't do this in the `IperfFlowMeasurement`, but this is a Test module which may be used in different usecases.

lnst/Tests/Iperf.py
```diff
@@ -32,59 +32,47 @@ def run(self):
...
+        stdout, stderr = server.communicate()
+        stdout = stdout.decode().strip()
+        stderr = stderr.decode().strip()
+
+        if server.returncode > 0:
+            msg = f"iperf {self._role} returncode = {server.returncode}"
+            logging.error(msg)
+            logging.error(f"iperf stderr: {stderr}")
+            self._res_data["msg"] = msg
+            self._res_data["stderr"] = stderr
```
jtluka, line 44: I think we should log also the stdout. Similar to what is logged when `returncode == 0`.

lnst/Tests/Iperf.py
```diff
@@ -32,59 +32,47 @@ def run(self):
...
+            self._res_data["msg"] = msg
             self._res_data["stderr"] = stderr
-            logging.error(self._res_data["msg"])
             return False
 
-        try:
-            self._check_json_sanity()
-        except:
-            self._res_data["msg"] = "Iperf provided incomplete json data"
+        if not self._is_json_sane(self._res_data["data"]):
```
jtluka, line 58: Since we are renaming this - how about `_is_json_complete()`? The error message states "incomplete json data", so I think we could align the name with the message, wdyt?

jtluka: Also, the method takes `self._res_data["data"]` - why can't we change this call to:
```python
if not self._is_json_sane():
```
and make the method fetch it from `self._res_data["data"]`?

Otherwise it does not make sense to have this as a method since self is redundant.

lnst/Tests/Iperf.py
```diff
@@ -32,59 +32,47 @@ def run(self):
...
+            msg = "Iperf provided incomplete json data"
+            logging.error(msg)
+            logging.error(f"iperf stderr: {stderr}")
+            self._res_data["msg"] = msg
             self._res_data["data"] = stdout
```
jtluka, line 63: This was already in the code before this MR, but we shoud probably not overwrite something that is used for condition check.

When I checked the code further this statement is kind of redundant. The stdout must contain a valid json and is saved in `self._res_data["data"] = json.loads(stdout)` statement few lines above.

jtluka (changes requested):
Overall looks fine. Please address my comments. Also, do few test runs on the internal failing iperf test cases where we would benefit from the added logging.

lnst/Tests/Iperf.py
Kuba314: I don't think we should bind it to the internal implementation of the class. We're checking whether a json object returned by iperf is complete and that has nothing to do with the iperf arguments for example. I changed the method to be a `staticmethod` and changed the name to `_is_json_complete()`.

lnst/Tests/Iperf.py
Kuba314: Right, I removed the `self._res_data["data"] = stdout` line.

lnst/Tests/Iperf.py
Kuba314: I added `logging.debug()` of `stdout` into every error branch.

lnst/Tests/Iperf.py
Kuba314: I brought it back with `server.wait()` and put `.communicate()` into the `finally` block. There might be some potential buffering issues, but I don't understand it fully yet. I'll have to test this I guess. The alternative to this would be to copy the `communicate()` line into the `try` block as well (I think).

Kuba314: > `Popen.wait()` Note: This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

We should be using `.communicate()` definitely. I changed it so. Should be good now.

lnst/Tests/Iperf.py
jtluka: I'm ok with staticmethod, resolving.

jtluka (approved):
Looks good.

lnst/Tests/Iperf.py
```diff
@@ -34,57 +35,52 @@ def run(self):
 
         try:
             stdout, stderr = server.communicate()
-            stdout = stdout.decode()
-            stderr = stderr.decode()
         except KeyboardInterrupt:
-            pass
+            server.send_signal(signal.SIGINT)
```
olichtne, line 39: not sure this `send_signal` is required as when we call `Job.kill()` we send the signal to the whole process GROUP, not just the process itself, so both processes should get the signal from the start.

lnst/Tests/Iperf.py
```diff
@@ -34,57 +35,52 @@ def run(self):
...
+            stdout, stderr = server.communicate()
```
olichtne, line 40: should this be called for other exceptions/signals as well? in case some other issue happens it's probably still useful to try to retrieve the stdout/stderr values.

olichtne: at that point i guess it should just be a non-blocking attempt at retrieving the data...

olichtne (changes requested):
just some minor comments to clear up

Kuba314:
Currently I can't test this properly because of #354

lnst/Tests/Iperf.py
Kuba314: I'm currently unsure whether this `KeyboardInterrupt` exception can ever happen, because I can't seem to find a way to trigger it. Any ideas?

olichtne: this imo should trigger if you use a recipe like this:

```python
class MyRecipe(Recipe):
    def test(self):
        iperf_job = self.matched.m1.run(IperfServer(...), bg=True)
        sleep(5) # let it start
        iperf_job.kill(signal.SIGINT)
```

Kuba314: Doesn't seem to do anything:
```py
class MyRecipe(BaseRecipe):
    host1 = HostReq()
    host1.eth0 = DeviceReq(label="xd")

    def test(self) -> None:
        m1 = self.matched.host1
        iperf_job = m1.run(IperfServer(), bg=True)
        time.sleep(5)
        iperf_job.kill(signal.SIGINT)
        time.sleep(5)
```
```
2024-01-17 13:44:57           (host1)        -   DEBUG: Running job 0 with pid "1758" ({'job_id': 0, 'type': 'module', 'json': False, 'module': IperfServer()})
2024-01-17 13:44:57           (host1)        -   DEBUG: compiled command:  iperf3 -s  -J    
2024-01-17 13:44:57           (host1)        -   DEBUG: running as server ...
2024-01-17 13:45:02       (localhost)        -   DEBUG: Sending signal 2 to job 0 on host host1
2024-01-17 13:45:02           (host1)        -   DEBUG: Sending signal 2 to pid 1758
2024-01-17 13:45:07           (host1)        -    INFO: Killing all forked processes.
2024-01-17 13:45:07           (host1)        -   DEBUG: Killing all jobs
2024-01-17 13:45:07           (host1)        -   DEBUG: Sending signal 9 to pid 1758
2024-01-17 13:45:07       (localhost)        -    INFO: Result: FAIL, What: Job finished: type(module), host(host1), netns(None), IperfServer()
2024-01-17 13:45:07           (host1)        -    INFO: Restoring system configuration
```

olichtne: hmmm, confirmed on my side as well... i think this would need some hands on debugging with the agent process via a vm as i can't easily inspect the container version of the agent process.

Kuba314: Might have found the issue. When running a job, we reset `SIGINT` to `SIG_DFL` at https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Job.py#L99. I tested this manually and suddenly the process stopped raising a `KeyboardInterrupt`. Apprently there is a `signal.default_int_handler`, which should be set to `SIGINT` instead...

olichtne: ah, i was looking at that yesterday actually but `SIG_DFL` should reset to the default handler and from what i remember this used to work... so maybe this has changed at some point...

Kuba314: https://stackoverflow.com/a/22916853

olichtne (dismissed):
looks good now.

olichtne:
> ~Currently I can't test this properly because of #354~
> 
> EDIT: Planning to test this PR.

still planning to do some tests or is that also done now?

Kuba314:
@olichtne So... I tested this PR but I'm not sure what the behaviour should be exactly...

I SSHed into the client agent and tried sending `INT` via `kill`:
- INTing iperf process directly makes the `IperfClient` module `FAIL` and even if there's data, after all iterations are done, there's a `ZeroDivisionError` when trying to get `time_aligned_results`.
-  INTing the lnst-agent process which starts iperf does nothing (because `KeyboardInterrupt` is caught and there's a `pass` followed by another `.communicate()` call). I tried changing `pass` to `server.send_signal(signal.SIG_INT)`, which results in the same behaviour as the previous point -> `iperf` gets SIGINT, finishes, but for some reason we get `ZeroDivisionError`.
- INTing main `lnst-agent` process kills the agent (`Caught signal 2 -> dying` on agent, `host1 hard disconnected` on controller side)

Kuba314:
I added back the signal sending from the `KeyboardInterrupt` handler to forward it to the forked lnst-agent process and the iperf process. Should be good now. Ready to be merged.

olichtne (approved):

Issue #233, opened 2022-03-03
Timeout is evaluated as test pass


enhaut:
I hit this issue while i was working on #232.  
I used `HelloWorldRecipe` from our [wiki](https://lnst.readthedocs.io/en/latest/installation.html#creating-an-executable-helloworld-test-script). That is just simple ping test, machines should ping each other.   
The issue is, that we have in `HelloWorldRecipe` ping command without `-c` argument - without the limit, it runs endlessly.  
  
The second and main issue, is that endless ping is not blocking whole test, it is killed, but raising an timeout **is not** evaluated as a fail.

PR #350, opened 2023-11-21
Report job timeouts as FAIL

Kuba314:
### Description
Job timeouts should be reported as failures.

### Tests
`J:8577504`

### Reviews
@jtluka 

Closes: #233

### DIFF
--- a/container_files/Dockerfile
+++ b/container_files/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-FROM fedora:36
+FROM fedora:39
 
 RUN dnf install -y initscripts \
     iputils \
--- a/lnst/Agent/Agent.py
+++ b/lnst/Agent/Agent.py
@@ -11,6 +11,7 @@ __author__ = """
 jpirko@redhat.com (Jiri Pirko)
 """
 
+import importlib.machinery
 import signal
 import logging
 import os, stat
@@ -19,7 +20,6 @@ import datetime
 import socket
 import ctypes
 import multiprocessing
-import imp
 import types
 from time import sleep
 from inspect import isclass
@@ -161,7 +161,8 @@ class RemoteMethods:
         if module_name in self._dynamic_modules:
             return
         module_path = self._cache.get_path(res_hash)
-        module = imp.load_source(module_name, module_path)
+        module_loader = importlib.machinery.SourceFileLoader(module_name, module_path)
+        module = module_loader.load_module()
         self._dynamic_modules[module_name] = module
 
     def init_cls(self, cls_name, module_name, args, kwargs):
--- a/lnst/Agent/Job.py
+++ b/lnst/Agent/Job.py
@@ -131,13 +131,12 @@ class Job(object):
             os.killpg(self._pid, sig)
 
             if sig == signal.SIGKILL:
-                self.set_finished(dict(type = "job_finished",
-                                       job_id = self._id,
-                                       result = dict(passed = False,
-                                                     res_data = "Job killed",
-                                                     type = "result")))
-
-                send_data(self._child_pipe, self.get_result())
+                result = dict(type = "job_finished",
+                              job_id = self._id,
+                              result = dict(passed = False,
+                                            res_data = "Job killed",
+                                            type = "result"))
+                send_data(self._child_pipe, result)
             return True
         except OSError as exc:
             logging.error(str(exc))
--- a/lnst/Recipes/ExampleRecipe.py
+++ b/lnst/Recipes/ExampleRecipe.py
@@ -39,21 +39,33 @@ class MyRecipe(BaseRecipe):
         self.matched.m1.eth0.up()
         self.matched.m2.eth0.ip_add(ipaddress("192.168.1.2/24"))
         self.matched.m2.eth0.up()
-        ping_job = self.matched.m1.run(Ping(dst=self.matched.m2.eth0,
-                                                interval=0,
-                                                iface=self.matched.m1.eth0))
 
-        netserver_job = self.matched.m1.run(Netserver(bind=self.matched.m1.eth0),
-                                            bg=True)
+        self.matched.m1.run(
+            Ping(
+                dst="192.168.1.2/24",
+                interface=self.matched.m1.eth0,
+            )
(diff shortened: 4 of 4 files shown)


lnst/Controller/Job.py
```diff
@@ -166,6 +166,7 @@ def start(self, bg=False, timeout=DEFAULT_TIMEOUT):
             if not self.wait(timeout):
                 logging.debug("Killing timed-out job")
                 self.kill()
+                self._netns._machine._add_recipe_result(JobFinishResult(self, timed_out=True))
```
jtluka, line 169: I think this should go into Machine, where you can define `job_timeouted()` method, similar to [`job_finished()`](https://github.com/LNST-project/lnst/blob/cac4127d460983d980c98a4d96779462f881351f/lnst/Controller/Machine.py#L542) that would get called from [`wait_for_job()`](https://github.com/LNST-project/lnst/blob/cac4127d460983d980c98a4d96779462f881351f/lnst/Controller/Machine.py#L512) if the timeout is hit.

Additionally I'm wondering if we should raise an Exception if this happens. When a command timeouts it may indicate a potential issue and recipe should not proceed further.

olichtne: there is more to this though... when we call `job.kill(SIGKILL)` this ends up being called on the agent side: https://github.com/LNST-project/lnst/blob/cac4127d460983d980c98a4d96779462f881351f/lnst/Agent/Job.py#L134C62-L134C62

which should result in a job finished message to the controller marking the job as killed.

which will get to create a `JobFinishedResult` in https://github.com/LNST-project/lnst/blob/cac4127d460983d980c98a4d96779462f881351f/lnst/Controller/Machine.py#L542 so introducing this timedout result here in this PR will actually create two results for one job which i don't think is ok.

i guess we should instead make sure that when this JobFinished is a "killed" it should be marked as a FAIL instead of adding this timed out attribute.

Kuba314: If we want to be raising exceptions instead, I think we should add an argument to this method maybe? It would be something like `raise_on_timeout=True` (and would work only when `bg=True`).

We then don't have to add this `job_timeouted()` stuff at all (theoretically), but it might be nice to have.

olichtne: hang on i think there's a completely different bug here...

https://github.com/LNST-project/lnst/blob/cac4127d460983d980c98a4d96779462f881351f/lnst/Agent/Job.py#L136C20-L136C20 already sets `passed=False` but we're not seeing it...

and looking at the sequence we do:
```python
set_finished()
send_data(self._child_pipe)
```

but `set_finished()` also calls `self._child_pipe.close()`

i think this is an issue of us closing the connection pipe before sending result data back and that's why we're maybe not seeing this correctly?

olichtne: it looks like my suspicion is correct:
running with:
```python
for run in r.runs:
     logging.info(summary_fmt.format_run(run))

     for i in run.results:
         print(i, i.description)
```
we get this:
```
<lnst.Controller.RecipeResults.DeviceMethodCallResult object at 0x7f63eaee3790> Calling Device method m1.eth0.ip_add(Ip4Address(192.168.1.1/24))
<lnst.Controller.RecipeResults.DeviceMethodCallResult object at 0x7f63eae64610> Calling Device method m1.eth0.up()
<lnst.Controller.RecipeResults.DeviceMethodCallResult object at 0x7f63eaee1e50> Calling Device method m2.eth0.ip_add(Ip4Address(192.168.1.2/24))
<lnst.Controller.RecipeResults.DeviceMethodCallResult object at 0x7f63eb904dd0> Calling Device method m2.eth0.up()
<lnst.Controller.RecipeResults.JobStartResult object at 0x7f63eaee2990> Job started: type(module), host(m1), netns(None), Ping(dst='192.168.1.2/24', count=1000000, interface=Device(machine=m1, id=eth0, name=eth1, ifindex=3))
<lnst.Controller.RecipeResults.JobStartResult object at 0x7f63eaee4950> Job started: type(module), host(m1), netns(None), Netserver(bind='192.168.1.1/24')
<lnst.Controller.RecipeResults.JobFinishResult object at 0x7f63eae9ae90> Job finished: type(module), host(m1), netns(None), Netserver(bind='192.168.1.1/24')
<lnst.Controller.RecipeResults.JobStartResult object at 0x7f63eb905d10> Job started: type(module), host(m2), netns(None), Netperf(server='192.168.1.1/24', duration=1, confidence='99,5', runs='5', debug=0, max_deviation={'type': 'percent','value': 20.0}, testname='TCP_STREAM')
<lnst.Controller.RecipeResults.JobFinishResult object at 0x7f63eaeac450> Job finished: type(module), host(m2), netns(None), Netperf(server='192.168.1.1/24', duration=1, confidence='99,5', runs='5', debug=0, max_deviation={'type': 'percent', 'value': 20.0}, testname='TCP_STREAM')
```

and for every `JobStartedResult` result, here should be a `JobFinishedResult` however, we can see that the Ping finished result is missing.

olichtne: yeah, so this is a much more complicated issue that it seems at first, i'm looking into it myself right now but if you want to try @Kuba314 we can discuss more details tomorrow morning... but this will not be a simple issue to track down even if in the end it may result in a simple fix... not sure yet.

olichtne: nevermind... i managed to actually find the "simple fix", slightly by accident...

I pushed the additional Agent side commit to this MR.

Please do double check and test this, and if everything works we can probably rebase this PR and use this fix instead of the `JobFinished(timedout=true)`

Kuba314:
```
Result: FAIL, What: Job finished: type(module), host(host1), netns(None), Ping(dst=Ip4Address(192.168.101.2/24), interface=Ip4Address(192.168.101.1/24), count=1000000, interval=0.2, size=56)
```

The failed result appeared. I'm going to drop the first commit.

Kuba314: Done.

olichtne (approved):

jtluka (approved):
Looks good.

enhaut (approved):

Issue #354, opened 2023-12-13
lnst-agent crashes on controller KeyboardInterrupt during Iperf


Kuba314:
Running a standard `SimpleNetworkRecipe`, waiting until `IperfClient` and `IperfServer` are both running, then sending `SIGINT` to controller (or just hitting `^C`) crashes both agents with the following exception:
```
Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/lnst-MdZhQDa5-py3.11/bin/lnst-agent", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/root/lnst/lnst/Agent/__main__.py", line 45, in main
    agent.run()
  File "/root/lnst/lnst/Agent/Agent.py", line 1000, in run
    self._process_msg(msg[1])
  File "/root/lnst/lnst/Agent/Agent.py", line 1065, in _process_msg
    job.join()
    ^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'join'
```

jtluka:
I think this could be solved by moving the `join()` and `set_finished()` calls to `JobContext` class, and imo it makes sense have all jobs related tasks within the JobContext. The Agent'would only forward the message to Controller.

jtluka:
> I think this could be solved by moving the `join()` and `set_finished()` calls to `JobContext` class, and imo it makes sense have all jobs related tasks within the JobContext. The Agent'would only forward the message to Controller.

That would mean that we'll have to call `job.join()` explicitly also in `kill_job()` and probably in other places, too. Not sure.

Kuba314:
So any time someone would call `job.kill()`, they'd also have to call `job.join()` and `job.set_finished()`?

What if we just didn't clear the job dict in `JobContext.cleanup()` and maybe instead introduced `JobContext.pop_job()`, which would be something similar to `JobContext.get_job()`, but would also delete the job from the dict? We could then call this method in the `"job_finished"` handler instead of `JobContext.get_job()`.

WDYT? @jtluka @olichtne

olichtne:
i'll have to try to reproduce this to inspect the state before i can be sure on how to solve this. i'll try to do that this week.

PR #355, opened 2024-01-02
Fix job cleanup on agents

Kuba314:
### Description
See commit message bodies for more info.

### Tests
Normal SimpleNetworkRecipe execution passes, interrupting with ^C during iperf test also finishes well and doesn't crash agents. Not sure if this is enough testing though...

Closes: #354

### DIFF
--- a/lnst/Agent/Agent.py
+++ b/lnst/Agent/Agent.py
@@ -404,12 +404,12 @@ class RemoteMethods:
 
     def kill_jobs(self):
         logging.info("Killing all forked processes.")
-        self._job_context.cleanup()
+        self._job_context.kill_all_jobs()
         return "Commands killed"
 
     def machine_cleanup(self):
         logging.info("Performing machine cleanup.")
-        self._job_context.cleanup()
+        self._job_context.kill_all_jobs()
 
         self.restore_system_config()
 
--- a/lnst/Agent/Agent.py
+++ b/lnst/Agent/Agent.py
@@ -1061,13 +1061,11 @@ class Agent:
             self._job_context.del_cmd(job)
             self._server_handler.send_data_to_ctl(msg)
         elif msg["type"] == "job_finished":
-            job = self._job_context.get_job(msg["job_id"])
+            job = self._job_context.pop_job(msg["job_id"])
             job.join()
 
             job.set_finished(msg["result"])
             self._server_handler.send_data_to_ctl(msg)
-
-            self._job_context.del_job(job)
 
         elif msg["type"] == "from_netns":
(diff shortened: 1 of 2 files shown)


olichtne (approved):
looks ok to me, did you check if the `cleanup` method was used anywhere else ? this is a public RPC API method so it could be getting called from the Controller as well, haven't checked myself...

Kuba314:
@olichtne The `cleanup()` method was on `JobContext`, which is only *used* in `RemoteMethods` and cannot actually be called via RPC, or have I misunderstood?

Btw, https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Agent.py#L1059 `get_cmd` and `del_cmd` don't actually exist anymore, so the whole `"exception"` handler can be removed? Which is weird because we are probably sometimes sending `"exception"` messages at:
- https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Agent.py#L1036
- https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Agent.py#L1046
- https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Agent.py#L1081

olichtne:
> @olichtne The cleanup() method was on JobContext, which is only used in RemoteMethods and cannot actually be called via RPC, or have I misunderstood?

ah, you're right, i misread that...

---

> Btw, https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Agent.py#L1059 get_cmd and del_cmd don't actually exist anymore, so the whole "exception" handler can be removed? Which is weird because we are probably sometimes sending "exception" messages at:

so there are 2 exception type messages:
1. what you've linked is exception messages that are provided to `send_data_to_ctl` method, this sends a message from an Agent process directly to the Controller process
2. there are exception messages from Jobs: https://github.com/LNST-project/lnst/blob/master/lnst/Agent/Job.py#L114 this is sent from a **Forked Agent process running a Job._run method** to the main Agent process... there it should be handled by the exception message handler you linked.

So I think this should be kept as this has a different, still existing use case.

However, I'm not sure why the `get_cmd` and `del_cmd` method calls are still there and why we haven't seen any problems with them recently.

This should be testable by runing a test module linke this:

```
class MyTestModule(BaseTestModule):
    def run():
        sleep(2) # simulate runtime
        raise Exception("exception from Job process")
```

olichtne:
got it. makes sense... i'd still keep these except branch blocks and the exception message handling in the Agent for now. It may look useless right now but i'm not 100% sure and i think it at least doesn't have to be part of this PR.

@Kuba314 anything else that you want to do for this PR since you're keeping it in Draft status? or can we proceed with merging this in?

PR #358, opened 2024-01-23
PerfMeasurements: add ability to simulate measurements

olichtne:
### Description
The "bottom" measurements - meaning the most core measurement in a list of measurements that we may run can now be "simulated".

This means that instead of actually running the measurement tools/commands we instead run the following command:

    echo simulated start

And instead of returning real measurement results we return the same structure of data with a single 0 value perf interval.

Combining this we can now run all of our ENRT recipes in a "simulation" that preserves the number of commands per test recipe run and should return results compatible with all of our tooling with just the basic 0 value.

The cpu measurement simulation does actually call the real measurement, this is because this isn't a "bottom/core" measurement and is instead designed to always wrap some other measurement.

### Tests
I want to use this to generate baseline to threshold mapping for all of the tests and machines that we currently use so i'll link those jobs to here once i have confirmed everything works.

### Reviews

Closes: #

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/BaseFlowMeasurement.py
@@ -1,3 +1,4 @@
+import time
 from dataclasses import dataclass
 import textwrap
 from typing import Optional, Union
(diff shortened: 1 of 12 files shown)


olichtne:
Testing successful in beaker job: `J:8848773`

olichtne:
`J:8917772` new test for the rebased MR

PR #359, opened 2024-02-02
Recipes/SRIOVNetnsOvSRecipe.py: deconfigure ovs bridge before resetti…

jtluka:
…ng switchdev mode

### Description

On a setup with Nvidia CX6DX NIC we hit an issue when switchdev mode of the NIC could not be set back to legacy mode.

After some testing, the issue was resolved after the switchdev mode was reset after the ovs bridge device was brought down and ports removed from the switch. Previously this was done after reset of switchdev.

Logically, the steps of deconfiguration should match the configuration steps, in reversed order.

### Tests

Already tested in J:8877925

### Reviews

@Axonis

### DIFF
--- a/lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
+++ b/lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
@@ -160,7 +160,14 @@ class SRIOVNetnsOvSRecipe(
         Finally virtual function is deleted.
         """
         host1, host2 = self.matched.host1, self.matched.host2
-        for i, host in enumerate([host1, host2]):
+
+        for host in [host1, host2]:
+            for dev in [host.eth0, host.vf_representor_eth0, host.br0]:
+                dev.down()
+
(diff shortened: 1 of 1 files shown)


lnst/Recipes/ENRT/SRIOVNetnsOvSRecipe.py
```diff
@@ -160,6 +160,14 @@ def test_wide_deconfiguration(self, config):
...
         """
         host1, host2 = self.matched.host1, self.matched.host2
+
+        for host in [host1, host2]:
+            for dev in [host.eth0, host.vf_representor_eth0, host.br0]:
+                dev.down()
+
+            host.br0.port_del(host.eth0)
+            host.br0.port_del(host.vf_representor_eth0)
+
```
Axonis, line 170: Refactor to single loop, otherwise looks good.

Axonis (dismissed):

Axonis (approved):

PR #363, opened 2024-02-21
External.TRex.UDPSimple: use src/dst mac and ip addresses

olichtne:
### Description
The default trex packet stream generator didn't correctly use the provided source and destination ip and mac addresses.

This wasn't noticed until now because the test is typically run in a lab where multiple machine pairs were isolated from eachother.

This however isn't MANDATORY and so to avoid conflicts on a shared network we should respect the provided values.

### Tests

### Reviews

Closes: #

### DIFF
--- a/lnst/External/TRex/UDPSimple.py
+++ b/lnst/External/TRex/UDPSimple.py
@@ -23,8 +23,8 @@ class UDPSimple(object):
 
     def create_stream (self, **kwargs):
         # Use port's configured mac and ip addresses
-        L2 = Ether()
-        L3 = IP()
+        L2 = Ether(src=kwargs["src_mac"], dst=kwargs["dst_mac"])
+        L3 = IP(src=kwargs["src_ip"], dst=kwargs["dst_ip"])
         L4 = UDP()
 
         size = kwargs.get("msg_size", 64)


lnst/External/TRex/UDPSimple.py
```diff
@@ -23,8 +23,8 @@ def _import_optionals():
 
     def create_stream (self, **kwargs):
         # Use port's configured mac and ip addresses
-        L2 = Ether()
-        L3 = IP()
+        L2 = Ether(src=kwargs["src_mac"], dst=kwargs["dst_mac"])
+        L3 = IP(src=kwargs["src_ip"], dst=kwargs["dst_ip"])
```
jtluka, line 27: Are these always present in kwargs?

jtluka (approved):
Besides the only comment, looks ok.

lnst/External/TRex/UDPSimple.py
olichtne: yes, this is where these get generated: https://github.com/LNST-project/lnst/blob/master/lnst/External/TRex/TRexLib.py#L96C28-L96C44

and these in turn come from the mandatory flows parameter here: https://github.com/LNST-project/lnst/blob/master/lnst/Tests/TRex.py#L13

jtluka: Ok, thanks for clarifying this.

PR #364, opened 2024-03-09
Recipes/ENRT/ConfigMixins: Add FirewallMixin.py

SirPhuttel:
The set of classes contained in this module enable users to push
firewall rulesets onto the hosts used for test runs.

There are five flavors to choose from for the actual firewall
implementation, with descriptive names:

* NftablesMixin
* IptablesMixin
* Ip6tablesMixin
* EbtablesMixin
* ArptablesMixin

To use the provided functionality, inherit one of them and either
provide a dictionary of "host: ruleset" entries in a property
'firewall_rulesets' or overwrite the 'firewall_rulesets_generator()'
method with a generator yielding as many dictionaries a desired for
performing a test run with. The latter provides a simple means of
testing with gradually increasing or extending rulesets.

In order to perform post-processing on rulesets after a test run,
override the 'firewall_rulesets' setter method.

Changes since v1:
- Make use of ABC and abstractmethod
- Use more modern f"..." format strings
- Explicitly delete stored subconfig variables to free memory
- Use 'rulesets' where a hash of {host: ruleset} is assigned
- Keep stored rulesets in config object, too
- Merge BulkFirewallMixin into FirewallMixin
- Rename Ip46tablesMixin to IptablesBaseMixin
- Fix for buggy _extract_tables_n_chains() implementation, also
  skip empty ruleset lines
- Dump and restore rule counter values along with iptables rulesets
- Implement FirewallControl, a class to instantiate on agents
- Eliminate the use of temporary files
- Update 'firewall_rulesets' attribute after each test

### Tests
I wrote a few hello world recipes for testing, they're here:
https://github.com/SirPhuttel/lnst/commit/ea21b0e6e3da95a66bdaba98623bc1ddada2d2b1

### Bugs
Overriding a parent's class's property is not easy if a setter implementation
is also present. For instance, when inheriting NftablesMixin and only
overriding the getter, it has to be decorated using
'@NftablesMixin.firewall_rulesets.getter' instead of just '@property'. When
overriding both getter and setter though, the expected '@property' and
'@firewall_rulesets.setter' suffice.

### Reviews
@jtluka
@enhaut
@olichtne
(Thanks in advance!)

### DIFF
--- a/lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
@@ -1,0 +1,156 @@
+from abc import ABC, abstractmethod
+from lnst.Recipes.ENRT.ConfigMixins import BaseSubConfigMixin
+from lnst.RecipeCommon.FirewallControl import FirewallControl
+import copy
+
+class FirewallMixin(BaseSubConfigMixin, ABC):
+    """
+    A config mixin to apply custom firewall rulesets on hosts.
+    Do not inherit directly, use one of the derived classes below instead.
+    """
+
+    _fwctl = {}
+
+    def fwctl(self, host):
+        try:
+            return self._fwctl[host]
+        except KeyError:
+            self._fwctl[host] = host.init_class(FirewallControl)
+            return self._fwctl[host]
+
+    @property
+    def firewall_rulesets(self):
+        """
+        This property holds a dictionary of firewall rulesets in textual
+        representation, indexed by the host it should be applied to. A typical
+        use is:
+
+        { self.matched.host1: <host1 ruleset>,
+          self.matched.host2: <host2 ruleset> }
+
+        This property is used by the default `firewall_rules_generator()`
+        method. Overwriting the latter from a recipe constitutes an alternative
+        to implementing it.
+
+        The rulesets will be applied by a derived class's _apply_ruleset()
+        method, i.e. typically fed into 'nft -f' or 'iptables-restore'.
+        """
+        return {}
+
+    @firewall_rulesets.setter
+    def firewall_rulesets(self, rulesets):
+        """
+        This setter is called with all hosts' rulesets after each test run.
+        Overwrite it to perform post processing or analysis on contained state.
+        """
+        pass
+
+    @property
+    def firewall_rulesets_generator(self):
+        """
+        A generator yielding { host: ruleset, ... } to apply for a test run.
+        Each yield will turn into a new sub configuration and thus be tested
+        separately.
+        """
+        return [self.firewall_rulesets]
+
+    def generate_sub_configurations(self, config):
+        for parent_config in super().generate_sub_configurations(config):
+            for rulesets in self.firewall_rulesets_generator:
+                new_config = copy.copy(parent_config)
+                new_config.firewall_rulesets = rulesets
+                yield new_config
+
+    @abstractmethod
+    def _apply_ruleset(self, host, ruleset):
+        ...
+
+    def apply_sub_configuration(self, config):
+        super().apply_sub_configuration(config)
+
+        stored = {}
+        for host, ruleset in config.firewall_rulesets.items():
+            stored[host] = self._apply_ruleset(host, ruleset)
+        config.stored_firewall_rulesets = stored
+
+    def generate_sub_configuration_description(self, config):
+        desc = super().generate_sub_configuration_description(config)
+
+        for host, ruleset in config.firewall_rulesets.items():
+            nlines = len(ruleset.split("\n"))
+            desc.append(f"Firewall: ruleset with {nlines} lines on host {host}")
+
+        return desc
+
+    def remove_sub_configuration(self, config):
+        applied = {}
+        for host, ruleset in config.stored_firewall_rulesets.items():
+            applied[host] = self._apply_ruleset(host, ruleset)
+        self.firewall_rulesets = applied
+
+        del config.stored_firewall_rulesets
+        del config.firewall_rulesets
+
+        return super().remove_sub_configuration(config)
+
+class NftablesMixin(FirewallMixin):
+    """
+    An nftables backend for FirewallMixin.
+    """
+
+    def _apply_ruleset(self, host, ruleset):
+        old = self.fwctl(host).apply_nftables_ruleset(ruleset.encode('utf-8'))
+        return old.decode('utf-8')
+
+class IptablesBaseMixin(FirewallMixin):
+    """
+    A common base class for all the iptables-like FirewallMixin backends.
+    Do not inherit directly, use one of the *tablesMixin classes instead.
+    """
+
+    @property
+    @abstractmethod
+    def iptables_command(self):
+        ...
+
+    def _apply_ruleset(self, host, ruleset):
+        ruleset = ruleset.encode('utf-8')
+        cmd = self.iptables_command.encode('utf-8')
+        old = self.fwctl(host).apply_iptableslike_ruleset(cmd, ruleset)
+        return old.decode('utf-8')
+
+class IptablesMixin(IptablesBaseMixin):
+    """
+    An iptables backend for FirewallMixin.
+    """
+
+    @property
+    def iptables_command(self):
+        return "iptables"
+
+class Ip6tablesMixin(IptablesBaseMixin):
+    """
+    An ip6tables backend for FirewallMixin.
+    """
+
+    @property
+    def iptables_command(self):
+        return "ip6tables"
+
+class EbtablesMixin(IptablesBaseMixin):
+    """
+    An ebtables backend for FirewallMixin.
+    """
+
+    @property
+    def iptables_command(self):
+        return "ebtables"
+
+class ArptablesMixin(IptablesBaseMixin):
+    """
+    An arptables backend for FirewallMixin.
+    """
+
+    @property
+    def iptables_command(self):
+        return "arptables"
--- a/lnst/Common/ExecCmd.py
+++ b/lnst/Common/ExecCmd.py
@@ -56,12 +56,13 @@ def log_output(log_func, out_type, out):
              "----------------------------"
              % (out_type, out))
 
-def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=False):
+def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=False, stdin=None):
     cmd = cmd.rstrip(" ")
     logging.debug("Executing: \"%s\"" % cmd)
     subp = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE, close_fds=True)
-    (data_stdout, data_stderr) = subp.communicate()
+                            stderr=subprocess.PIPE, stdin=subprocess.PIPE,
+                            close_fds=True)
+    (data_stdout, data_stderr) = subp.communicate(input = stdin)
     data_stdout = data_stdout.decode()
     data_stderr = data_stderr.decode()
 
--- a/lnst/RecipeCommon/FirewallControl.py
+++ b/lnst/RecipeCommon/FirewallControl.py
@@ -1,0 +1,46 @@
+from lnst.Common.ExecCmd import exec_cmd
+
+class FirewallControl(object):
+    def _extract_tables_n_chains(self, ruleset):
+        out = {}
+        curtable = None
+        for line in ruleset.split('\n'):
+            if len(line.strip()) == 0:
+                continue
+            if line[0] == '*':
+                curtable = line[1:]
+                out[curtable] = []
+            elif line[0] == ':':
+                words = line.split(' ')
+                if not words[1] == '-': # ignore user-defined chains
+                    out[curtable].append(words[0][1:])
+        return out
+
+    def _append_missing_parts(self, dst_ruleset, src_ruleset):
+        for table, chains in self._extract_tables_n_chains(src_ruleset).items():
+            tline = f'*{table}\n'
+            if dst_ruleset.find(tline) >= 0:
+                continue
+            dst_ruleset += tline
+            for chain in chains:
+                dst_ruleset += f':{chain} ACCEPT [0:0]\n'
+            dst_ruleset += 'COMMIT\n'
+        return dst_ruleset
+
+    def apply_nftables_ruleset(self, ruleset):
+        ruleset = f"flush ruleset\n{ruleset.decode('utf-8')}"
+        old, err = exec_cmd("nft list ruleset",
+                            report_stderr=True, log_outputs=False)
+        exec_cmd("nft -f -", report_stderr=True,
+                 stdin=ruleset.encode('utf-8'))
+        return old.encode('utf-8')
+
+    def apply_iptableslike_ruleset(self, cmd, ruleset):
+        ruleset = ruleset.decode('utf-8')
+        cmd = cmd.decode('utf-8')
+        old, err = exec_cmd(f"{cmd}-save --counters",
+                            report_stderr=True, log_outputs=False)
+        ruleset = self._append_missing_parts(ruleset, old)
+        exec_cmd(f"{cmd}-restore --counters", report_stderr=True,
+                 stdin=ruleset.encode('utf-8'))
+        return old.encode('utf-8')


lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+
+    @property
+    def bulk_firewall_rules_generator(self):
+        """
+        A generator yielding { host: ruleset, ... } to apply for a test run.
+        Each yield will turn into a new sub configuration and thus be tested
+        separately.
+        """
+        raise NotImplementedError("Method bulk_firewall_rules_generator must be defined by a child class.")
```
enhaut, line 91: Same as above

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        { self.matched.host1: <host1 ruleset>,
+          self.matched.host2: <host2 ruleset> }
+
+        The rulesets will be applied by a derived class's _apply_ruleset()
+        method, i.e. typically fed into 'nft -f' or 'iptables-restore'.
+        """
+        return {}
+
+    def _apply_ruleset(self, host, path):
+        raise NotImplementedError("Method _apply_ruleset must be defined by a child class.")
```
enhaut, line 29: What about marking this as [`abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod)? Imo,  it's safer as it doesn't allow you to even instantiate the class. Also, `abstractmethods` would make it easier for static analysis tools.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def _dump_ruleset(self, host, config):
+        raise NotImplementedError("Method _dump_ruleset must be defined by a child class.")
```
enhaut, line 32: Same as above

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+
+class Ip46tablesMixin(FirewallMixin):
+    """
+    A common base class for all the iptables-like FirewallMixin backends.
+    Do not inherit directly, use one of the *tablesMixin classes instead.
+    """
+
+    @property
+    def iptables_command(self):
+        raise NotImplementedError("Property iptables_command must be defined by a child class.")
```
enhaut, line 119: Same as above

enhaut (commented):
Just a minor comments from my side, overall looks good to me.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def remove_sub_configuration(self, config):
+        """
+        override parent's implementation:
+        If the stored ruleset does not contain all tables of the current one,
+        rules remain in place.
+        """
+        for host, ruleset in config.firewall_rulesets.items():
+            stored = self.stored_rulesets[host]
+            self.stored_rulesets[host] = self.append_missing_parts(stored, ruleset)
```
jtluka, line 160: You're missing underscore in `self.append_missing_parts`

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        if not config.firewall_rulesets:
+            config.firewall_rulesets = self.firewall_rulesets
+
+        self.stored_rulesets = {}
+        for host, ruleset in config.firewall_rulesets.items():
+            self.stored_rulesets[host] = self._dump_ruleset(host, config)
+
+            name = self._push_ruleset(host, ruleset)
+            self._apply_ruleset(host, name)
+            host.run('rm -f "%s"' % name)
```
jtluka, line 57: Maybe use more convenient way of f-strings:
```python
host.run(f"rm -f \"{name}\"")
```

The expression in curly brackets can be also a function call.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        return desc
+
+    def remove_sub_configuration(self, config):
+        for host, ruleset in self.stored_rulesets.items():
+            # XXX: this may spam debug log, but counter values may be worth logging
+            logging.debug("Dropping Firewall ruleset on host %s:\n%s\n" % (host, self._dump_ruleset(host, config)))
+            name = self._push_ruleset(host, ruleset)
+            self._apply_ruleset(host, name)
+            host.run('rm -f "%s"' % name)
+
```
jtluka, line 74: You problably want to delete both `self.stored_rulesets` and `config.firewall_rulesets`:

```python
del self.stored_rulesets
del config.firewall_rulesets
```

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+    def bulk_firewall_rules_generator(self):
+        """
+        A generator yielding { host: ruleset, ... } to apply for a test run.
+        Each yield will turn into a new sub configuration and thus be tested
+        separately.
+        """
+        raise NotImplementedError("Method bulk_firewall_rules_generator must be defined by a child class.")
+
+    def generate_sub_configurations(self, config):
+        for ruleset in self.bulk_firewall_rules_generator:
```
jtluka, line 94: Could you use plural for "ruleset" since we're assigning it later to `new_config.firewall_rulesets = ruleset`?

Minor thing, but I had to stop for a while why it's just ONE ruleset, so probably worth to change to make it more readable.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        raise NotImplementedError("Method _dump_ruleset must be defined by a child class.")
+
+    def _push_ruleset(self, host, ruleset):
+        """
+        Push `ruleset` into a temporary file on `host`, returning its path.
+        """
+        fd, name = mkstemp(prefix="FirewallMixin", text=True)
+        write(fd, ruleset.encode('utf-8'))
+        close(fd)
+        remote_path = host.copy_file_to_machine(name)
```
jtluka, line 41: We should probably remove these files in `remove_sub_configuration`. No need to fix this now, since there's no convenient method to do this through Host/Namespace API, but we should create a github issue to deal with that later.

jtluka (commented):
Besides some minor issues, I think this looks ok.

olichtne:
> Is there a test suite I could integrate any unit tests for the config mixin?

not really... this lnst repository doesn't really have "self tests" we run all of our recipes regularly and typically also use those to schedule a test run in beaker with the modified code from the PR to provide a proof of testing.

> The applied ruleset should be dumped

what exactly do you mean by dumped?

do you want to print the information for "debug purposes" or just because it could be "interesting"? or is this information IMPORTANT in some way and should be relevant for the PASS/FAIL of a recipe?

This also provides context for what to do with

> I wrote a few hello world recipes for testing, they're here: https://github.com/SirPhuttel/lnst/commit/630fb76ceab555db6418efa16e04755c6b39ac6d

as if these are eventually meant to do some form of MEASUREMENT which we want to regularly evaluate... we'd want to look into integrating these tests into the `lnst.Recipes.ENRT` test set

If these are "significantly" different, we probably want to create a new test set .... `lnst.Recipes.Firewalls` for example...

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        raise NotImplementedError("Method _apply_ruleset must be defined by a child class.")
+
+    def _dump_ruleset(self, host, config):
+        raise NotImplementedError("Method _dump_ruleset must be defined by a child class.")
+
+    def _push_ruleset(self, host, ruleset):
+        """
+        Push `ruleset` into a temporary file on `host`, returning its path.
+        """
+        fd, name = mkstemp(prefix="FirewallMixin", text=True)
```
olichtne, line 38: why use the lower level `mkstemp` function and manual call of `close` ? wouldn't the context manager way make this a bit simpler?

```python
with NamedTemporaryFile(prefix="FirewallMixin") as fd:
    write(fd, ruleset.encode())
    close(fd)
    remote_path = host.copy_file_to_machine(fd.name)
```

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        unlink(name)
+        return remote_path
+
+    def apply_sub_configuration(self, config):
+        super().apply_sub_configuration(config)
+
+        if not config.firewall_rulesets:
+            config.firewall_rulesets = self.firewall_rulesets
+
+        self.stored_rulesets = {}
```
olichtne, line 51: `stored_rulesets` should also probably be part of the `config` object

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
olichtne: this is what removes the pushed files - i think this answers your other comment in https://github.com/LNST-project/lnst/pull/364/files#r1519638084

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        for host, ruleset in self.stored_rulesets.items():
+            # XXX: this may spam debug log, but counter values may be worth logging
+            logging.debug("Dropping Firewall ruleset on host %s:\n%s\n" % (host, self._dump_ruleset(host, config)))
+            name = self._push_ruleset(host, ruleset)
+            self._apply_ruleset(host, name)
+            host.run('rm -f "%s"' % name)
+
+        return super().remove_sub_configuration(config)
+
+class BulkFirewallMixin(FirewallMixin):
```
olichtne, line 77: i feel like this class - the generator method and the subconfig generator should be part of the base `FirewallMixin` class... i mean, without this, the subclasses don't really do anything on their own, and with this addition, they would still by default do the default "nothing" operation.

But I'm not sure if you had some more specific reason to separate this into its own class?

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+    A common base class for all the iptables-like FirewallMixin backends.
+    Do not inherit directly, use one of the *tablesMixin classes instead.
+    """
+
+    @property
+    def iptables_command(self):
+        raise NotImplementedError("Property iptables_command must be defined by a child class.")
+
+    def _apply_ruleset(self, host, path):
+        host.run('%s-restore "%s"' % (self.iptables_command, path))
```
olichtne, line 122: this would look nicer as an f-string:
```
host.run(f'{self.iptables_command}-restore "{path}"')
```

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+
+    def _apply_ruleset(self, host, path):
+        host.run('nft flush ruleset')
+        host.run('nft -f "%s"' % path)
+
+    def _dump_ruleset(self, host, config):
+        return host.run('nft list ruleset').stdout
+
+class Ip46tablesMixin(FirewallMixin):
```
olichtne, line 111: would it make more sense to name this as `IptablesBaseMixin` instead? seeing that there's also `ebtables` and `arptables` that inherit from this?

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+    def _extract_tables_n_chains(self, ruleset):
+        out = {}
+        curtable = None
+        for line in ruleset.split('\n'):
+            if line[0] == '*':
+                out[line] = []
+            elif line[0] == ':':
+                words = line.split(' ')
+                if not words[1] == '-': # ignore user-defined chains
+                    out[curtable].append(words[0][1:])
```
olichtne, line 133: `curtable` is always `None` here is that correct?

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,197 @@
...
+        return out
+
+    def _append_missing_parts(self, dst_ruleset, src_ruleset):
+        for table, chains in self._extract_tables_n_chains(src_ruleset).items():
+            tline = '*' + table + '\n'
```
olichtne, line 138: the `table` here is coming from https://github.com/LNST-project/lnst/pull/364/files#diff-c1ca3132235b422fab430b7645942020bf1635e314f37db433a3ca03575f4e2dR129 which i think means that it fits the format of `*....` with the `\n` cut off

but here you assign `tline = '*' + "*...."` so you'll end up with double `*` characters is that correct?

olichtne (changes requested):

SirPhuttel:
On Tue, Mar 12, 2024 at 04:49:25AM -0700, Ondrej Lichtner wrote:
> @olichtne requested changes on this pull request.
> 
> 
> 
> > +        The rulesets will be applied by a derived class's _apply_ruleset()
> +        method, i.e. typically fed into 'nft -f' or 'iptables-restore'.
> +        """
> +        return {}
> +
> +    def _apply_ruleset(self, host, path):
> +        raise NotImplementedError("Method _apply_ruleset must be defined by a child class.")
> +
> +    def _dump_ruleset(self, host, config):
> +        raise NotImplementedError("Method _dump_ruleset must be defined by a child class.")
> +
> +    def _push_ruleset(self, host, ruleset):
> +        """
> +        Push `ruleset` into a temporary file on `host`, returning its path.
> +        """
> +        fd, name = mkstemp(prefix="FirewallMixin", text=True)
> 
> why use the lower level `mkstemp` function and manual call of `close` ? wouldn't the context manager way make this a bit simpler?
> 
> ```python
> with NamedTemporaryFile(prefix="FirewallMixin") as fd:
>     write(fd, ruleset.encode())
>     close(fd)
>     remote_path = host.copy_file_to_machine(fd.name)
> ```

You meant manual call of unlink(), right? Seems sensible, though I don't
see the benefit (but I see everything through procedural goggles
anyway).

> 
> > +        Push `ruleset` into a temporary file on `host`, returning its path.
> +        """
> +        fd, name = mkstemp(prefix="FirewallMixin", text=True)
> +        write(fd, ruleset.encode('utf-8'))
> +        close(fd)
> +        remote_path = host.copy_file_to_machine(name)
> +        unlink(name)
> +        return remote_path
> +
> +    def apply_sub_configuration(self, config):
> +        super().apply_sub_configuration(config)
> +
> +        if not config.firewall_rulesets:
> +            config.firewall_rulesets = self.firewall_rulesets
> +
> +        self.stored_rulesets = {}
> 
> `stored_rulesets` should also probably be part of the `config` object

ACK.

> 
> > +        for host, ruleset in config.firewall_rulesets.items():
> +            desc.append("Firewall: ruleset with %d lines on host %s" % (len(ruleset.split("\n")), host))
> +
> +        return desc
> +
> +    def remove_sub_configuration(self, config):
> +        for host, ruleset in self.stored_rulesets.items():
> +            # XXX: this may spam debug log, but counter values may be worth logging
> +            logging.debug("Dropping Firewall ruleset on host %s:\n%s\n" % (host, self._dump_ruleset(host, config)))
> +            name = self._push_ruleset(host, ruleset)
> +            self._apply_ruleset(host, name)
> +            host.run('rm -f "%s"' % name)
> +
> +        return super().remove_sub_configuration(config)
> +
> +class BulkFirewallMixin(FirewallMixin):
> 
> i feel like this class - the generator method and the subconfig generator should be part of the base `FirewallMixin` class... i mean, without this, the subclasses don't really do anything on their own, and with this addition, they would still by default do the default "nothing" operation.
> 
> But I'm not sure if you had some more specific reason to separate this into its own class?

Hmm. I started with just FirewallMixin and implemented the bulk variant
in a second step. The non-bulk variant just used self.firewall_rulesets,
and with the addition of the bulk variant I started adding the current
ruleset to config. Hence the conditional config.firewall_rulesets
population in apply_sub_configuration(). I guess unifying them and
making the generator by default just use self.firewall_rulesets would
indeed simplify things.

> > +
> +    def _dump_ruleset(self, host, config):
> +        return host.run('nft list ruleset').stdout
> +
> +class Ip46tablesMixin(FirewallMixin):
> +    """
> +    A common base class for all the iptables-like FirewallMixin backends.
> +    Do not inherit directly, use one of the *tablesMixin classes instead.
> +    """
> +
> +    @property
> +    def iptables_command(self):
> +        raise NotImplementedError("Property iptables_command must be defined by a child class.")
> +
> +    def _apply_ruleset(self, host, path):
> +        host.run('%s-restore "%s"' % (self.iptables_command, path))
> 
> this would look nicer as an f-string:
> ```
> host.run(f'{self.iptables_command}-restore "{path}"')
> ```

ACK.

> > +            new_config.firewall_rulesets = ruleset
> +            yield new_config
> +
> +class NftablesMixin(FirewallMixin):
> +    """
> +    An nftables backend for FirewallMixin.
> +    """
> +
> +    def _apply_ruleset(self, host, path):
> +        host.run('nft flush ruleset')
> +        host.run('nft -f "%s"' % path)
> +
> +    def _dump_ruleset(self, host, config):
> +        return host.run('nft list ruleset').stdout
> +
> +class Ip46tablesMixin(FirewallMixin):
> 
> would it make more sense to name this as `IptablesBaseMixin` instead? seeing that there's also `ebtables` and `arptables` that inherit from this?

Good point, thanks! It's also consistent with other such base-classes.

> > +    def iptables_command(self):
> +        raise NotImplementedError("Property iptables_command must be defined by a child class.")
> +
> +    def _apply_ruleset(self, host, path):
> +        host.run('%s-restore "%s"' % (self.iptables_command, path))
> +
> +    def _extract_tables_n_chains(self, ruleset):
> +        out = {}
> +        curtable = None
> +        for line in ruleset.split('\n'):
> +            if line[0] == '*':
> +                out[line] = []
> +            elif line[0] == ':':
> +                words = line.split(' ')
> +                if not words[1] == '-': # ignore user-defined chains
> +                    out[curtable].append(words[0][1:])
> 
> `curtable` is always `None` here is that correct?

Oh, that's indeed totally broken.

[...]
> +    def _append_missing_parts(self, dst_ruleset, src_ruleset):
> +        for table, chains in self._extract_tables_n_chains(src_ruleset).items():
> +            tline = '*' + table + '\n'
> 
> the `table` here is coming from https://github.com/LNST-project/lnst/pull/364/files#diff-c1ca3132235b422fab430b7645942020bf1635e314f37db433a3ca03575f4e2dR129 which i think means that it fits the format of `*....` with the `\n` cut off

It's more or less the same bug: I should have extracted the table name
into curtable and use that as key instead of the whole input line.

> but here you assign `tline = '*' + "*...."` so you'll end up with double `*` characters is that correct?

Definitely not. Sorry for the buggy code, I surely tested this at some
point. Looks like some last minute rewrite in between. :(

> > +        unlink(name)
> +        return remote_path
> +
> +    def apply_sub_configuration(self, config):
> +        super().apply_sub_configuration(config)
> +
> +        if not config.firewall_rulesets:
> +            config.firewall_rulesets = self.firewall_rulesets
> +
> +        self.stored_rulesets = {}
> +        for host, ruleset in config.firewall_rulesets.items():
> +            self.stored_rulesets[host] = self._dump_ruleset(host, config)
> +
> +            name = self._push_ruleset(host, ruleset)
> +            self._apply_ruleset(host, name)
> +            host.run('rm -f "%s"' % name)
> 
> this is what removes the pushed files - i think this answers your other comment in https://github.com/LNST-project/lnst/pull/364/files#r1519638084

I wasn't sure if Jan suggested to instead register all those temp files
for removal in a common step but wanted to implement the suggested
changes first for a more informed reply.

BTW: I'm doing this temp file thing merely because it's not possible to
feed data via stdin to processes spawned by host.run(). But given the
multiple layers and possible unwanted side-effects of feeding large data
this way I guess the temp file method is the better choice, anyway.

Thanks, Phil

olichtne:
> You meant manual call of unlink(), right? Seems sensible, though I don't
see the benefit (but I see everything through procedural goggles
anyway).

yes, i mean the manual call of unlink()

The benefit is that in case there's an unhandled exception raised somewhere for some reason, the `with ... as fd:` block automatically takes care of the cleanup of the relevant objects.

When you're calling the `unlink` function manually you'd properly have to do a `try-except-finally` code structure and call the `unlink` from the `finally` block to ensure that it gets called.

> Hmm. I started with just FirewallMixin and implemented the bulk variant in a second step. The non-bulk variant just used self.firewall_rulesets,

yeah, i think the non-bulk variant is simply a special case of the bulk variant which is why it could be merged together.

> Sorry for the buggy code

no worries, this is why we do reviews :)

> BTW: I'm doing this temp file thing merely because it's not possible to
> feed data via stdin to processes spawned by host.run(). But given the
> multiple layers and possible unwanted side-effects of feeding large data
> this way I guess the temp file method is the better choice, anyway.

yeah... at some point it is a bit more beneficial to write a module/class that runs directly on the agent instead of running code on the controller and sending data to the agent, i think this isn't very well documented but you can do this:

```python
class MyCustomCodeClass():
    def method1():
        logging.debug("this runs on the agent")

class MyRecipe():
    def test():
        host1_custom_class = self.host1.init_class(MyCustomCodeClass)
        host1_custom_class.method()
```

more complex example can be seen in: https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/BasePvPRecipe.py#L144

which uses this "code class":  https://github.com/LNST-project/lnst/blob/master/lnst/RecipeCommon/LibvirtControl.py

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: DONE, thanks for pointing me at ABC and @abstractmethod. I wasn't aware such "fancy" stuff exists in Python.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: DONE, thanks for pointing me at ABC and @AbstractMethod. I wasn't aware such "fancy" stuff exists in Python.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: Indeed, thanks!

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: I converted the code to using f-strings everywhere. My Python-Fu is ~15 years old.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: DONE.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: By introducing the FirewallControl class running on the hosts, I was able to pretty easily ellminate use of these temporary files. Many thanks to @olichtne for pointing it out, this simplified the code quite a bit!

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: Merged both and they integrate nicely by making the generator method return '[self.firewall_rulesets]' by default.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: DONE.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: Fixed, thanks!

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
```diff
@@ -0,0 +1,155 @@
...
+    @property
+    def firewall_rulesets_generator(self):
+        """
+        A generator yielding { host: ruleset, ... } to apply for a test run.
+        Each yield will turn into a new sub configuration and thus be tested
+        separately.
+        """
+        return [self.firewall_rulesets]
+
+    def generate_sub_configurations(self, config):
```
olichtne, line 57: one more thing here, this method needs to call `super()` like this:
```python
def generate_sub_configurations(self, config):
    for parent_config in super().generate_sub_configurations(config):
        ...
        # the rest of the method but now using parent_config instead of config
```
As seen in other config mixins, e.g. https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/ConfigMixins/OffloadSubConfigMixin.py#L31C1-L31C74

the `generate_sub_configurations` method is in ENRT package designed as a [collaborative inheritance](https://medium.com/swlh/cooperative-multiple-inheritance-paradigm-in-python-f048b7ecdb29) method, this is part of the "ConfigMixin" design.

This is specific to the ENRT test set (so lnst normally doesn't use this it's just this specific test set). And the collaborative inheritance is a python speciality...

basically:

```python
SomeRecipe(ConfigMixin1, ConfigMixin2, BaseEnrtRecipe):
    pass

SomeRecipe2(ConfigMixin2, ConfigMixin1, BaseEnrtRecipe):
    pass
```

Means that if both config mixin classes define the `generate_sub_configurations` methods, the **collaborate** on the result via inheritance. The call order of the collaboration depends on the class definition and what the order of base classes was PLUS the way that you use the `super()` call in your collaborative methods.

In short - `super` calls "the next in line" method, not "strict parent" method and when every definition of the method in the MRO correctly uses the `super` call, they together pass around the same `config` object and each method adds its own part to it.

If one definition of the method in the class inheritance hierarchy DOESN'T call `super` correctly it interrupts the chain of "next in line" and the design stops working. This may be intentional in some cases but in this one it's not wanted.

olichtne (changes requested):
one last comment, otherwise this looks good to me.

lnst/Recipes/ENRT/ConfigMixins/FirewallMixin.py
SirPhuttel: Ah, indeed. Thanks for the details. I don't think colaborative inheritance is my problem, as I do call the super() object's method. But I missed the fact that it is a generator and thus will possibly return multiple different sub configs and I should mix into all of them.

olichtne (approved):

olichtne:
if there's no other comments from @jtluka tomorrow i can merge this in.

PR #365, opened 2024-03-21
Replace CoalescingHWConfigMixin with MultiCoalescingHWConfigMixin

jtluka:
### Description

The CoalescingHWConfigMixin is only capable of disabling adaptive TX/RX
setting. Typically the user disables adaptive coalescing and configures
specific values of rx/tx-usecs and others. Not setting these would
result in non-deterministic behaviour of the setup.

The MultiCoalescingHWConfigMixin is capable of setting any coalescing
options, so using it instead of CoalescingHWConfigMixin makes sense.

### Tests

Test internally in generic_team_tracker/-/issues/829

### Reviews

@olichtne


olichtne:
so i guess the main question now is, is the `CoalescingHWConfigMixin` used anywhere and does it have any actual use? maybe we should simply remove it at this point and rename the `MultiCoalescingHWConfigMixin` into just `CoalescingHWConfigMixin` as it supersedes all of the functionality and we "know" that the old implementation leads to unpredictable results which i think we never want?

olichtne:
and while we're at it, the `MultiCoalescingMixin` should probably include some documentation of the dictionary format for the `coalescing_settings` parameter... some example or something?

jtluka:
> so i guess the main question now is, is the `CoalescingHWConfigMixin` used anywhere and does it have any actual use? maybe we should simply remove it at this point and rename the `MultiCoalescingHWConfigMixin` into just `CoalescingHWConfigMixin` as it supersedes all of the functionality and we "know" that the old implementation leads to unpredictable results which i think we never want?

The mixin is currently used internally in ShortLivedConnectionsRecipe only with `"adaptive_tx_coalescing": true` and `"adaptive_rx_coalescing": true`. All of the production ready tests are in old lab that will be decomissioned, so imo we still could replace the mixin as you suggested as the impact would be minimal.

jtluka:
> and while we're at it, the `MultiCoalescingMixin` should probably include some documentation of the dictionary format for the `coalescing_settings` parameter... some example or something?

Good idea, I'll include it in this MR.

jtluka:
> > so i guess the main question now is, is the `CoalescingHWConfigMixin` used anywhere and does it have any actual use? maybe we should simply remove it at this point and rename the `MultiCoalescingHWConfigMixin` into just `CoalescingHWConfigMixin` as it supersedes all of the functionality and we "know" that the old implementation leads to unpredictable results which i think we never want?
> 
> The mixin is currently used internally in ShortLivedConnectionsRecipe only with `"adaptive_tx_coalescing": true` and `"adaptive_rx_coalescing": true`. All of the production ready tests are in old lab that will be decomissioned, so imo we still could replace the mixin as you suggested as the impact would be minimal.

olichtne (approved):

PR #369, opened 2024-05-28
Tests/Ping: log stdout

jtluka:
### Description

Sometimes it's useful to have ping stdout available to analyze the connectivity issues.

### Tests

Tested in RH Beaker lab J:9341216

### Reviews

@olichtne


olichtne (changes requested):
you can also add the stdout to `self._res_data['stdout']` so that this is accessible from the https://github.com/LNST-project/lnst/blob/master/lnst/Controller/Job.py#L88 job object and so that this is then also accessible in the lrc file.

olichtne (approved):

PR #370, opened 2024-05-28
Conntrack: insertion rate recipe, full table insertion rate recipe

enhaut:
### Description
* added `WaitForConditionModule` which implements waiting for condition on _remote_ agent
* FirewallMixin: fixed pickling for dynamic derived classes
* added conntrack insertion rate recipe
* added conntrack full table insertion rate recipe

### Tests
Waiting for condition module and firewall mixin changes are used in conntrack recipes, so test run of both conntrack insertion rate and ct full table insertion rate recipes is sufficient to test all the changes from this mr.

`J:9753788`

Closes https://github.com/LNST-project/lnst/issues/372

### DIFF
--- a/lnst/Controller/Namespace.py
+++ b/lnst/Controller/Namespace.py
@@ -24,6 +24,7 @@ from lnst.Devices.RemoteDevice import RemoteDevice
 from lnst.Controller.Common import ControllerError
 from lnst.Controller.Job import Job
 from lnst.Controller.RecipeResults import ResultLevel
+from lnst.Common.conditions.WaitForConditionModule import WaitForConditionModule
 
 class HostError(ControllerError):
     pass
@@ -128,6 +129,12 @@ class Namespace(object):
         job = self.prepare_job(what, fail, json, desc, job_level)
         job.start(bg, timeout)
         return job
+    
+    def wait_for_condition(self, condition: WaitForConditionModule):
+        job = self.prepare_job(condition)
+        job.start(bg=True)
+
+        return self._machine.wait_for_job(job, condition.timeout)
 
     def __getattr__(self, name):
         """direct access to Device objects
--- a/lnst/Recipes/ENRT/CTFulltableInsertionRateRecipe.py
+++ b/lnst/Recipes/ENRT/CTFulltableInsertionRateRecipe.py
@@ -1,0 +1,11 @@
+from .CTInsertionRateNftablesRecipe import CTInsertionRateNftablesRecipe
+from .ConfigMixins.LongLivedConnectionsMixin import LongLivedConnectionsMixin
+
+class CTFulltableInsertionRateRecipe(LongLivedConnectionsMixin, CTInsertionRateNftablesRecipe):
+    """
+    The recipe measures the insertion rate of new entries to already full conntrack table.
+    Since conntrack table is implemented as hash table, "full" depends on total count of 
+    available buckets. Should be more than 75%. 
+    """
+    pass
+
--- a/lnst/Recipes/ENRT/CTInsertionRateNftablesRecipe.py
+++ b/lnst/Recipes/ENRT/CTInsertionRateNftablesRecipe.py
@@ -1,0 +1,30 @@
+from .SimpleNetworkRecipe import SimpleNetworkRecipe
+from .ConfigMixins.NftablesConntrackMixin import NftablesConntrackMixin
+
+from lnst.Common.Parameters import ConstParam
+from lnst.RecipeCommon.Perf.Evaluators import NonzeroFlowEvaluator
+
+
+class CTInsertionRateNftablesRecipe(NftablesConntrackMixin, SimpleNetworkRecipe):
+    """
+    The recipe measures insertion rate of conntrack entries on receiver side.
+    This is done by using tcp_crr test which opens connections (that trigger 
+    adding a new conntrack entry), sends request, waits for response and closes
+    the socket. Thats done in a sequence, so we indirectly measure performance
+    of conntrack table.
+
+    It's important to keep perf_msg_sizes as small as possible, as transferring
+    any amount of data take some time and tcp_crr works in a sequence, so data
+    transfer blocks the process of opening new connections and affects the
+    results.
+    """
+    net_perf_tool = ConstParam("neper")
+    perf_tests = ConstParam(["tcp_crr"])
+    perf_msg_sizes = ConstParam([1])
+
+    @property
+    def net_perf_evaluators(self):
+        return [
+            NonzeroFlowEvaluator(["generator_results"])
+        ]  # only generator measures CCs, receiver always reports 0
+
--- a/lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
@@ -1,0 +1,220 @@
+import time
+import signal
+import logging
+from math import ceil
+
+from lnst.Tests.LongLivedConnections import LongLivedServer, LongLivedClient
+from lnst.Common.IpAddress import interface_addresses
+from lnst.Common.Parameters import IntParam
+from lnst.Common.Parameters import IPv4NetworkParam, IPv6NetworkParam
+from lnst.Common.conditions.WaitForEstablishedConnections import (
+    WaitForEstablishedConnections,
+)
+from lnst.RecipeCommon.Perf.PerfTestMixins.BasePerfTestTweakMixin import (
+    BasePerfTestTweakMixin,
+)
+
+
+class LongLivedConnectionsMixin(BasePerfTestTweakMixin):
+    """
+    This mixin adds support for long-lived connections.
+
+    Based on `long_lived_conns` parameter, it will create a number of
+    long-lived connections between the hosts. Receiver is a server, while
+    generator is a client.
+
+    Only long lived connections IPs are handled by this mixin.
+    Therefore, if your test requires perf IPs, it should be
+    configured by parent's test_wide_configuration() method.
+    IPs used for long-lived client and server are added based on
+    long_lived_conns_per_ip parameter, which defines size of
+    addressable space for connections. Sinde L4 can address up
+    to 65535 ports, it's obviously limited by that. Parameters
+    long_lived_conns_net4 and long_lived_conns_net6 are required
+    to define IP address space for connections.
+
+    Connections are not equally distributed among clients.
+    The first client (and others) will get long_lived_conns_per_ip
+    connections, while the last one will get the remaining connections.
+
+    Don't forget to set appropriate system-wide NO_FILES ulimit (if needed).
+    See LongLivedServer/LongLivedClient for more details.
+    """
+
+    long_lived_conns = IntParam(mandatory=True)
+    long_lived_conns_port = IntParam(default=20000)
+    long_lived_conns_per_ip = IntParam(default=20000)
+    long_lived_conns_net4 = IPv4NetworkParam(default="192.168.102.0/24", mandatory=True)
+    long_lived_conns_net6 = IPv6NetworkParam(default="fc01::/64", mandatory=True)
+
+    def test_wide_configuration(self):
+        host1, host2 = self.matched.host1, self.matched.host2
+        host1.eth0.keep_addrs_on_down()
+        host2.eth0.keep_addrs_on_down()
+
+        config = super().test_wide_configuration()
+        # L4 can address up to 65535 ports (size of addresable space
+        # defined in long_lived_conns_per_ip) Therefore opening connections
+        # may require multiple IPs.
+
+        ipv4_addr = interface_addresses(self.params.long_lived_conns_net4)
+        ipv6_addr = interface_addresses(self.params.long_lived_conns_net6)
+        config.long_lived_ips = {
+            host1: {"ipv4": [], "ipv6": []},
+            host2: {"ipv4": [], "ipv6": []},
+        }
+
+        host1.eth0.down()
+        host2.eth0.down()
+        for host in [host1, host2]:
+            for _ in range(self.servers_count):
+                ipv4 = next(ipv4_addr)
+                ipv6 = next(ipv6_addr)
+
+                host.eth0.ip_add(ipv4)
+                host.eth0.ip_add(ipv6)
+                config.long_lived_ips[host]["ipv4"].append(ipv4)
+                config.long_lived_ips[host]["ipv6"].append(ipv6)
+
+        host1.eth0.up()
+        host2.eth0.up()
+
+        self.wait_tentative_ips(config.configured_devices)
+
+        return config
+
+    def test_wide_deconfiguration(self, config):
+        host1, host2 = self.matched.host1, self.matched.host2
+
+        host1.eth0.remove_addrs_on_down()
+        host2.eth0.remove_addrs_on_down()
+
+        return super().test_wide_deconfiguration(config)
+
+    @property
+    def servers_count(self):
+        return ceil(self.params.long_lived_conns / self.params.long_lived_conns_per_ip)
+
+    def generate_perf_configurations(self, config):
+        for parent_config in super().generate_perf_configurations(config):
+            parent_config.long_lived_connections = []
+
+            for client_job, server_job in self.generate_jobs(config):
+                parent_config.long_lived_connections.append((client_job, server_job))
+
+            yield parent_config
+
+    def generate_jobs(self, config):
+        for ip_version in self.params.ip_versions:
+            host1, host2 = self.matched.host1, self.matched.host2
+            filtered_ips = zip(
+                config.long_lived_ips[host1][ip_version],
+                config.long_lived_ips[host2][ip_version],
+            )
+
+            for i, endpoint_pair in enumerate(filtered_ips):
+                generator_ip, receiver_ip = endpoint_pair
+
+                connections_count = self.calculate_client_connections(i)
+
+                server_job = self._prepare_server(
+                    host2.eth0, receiver_ip, connections_count
+                )
+                client_job = self._prepare_client(
+                    host1.eth0, receiver_ip, generator_ip, connections_count
+                )
+
+                yield client_job, server_job
+
+    def calculate_client_connections(self, client_id):
+        client_id += 1  # 0-based index
+        if client_id < self.servers_count:
+            return self.params.long_lived_conns_per_ip
+
+        opened_connections = (
+            self.servers_count - 1
+        ) * self.params.long_lived_conns_per_ip
+
+        # remaining connections are handled by last client
+        return self.params.long_lived_conns - opened_connections
+
+    def _prepare_server(self, receiver_nic, receiver_ip, conns_count):
+        server = LongLivedServer(
+            server_ip=receiver_ip,
+            server_port=self.params.long_lived_conns_port,
+            connections_count=conns_count,
+        )
+
+        job = receiver_nic.netns.prepare_job(server)
+
+        return job
+
+    def _prepare_client(self, generator_nic, receiver_ip, generator_ip, conns_count):
+        client = LongLivedClient(
+            server_ip=receiver_ip,
+            server_port=self.params.long_lived_conns_port,
+            client_ip=generator_ip,
+            connections_count=conns_count,
+        )
+
+        job = generator_nic.netns.prepare_job(client)
+
+        return job
+
+    def generate_perf_configuration_description(self, config):
+        desc = super().generate_perf_configuration_description(config)
+
+        for client_job, server_job in config.long_lived_connections:
+            desc.append(
+                f"Long-lived connection between {client_job.what} and {server_job.what}"
+            )
+
+        return desc
+
+    def apply_perf_test_tweak(self, config):
+        super().apply_perf_test_tweak(config)
+
+        for _, server_job in config.long_lived_connections:
+            server_job.start(bg=True)
+
+        time.sleep(2)  # just to be sure servers are up
+
+        for client_job, _ in config.long_lived_connections:
+            client_job.start(bg=True)
+
+        for version in self.params.ip_versions:
+            for stream in self.params.perf_tests:
+                addr = (
+                    self.params.long_lived_conns_net4
+                    if version == "ipv4"
+                    else self.params.long_lived_conns_net6
+                )
+
+                condition = WaitForEstablishedConnections(
+                    addr, stream, self.params.long_lived_conns, timeout=300
+                )
+                prepared = self.matched.host2.wait_for_condition(condition)
+
+        logging.info("Long-lived connections established")
+
+    def remove_perf_test_tweak(self, config):
+        for client_job, server_job in config.long_lived_connections:
+            kwargs = {}
+
+            if client_job.what.runtime_estimate():
+                kwargs["timeout"] = client_job.what.runtime_estimate()
+            else:
+                client_job.kill(signal.SIGINT)
+                server_job.kill(signal.SIGINT)
+                # no timeout defined => it'll use default timer
+
+            try:
+                client_job.wait(**kwargs)
+                server_job.wait(**kwargs)
+            finally:
+                client_job.kill()
+                server_job.kill()
+
+        del config.long_lived_connections
+
+        return super().remove_perf_test_tweak(config)
--- a/lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
@@ -1,0 +1,52 @@
+import time
+
+from .FirewallMixin import NftablesMixin
+
+
+class NftablesConntrackMixin(NftablesMixin):
+    @NftablesMixin.firewall_rulesets.getter
+    def firewall_rulesets(self):
+        nic = self.matched.host2.eth0.name
+        ruleset = (
+            f"""table inet NotrackUnrelatedTraffic {{
+  chain controllerInfNoTrackIn {{
+    type filter hook prerouting priority raw;
+
+    iif != {nic} counter notrack
+    policy accept
+  }}
+  chain controllerInfNoTrackOut {{
+    type filter hook output priority raw;
+
+    iif != {nic} counter notrack
+    policy accept
+  }}
+  chain InterfaceUnderTest {{
+    type filter hook input priority 0; policy accept;
+
+    # Allow established connections to pass
+    iif {nic} ct state {{ new, established, related }} accept
+  }}
+}}"""
+        )
+        return {self.matched.host2: ruleset}
+
+    def apply_sub_configuration(self, config):
+        super().apply_sub_configuration(config)
+        host2 = self.matched.host2
+
+        host2.run("sysctl -w net.netfilter.nf_conntrack_tcp_timeout_syn_sent=3600")
+        # ^^ unloading the module in remove_sub_configuration() will reset it
+
+    def remove_sub_configuration(self, config):
+        # conntrack-related modules to be unloaded to not 
+        # interfere with following tests. Also, it flushes
+        # conntrack table.
+        super().remove_sub_configuration(config)
+
+        host2 = self.matched.host2
+
+        host2.run("systemctl stop nftables")
+        time.sleep(2)
+        host2.run("modprobe -rv nft_ct nf_conntrack_netlink nf_conntrack nft_counter")
+
--- a/lnst/Tests/LongLivedConnections.py
+++ b/lnst/Tests/LongLivedConnections.py
@@ -1,0 +1,152 @@
+import time
+import select
+import socket
+import logging
+import resource
+import threading
+
+from .BaseTestModule import BaseTestModule
+from lnst.Common.Parameters import IntParam, IpParam
+
+
+class BaseLongLivedTestModule(BaseTestModule):
+    server_ip = IpParam(mandatory=True)
+    server_port = IntParam(mandatory=True)
+    duration = IntParam(default=0)  # 0 means run indefinitely, until SIGINT is received
+    connections_count = IntParam(default=1)
+
+    def run(self):
+        self._running = True
+        soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+
+        self._set_rlimit(hard, hard)
+        self._start()
+        if self.params.duration:
+            time.sleep(self.params.duration)
+        else:
+            try:
+                while True:
+                    time.sleep(1)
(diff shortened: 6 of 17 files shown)


enhaut:
This is ready to review, I am just still working on rewriting waiting for opening all the connections

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Don't forget to set appropriate NO_FILES ulimit (if needed).
+    See LongLivedServer/LongLivedClient for more details.
+    """
+
+    long_lived_conns = IntParam(mandatory=True)
+    long_lived_conns_port = IntParam(default=20000)
+    long_lived_conns_per_ip = IntParam(default=20000)
+
+    def _generate_ip_endpoints(self, config):
```
olichtne, line 43: please reorder the methods top -> down in order of abstraction/importance as we have in other recipes/mixins...

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def calculate_client_connections(self, client_id):
+        client_id += 1  # 0-based index
+        if client_id < self.servers_count:
+            return self.params.long_lived_conns_per_ip
+
+        # remaining connections are handled by last client
+        return (
+            self.params.long_lived_conns
+            - (self.servers_count - 1) * self.params.long_lived_conns_per_ip
```
olichtne, line 95: isn't this just `self.params.long_lived_conns % self.servers_count`? as in modulo (remainder)

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+            server_port=self.params.long_lived_conns_port,
+            client_ip=generator.address,
+            connections_count=conns_count,
+        )
+
+        job = generator.device.netns.prepare_job(client)
+
+        return job
+
+    def wait_agent_for_establish_conns(self, agent, timeout=60):
```
olichtne, line 121: discussed live in a tech meeting, IMO you should be able to implement a Test module for this instead:
```
agent.run(ConnectionsEstablished(num_of_connections), timeout=timeout)
```
where
```
class ConnectionsEstablished(BaseTestModule):
     num_of_connections = IntParam
     def run(self):
        while True:
            stdout = exec(ss | ESTAB | wc -l)
            if int(stdout) >= num_of_connections:
               return True
```

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+        # L4 can address up to 65535 ports (size of addresable space
+        # defined in long_lived_conns_per_ip) Therefore opening connections
+        # may require multiple IPs.
+        host1, host2 = self.matched.host1, self.matched.host2
+
+        ipv4_addr = interface_addresses(self.params.net_ipv4)
+        ipv6_addr = interface_addresses(self.params.net_ipv6)
+
+        for _ in range(2):
+            # 2 addresses are already assigned by super().test_wide_configuration()
```
olichtne, line 150: **any** number of addresses may be assigned by the base Recipe, this uses an assumption that that equals 2, but technically we should probably reuse the ip address network iterator from the base Recipe - e.g. it can be passed via the `config` object:
```
BaseRecipe():
    def testwide():
        config.ipv4_addr_net = interface_addresses(self.params.net_ipv4)

LongLivedConnectionsMixin:
    def testwide():
        config = super().testwide()
        ipv4_addr = config.ipv4_addr_net
```

I realize this requires updating the base recipes that this mixin can be applied to, maybe there's a way to detect the "last used" address in a different way, but this is the "clean" way to implement this.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+            # those are used for perf tests. Generator needs to be moved to 3rd address
+            next(ipv4_addr)
+            next(ipv6_addr)
+
+        host1.eth0.down()
+        host2.eth0.down()
+        for host in [host1, host2]:
+            for _ in range(self.servers_count):
+                config.configure_and_track_ip(host.eth0, next(ipv4_addr))
+                config.configure_and_track_ip(host.eth0, next(ipv6_addr))
```
olichtne, line 160: since we're splitting the ip addresses into two categories:
1. perf addresses implemented by the base recipe
2. addresses for the long lived connections

we should IMO split them in how we track them in the `config` object as well - this way they mix for how the `generate_perf_endpoints` uses them and so you have to do that weird slicing override, if these ip addresses were split you wouldn't have to do that.

olichtne (changes requested):

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,174 @@
...
+import time
+import select
+import resource
+from typing import Union
+
+from lnst.Common.IpAddress import Ip4Address, Ip6Address, BaseIpAddress
+from lnst.Common.Parameters import IpParam, IntParam, ListParam, StrParam
+
+
+def get_inet(ip: Union[Ip4Address, Ip6Address]):
```
jtluka, line 14: I'd rename this to `get_af()`, it is address family, result can be either AF_INET or AF_INET6

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,174 @@
...
+    return socket.AF_INET
+    if isinstance(ip, Ip4Address):
+        return socket.AF_INET
+    elif isinstance(ip, Ip6Address):
+        return socket.AF_INET6
+    else:
+        raise ValueError(f"Unknown IP address type: {ip}")
+
+
+class LongLivedMixin(BaseTestModule):
```
jtluka, line 24: Why we call this a mixin ? To me this seems to be just a base class for the server, client implementations.

I'd understand if we do inheritance with:
```
class LongLivedMixin():
   pass

class LongLivedServer(BaseTestModule, LongLivedMixin):
   pass

class LongLivedClient(BaseTestModule, LongLivedMixin):
   pass
```

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,174 @@
...
+
+        return sck
+
+    def _start(self):
+        for _ in range(self.params.connections_count):
+            conn = self._start_connection()
+            self._connections.append(conn)
+
+        logging.info("All connections established, waiting")
+        print("Connected to all")
```
jtluka, line 164: Debug statement leftover? Move to logging if useful, but seems the previous statement basically does the same.

.github/runner_full.py
jtluka, line 1: I'd probably mark this more visible with TODO: REMOVE or so, no big deal, just got confused a bit.

jtluka (commented):
Took a look at all of the code, few minor comments, but otherwise looks ok.

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,174 @@
...
+    def _start_connection(self):
+        sck = socket.socket(get_inet(self.params.server_ip), socket.SOCK_STREAM)
+        sck.bind(
+            (self.params.client_ip, 0)
+        )  # needs to be binded to specific IP to respect flow IPs
+        sck.connect((str(self.params.server_ip), self.params.server_port))
+
+        sck.setblocking(False)
+        #  ^^^ sets socket into a "something may be written to there" state,
+        # so it won't close connection when jumping out of this function context
```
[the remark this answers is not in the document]
enhaut: Python tries to save resources, so it closes all non-used connections when leaving the context. We don't read/write to the connection, so I guess, it thinks it's not used at all and closes it. 

Not sure, but I believe it's the same for files or any other resource.

lnst/Tests/LongLivedConnections.py
enhaut: Yep, thanks.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
enhaut: Not really, there are couple of edge cases comming from `%` returns 0 if `long_lived_conns` is dividable by `servers_count`, so it would need to look like:


```python
    def calculate_client_connections(self, client_id):
        client_id += 1  # 0-based index
        remaining_conns = self.params.long_lived_conns % self.servers_count

        if client_id < self.servers_count or not remaining_conns:
            if self.params.long_lived_conns < self.params.long_lived_conns_per_ip:
                return self.params.long_lived_conns

            return self.params.long_lived_conns_per_ip

        # remaining connections are handled by last client
        return remaining_conns
```

But i've refactored my implementation a bit, hope it's more readable, wdyt?

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
enhaut: Yep, I did something similar to this but more generic, see `Agent: waiting for condition on agent side` and `Agent/Conditions: Waiting for connections establishment` commits

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
enhaut: Oh, thats something i didn't count with. Thanks.

I've added a separate `long_lived_conns_net{4,6}` params.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+        return job
+
+    def wait_agent_for_establish_conns(self, agent, timeout=60):
+        # Yeah, a bit hacky but simplest way of waiting.
+        # The problem is, that LNST doesn't support waiting
+        # for a specific condition at a REMOTE agent.
+
+        # TODO: refactor to use condition wait, when implemented
+        while timeout:
+            job = agent.run("ss | grep ESTAB | wc -l")
```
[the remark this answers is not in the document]
enhaut: Well yes, but actually no. There are a couple (up to 10) other `ESTAB` connections on the machine, so it's fine to count just those as we plan to use the mixin for hundreds of connections.

Anyway, i've reworked this a bit and now, i am counting connections based on IPs, as you mentioned.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def remove_sub_configuration(self, config):
+        # conntrack-related modules to be unloaded to not 
+        # interfere with following tests. Also, it flushes
+        # conntrack table.
+        super().remove_sub_configuration(config)
+
+        host2 = self.matched.host2
+
+        # NOTE: sleeps are necessary to let kernel actually unload the modules
```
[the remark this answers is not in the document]
enhaut: They are. I don't know if `rmmod` is blocking, but it seems, it's not. So it requires a sleep in between.

Eventually, I can replace sleep with conditional wait on agent, but that might be a bit overkill.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
```diff
@@ -0,0 +1,67 @@
...
+    @NftablesMixin.firewall_rulesets.getter
+    def firewall_rulesets(self):
+        nic = self.matched.host2.eth0.name
+        ruleset = (
+            """table inet NotrackUnrelatedTraffic {
+  chain controllerInfNoTrackIn {
+    type filter hook prerouting priority raw;
+
+    iif != """
+            + nic
```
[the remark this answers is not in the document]
enhaut: sure

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
```diff
@@ -0,0 +1,67 @@
+import time
+
+from .FirewallMixin import NftablesMixin
+from lnst.Recipes.ENRT.ConfigMixins.FirewallMixin import NftablesMixin
+
+
+class NftablesConntrackMixin(NftablesMixin):
+    @NftablesMixin.firewall_rulesets.getter
```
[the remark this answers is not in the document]
enhaut: No, `@property` decorator overrides getters, setters and deleters within class. `NftablesMixn.firewall_rulesets.getter` overrides `getter` only.

So if that was `@property` I would need to add a setter method.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+        for client_job, server_job in config.long_lived_connections:
+            client_job.kill(signal.SIGINT)
+            server_job.kill(signal.SIGINT)
+            # ^^ both server and client runs in while(1) loop interruptable by SIGINT
+            # which shutdowns them down gracefully
+
+            try:
+                client_job.wait(timeout=client_job.what.runtime_estimate())
+                server_job.wait(timeout=server_job.what.runtime_estimate())
+            finally:
```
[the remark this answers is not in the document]
enhaut: I don't think it's needed. If `.wait()` throws an error, this block won't catch it, so the program crashes with exception from `wait` and prints out the traceback.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
enhaut: Makes sense, thx.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,230 @@
...
+                for endpoint_pair in parallel_endpoints
+                if isinstance(endpoint_pair.first.address, Ip6Address)
+            ]
+
+            ips.append(ipv4[slicer])
+            ips.append(ipv6[slicer])
+
+        return ips
+
+    def generate_perf_endpoints(self, config):
```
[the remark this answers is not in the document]
enhaut: fixed as part of changes below.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
enhaut: done

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+class LongLivedMixin(BaseTestModule):
+    server_ip = StrParam(mandatory=True)
+    server_port = IntParam(mandatory=True)
+    duration = IntParam(default=0)  # 0 means run indefinitely, until SIGINT is received
+    connections_count = IntParam(default=1)
+
+    def runtime_estimate(self):
+        return self.params.duration + 5
```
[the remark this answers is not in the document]
olichtne: i see, you're using the runtime estimate in the `wait` call here: https://github.com/LNST-project/lnst/pull/370/files#diff-92e500102f38565b7808a01046bb7ee6ecb108b5ec09b1ffadd9ef84eea44ba8R189

i think that could be an incorrect interpretation:
1. if `duration != 0` -> the `SIGINT` shouldn't be sent before the runtime estimate expires
2. if `duration == 0` -> sending the `SIGINT` is immediate and the `wait` call should use the `default` timeout to account for Agent-Controller connection delays. the `kill` call should be done immediately after `wait` times out regardless of a runtime estimate.

lnst/Tests/LongLivedConnections.py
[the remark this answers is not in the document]
olichtne: interesting, did you find some documentation explaining this? sounds like something to know more about...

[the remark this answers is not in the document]
enhaut: Nope, I believe this might be something interpreter-implementation specific. 

Here is simple reproducer:

```python
import socket as s
import psutil
import os


def print_open_conns():
    proc = psutil.Process(os.getpid())
    print(proc.connections())

def open_conn():
    print_open_conns()

    sck = s.socket(s.AF_INET, s.SOCK_STREAM)
    sck.connect(("1.1.1.1", 443))

    return sck

def client():
    print_open_conns()
    open_conn()  # not using sck after return => connection gets closed
    print_open_conns()

    print(open_conn())  # using sck after return => connection stays open
    print_open_conns()


if __name__ == "__main__":
    client()
```

I found an stack overflow thread discussing same behaviour for files, so i guess, this is the same case - https://stackoverflow.com/a/2404671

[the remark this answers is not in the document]
olichtne: right, but that's what my original question was about - you DO use the socket outside of the function that creates it and calls the `setblocking` - so that fits the reproducer case of `using sck after return -> connection stays open` you use it here: https://github.com/LNST-project/lnst/pull/370/files#diff-83a6e9f27f93f0570d09ced019772682991ea12c40d7f5416d5222630f28d786R142

lnst/Tests/LongLivedConnections.py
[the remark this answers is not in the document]
enhaut: Oh i see your point, reworked.

lnst/Tests/LongLivedConnections.py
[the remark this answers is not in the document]
enhaut: I wouldn't say variable assignment, is using the reference, the `print` function in reproducer is accessing socket object. 

Anyway, no idea how is this supposed to work, I believe it's closed by some sort of optimizer.

lnst/Agent/conditions/WaitForConditionModule.py
```diff
@@ -0,0 +1,49 @@
+import time
```
olichtne, line 1: IMO, this shouldn't be part of the `lnst.Agent` package, the main reason is that you define the instance of the class on the controller. Theoretically, lnst packages are designed in a way that the controller machine/environment doesn't have the Agent package available at all. This means that you wouldn't be able to properly import the class on the controller.

IMO this should be in one of:
* the `lnst.Common` package
* the `lnst.RecipeCommon` package
* a new `lnst.<some new package>` that is universal for both the agent and the controller installations (similar to how the `lnst.Devices` package exists), but imo this isn't correct for this case.

lnst/Agent/conditions/WaitForConditionModule.py
```diff
@@ -0,0 +1,49 @@
...
+    def _sig_handler(signum, _):
+        if signum == signal.SIGALRM:
+            raise TimeoutError("Timeout reached")
+
+        return
+    
+    @abstractmethod
+    def _condition(self):
+        pass
+
```
olichtne, line 49: you have an empty line at the end here.

lnst/Agent/conditions/WaitForConditionModule.py
```diff
@@ -0,0 +1,49 @@
...
+    @staticmethod
+    def _sig_handler(signum, _):
+        if signum == signal.SIGALRM:
+            raise TimeoutError("Timeout reached")
+
+        return
+    
+    @abstractmethod
+    def _condition(self):
```
olichtne, line 47: i'm thinking if maybe we should have a default implementation of this condition that is provided via an init parameter?

```
host.wait_for_condition(WaitForConditionModule(condition=some_function))
```

you could still override this via a derived class, but you could also use this for simple lambda function conditions without having to define an extra class for them.

lnst/Agent/conditions/WaitForEstablishedConnections.py
```diff
@@ -0,0 +1,58 @@
+import os
```
olichtne, line 1: again, i would move this entire package under `lnst.Common`

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,171 @@
...
+                logging.info("Interrupted, dying...")
+
+        self._stop()
+        self._set_rlimit(hard, soft)
+
+        return (
+            self._result
+        )  # is set as part of _stop (if # of conns matches expected # of conns)
+
+    def _set_rlimit(self, hard: int, soft: int):
```
olichtne, line 55: the order of these should be `(self, soft, hard)` so that we keep consistency with the `getrlimit` method...

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,171 @@
...
+        self._listening_thread = threading.Thread(target=self._listen)
+        self._listening_thread.start()
+
+        self._polling_thread = threading.Thread(target=self._poll_connections)
+        self._polling_thread.start()
+
+    def _stop(self):
+        self._result = (
+            True if len(self._connections) == self.params.connections_count else False
+        )
```
olichtne, line 86: this should be called AFTER `self._running = False` imo.

lnst/Controller/Namespace.py
```diff
@@ -265,3 +265,9 @@ def _update_device_id(self, value, name):
             return
 
         dev._id = name
+
+    def keep_addrs_on_dev_down(self, device: Device) -> None:
+        self.run(f"echo 1 > /proc/sys/net/ipv6/conf/{device.name}/keep_addr_on_down")
```
olichtne, line 270: this sounds like it should be a method for the `Device` class instead of in the `Namespace` class...

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+
+        return desc
+
+    def apply_sub_configuration(self, config):
+        super().apply_sub_configuration(config)
+
+        for client_job, server_job in config.long_lived_connections:
+            server_job.start(bg=True)
+            time.sleep(2)  # just to be sure the server is up
```
olichtne, line 174: depending on the number of client-server pairs it may be more sensible to split this into:
```
for _, server_job in config.long_lived_connections:
    server_job.start(bg=True)
time.sleep(2) # wait for all servers to start
for client_job, _ in config.long_lived_connections:
    client_job.start(bg=True)
```

this could save some extra sleeps.

olichtne (changes requested):
looks much better, just some minor comments added.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
```diff
@@ -0,0 +1,60 @@
+import time
+
+from .FirewallMixin import NftablesMixin
+from lnst.Recipes.ENRT.ConfigMixins.FirewallMixin import NftablesMixin
```
jtluka, line 4: These two imports are identical, or am I missing something?

lnst/Recipes/ENRT/CTInsertionRateNftablesRecipe.py
```diff
@@ -0,0 +1,18 @@
+from .SimpleNetworkRecipe import SimpleNetworkRecipe
+from .ConfigMixins.NftablesConntrackMixin import NftablesConntrackMixin
+
+from lnst.Common.Parameters import ConstParam
+from lnst.RecipeCommon.Perf.Evaluators import NonzeroFlowEvaluator
+
+
+class CTInsertionRateNftablesRecipe(NftablesConntrackMixin, SimpleNetworkRecipe):
```
jtluka, line 8: I'm a bit confused about this class name. It says "insertion rate", but it does not measure anything related to rule insertion - could you please clarify?

Maybe a documentation would help me to understand what this recipe does, so please add a docstring.

lnst/Recipes/ENRT/CTFulltableInsertionRateRecipe.py
```diff
@@ -0,0 +1,6 @@
+from .CTInsertionRateNftablesRecipe import CTInsertionRateNftablesRecipe
+from .ConfigMixins.LongLivedConnectionsMixin import LongLivedConnectionsMixin
+
+class CTFulltableInsertionRateRecipe(LongLivedConnectionsMixin, CTInsertionRateNftablesRecipe):
```
jtluka, line 4: Please add a documentation what this test does.

lnst/Agent/conditions/WaitForConditionModule.py
enhaut: The problem of this approach is, that the condition needs to be pickled and sent to the host. 

However, python's `pickle` module doesn't really pickles whole classes/functions, just a reference with instance data therefore we can't use `lamba` functions. Functions are pickled as a `module_path.function_name` so lambda functions cannot be pickled, as they don't have the name.

Support for "function condition" wouldn't be a problem. However, if the condition function is located in the same file as the recipe is, recipe module is then loaded on the host side which caused a weird cyclic import error. Spent some time trying to make it work but then I decided to simply move conditions to separate class and module which solves both of the problems.

There is also python library called `dill` which transfers whole class+instance data or lambda functions but that would make this MR even larger.

olichtne: agreed with the lambda part, discussed the named function a bit more in person, i'm interested in the generated errors to check if there isn't a simple solution.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
olichtne: thanks, looks better.

lnst/Recipes/ENRT/ConfigMixins/LongLivedConnectionsMixin.py
olichtne: ack, made new comments on those, resolving this older conversation...

lnst/Recipes/ENRT/CTInsertionRateNftablesRecipe.py
jtluka: Thanks, looks ok now.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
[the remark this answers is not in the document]
jtluka: I took a look at the man page and have two ideas:
* modprobe can take more than one module to unload, so maybe we could do `modprobe -r nft_ct nf_conntrack_netlink nf_conntrack nft_counter`.
* man page states that "There is usually no reason to remove modules" - do we really need to unload the modules? I understand the more cleanup we do, less potential impact to subsequent tests user can execute after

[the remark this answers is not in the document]
jtluka: Taking back second one as the comment in the code states that this is required to cleanup the ct table - BUT, is it possible to clear the table by other means?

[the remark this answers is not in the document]
jtluka: Maybe somehow through `/proc/sys/net/netfilter/nf_conntrack*` entries?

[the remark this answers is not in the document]
olichtne: why is `conntrack-tools` an overkill dependency to add for a ... conntrack test?

lnst/Tests/LongLivedConnections.py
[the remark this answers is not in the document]
olichtne: > The reproducer is supposed to reproduce closing the connection when leaving context

and it isn't reproducing....

the discussion is about what "usage" means only because that's the reason you used to explain your workaround... this reason doesn't seem to actually match reality, and your reproducer seems to prove that...

so there is something else happening, which i'd like to understand before introducing a weird workaround that no-one will be able to maintain or debug into the future when we justify it with reasons that we know aren't true.

I don't believe this should require digging into python interpreter source either...

again, i won't block the pr on this... but please be constructive with the discussions, nobody is trying to argue the semantics of single words such as "usage"... i only want to avoid generating code and especially code comments that are based on **assumptions** citing python behavior without proof or with direct reproducers disproving the assumptions.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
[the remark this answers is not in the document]
olichtne: right but... unloading the modules takes 24 seconds and it's done for every subconfiguration...

installing one additional dependency that achieves the same work in a fraction of the time may be worthwhile?

lnst/Devices/RemoteDevice.py
```diff
@@ -188,6 +188,13 @@ def __repr__(self):
 
         return "{}({})".format(self._dev_cls.__name__, ", ".join(args))
 
+    def keep_addrs_on_down(self) -> None:
+        self.netns.run(f"echo 1 > /proc/sys/net/ipv6/conf/{self.name}/keep_addr_on_down")
```
olichtne, line 192: i meant `Device` class, not `RemoteDevice`...

olichtne (changes requested):
some mistakes found for the previously resolved threads.

olichtne (dismissed):
looks good to me now.

lnst/Tests/LongLivedConnections.py
```diff
@@ -0,0 +1,167 @@
...
+import resource
+import threading
+from typing import Union
+
+from .BaseTestModule import BaseTestModule
+from lnst.Common.IpAddress import Ip4Address, Ip6Address, ipaddress
+from lnst.Common.Parameters import IntParam, StrParam
+
+
+def get_af(ip: Union[Ip4Address, Ip6Address]):
```
olichtne, line 14: ok so... this should have been asked earlier but i somehow missed it...

why have this function? the `IpAddress` class already defines a `family` attribute which returns the correct value for ipv4 and ipv6: https://github.com/LNST-project/lnst/blob/master/lnst/Common/IpAddress.py#L31

i assume that this is because `server_ip` and `client_ip` are `StrParam` so you want to do the conversion... but... why are they `StrParam`? shouldn't they just be `IpParam`? https://github.com/LNST-project/lnst/blob/master/lnst/Common/Parameters.py#L91

olichtne (changes requested):

olichtne (approved):
looks good now, we can merge tomorrow if you have the current version tested.

olichtne:
tests provided in `J:9753788` look ok, merging this in.

PR #373, opened 2024-06-26
BaseRESTConfigMixin: retrying of API call

enhaut:
### Description
For various reasons, it may happen that a remote resource is unreachable/doesn't respond with an expected response. Therefore, retry mechanism for API calls tries to send request `tries` times with `1sec` sleep in between.

### Tests
* Beaker job `J:9529957`
* manual tests of retries by changing `/etc/hosts` file to simulate resource is unreachable

### DIFF
--- a/lnst/Recipes/ENRT/ConfigMixins/BaseRESTConfigMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/BaseRESTConfigMixin.py
@@ -1,5 +1,7 @@
+import time
 import logging
 import requests
+from requests.exceptions import RequestException
 
 from lnst.Common.LnstError import LnstError
 from lnst.Common.Parameters import BoolParam, StrParam
@@ -10,6 +12,10 @@ class BaseRESTConfigMixin:
     rest_user = StrParam()
     rest_password = StrParam()
     ssl_verify = BoolParam(default=True, mandatory=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.BASE_DELAY = 1
 
     @staticmethod
     def __get_request_function(method: str):
@@ -28,11 +34,28 @@ class BaseRESTConfigMixin:
 
         return kwargs
 
-    def api_request(self, method: str, endpoint: str, response_code: int = 200, **kwargs) -> bytes:
+    def api_request(self, method: str, endpoint: str, response_code: int = 200, tries: int = 7, **kwargs) -> bytes:
         request = self.__build_request(endpoint, **kwargs)
         req_func = self.__get_request_function(method)
 
-        response = req_func(**request)
+        for i in range(1, tries+1):  # i starts from 1
+            try:
+                response = req_func(**request)
+            except RequestException as e:
+                delay = self.BASE_DELAY * 2 ** i
+
+                logging.error(
+                    f"API request ({i}/{tries}) failed. Retrying in {delay} seconds.",
+                    exc_info=True,
+                )
+
+                time.sleep(delay)
+                continue
+            
+            break  # request was successful
+        else:
+            raise LnstError(f"API request failed after {tries} tries")
+
         if response.status_code != response_code:
             raise LnstError(f"Request failed with status code {response.status_code}")
 


Axonis:
When working with retries, of of the best strategies is to use exponential backoff, with following formula 

> curr_delay = base_delay * (2^num_retries)

Which reduces chance of switch overload and gives it more time to respond to the request.

I'm not 100% sure about this hypothesis, but you could be hitting thundering herd problem (multiple requests hitting switch at the same time) with the switch response, where adding some randomness to the delay could help it as well = could be implemented as jitter sleep basically.

enhaut:
> When working with retries, of of the best strategies is to use exponential backoff, with following formula
> 
> > curr_delay = base_delay * (2^num_retries)
> 
> Which reduces chance of switch overload and gives it more time to respond to the request.

Makes sense. 

> I'm not 100% sure about this hypothesis, but you could be hitting thundering herd problem (multiple requests hitting switch at the same time) with the switch response, where adding some randomness to the delay could help it as well = could be implemented as jitter sleep basically.

That's kinda overkill for us, so I just moved delay calculation to `_calculate_retry_delay` method if anyone wanted to use its own calculation formula.

lnst/Recipes/ENRT/ConfigMixins/BaseRESTConfigMixin.py
```diff
@@ -28,11 +30,27 @@ def __build_request(self, endpoint: str, **kwargs):
...
+
+                time.sleep(self._calculate_retry_delay(i))
+                continue
+            
+            break  # request was successful
+        else:
+            raise LnstError(f"API request failed after {tries} tries")
+
         if response.status_code != response_code:
             raise LnstError(f"Request failed with status code {response.status_code}")
```
Axonis, line 56: To simplify code even further we could combine the try-except with status code check that you do after the block.
Also move base_delay somewhere, that isn't going to reinitialize it each call.

```python
try:
    response = req_func(**request)
    if response.status_code == response_code:
        logging.debug("API response: %s", response.content)
        return response.content
    else:
        logging.error(f"Request failed with status code {response.status_code} (attempt {i}/{tries})")
except RequestException as e:
    logging.error(f"API request (attempt {i}/{tries}) failed: {e}")

    delay = self.BASE_DELAY * 2 ** (i - 1)
    logging.info(f"Retrying in {delay} seconds...")
    time.sleep(delay)
```

olichtne:
is the `BestRESTConfigMixin` name in the commit header and PR header intentional? the Class is called `BaseRESTConfigMixin`... Best vs Base

lnst/Recipes/ENRT/ConfigMixins/BaseRESTConfigMixin.py
olichtne: agree with the delay calculation proposed.

I probably wouldn't include the status code check inside the try-except block as it increases the indentation level and doesn't actually change anything runtime wise.

Unless we also want to do retries also in case the response code isn't the expected value? I'm not sure we want that universally (i can imagine situations when a bad status code means we shouldn't retry...) but it could make sense in some situations?

enhaut: The idea behind retrying the calls is, to handle more or less networking problems. Non-expected http return codes implies problems with request itself, so i don't think it should be retried

olichtne: yeah, that's probably true, agreed. let's just do the delay calculation proposal and we can merge this in.

lnst/Recipes/ENRT/ConfigMixins/BaseRESTConfigMixin.py
```diff
@@ -28,11 +34,25 @@ def __build_request(self, endpoint: str, **kwargs):
...
         req_func = self.__get_request_function(method)
 
-        response = req_func(**request)
+        for i in range(1, tries+1):  # i starts from 1
+            try:
+                response = req_func(**request)
+            except RequestException as e:
+                delay = self.BASE_DELAY * 2 ** i
+
+                logging.error(f"API request ({i}/{tries}) failed: {e}. Retrying in {delay} seconds.")
```
olichtne, line 47: ah, sorry, one more thing. please add a `log_exception_traceback` here... the simple `str(e)` part can sometimes be quite useless but a log of the traceback can help with that.

PR #379, opened 2024-09-26
UseVfsMixin: add vf_trust parameter

jtluka:
### Description

This adds `vf_trust` parameter to UseVfsMixin. The Device class is extended with `vf_trust()` method.

### Tests

~~J:9865704~~ J:9939392

Note that sfc driver tests failed because the device does not support vf_trust. I have a separate patch set that fixes that along with different way to achieve same functionality.

### Reviews
@olichtne

### DIFF
--- a/lnst/Recipes/ENRT/UseVfsMixin.py
+++ b/lnst/Recipes/ENRT/UseVfsMixin.py
@@ -1,4 +1,4 @@
-from lnst.Common.Parameters import BoolParam
+from lnst.Common.Parameters import BoolParam, ChoiceParam
 from lnst.Controller.Requirements import DeviceReq
 from lnst.Recipes.ENRT.SRIOVDevices import SRIOVDevices
 
@@ -12,11 +12,16 @@ class UseVfsMixin:
     with VF Device instances. This allows user to interact with the network
     interfaces without additional changes to the code of recipe.
 
+    Mixin provides two boolean parameters:
+    * use_vfs - main parameter to enable or disable (default) use of VFs
+    * vf_trust - set the trust parameter of the used VFs, 'on' or 'off' (default)
+
     There are some limitations, for example pause frames cannot be configured
     since the VF do not support these.
     """
 
     use_vfs = BoolParam(default=False)
+    vf_trust = ChoiceParam(choices={'on', 'off'}, default='off')
 
     def test_wide_configuration(self):
         config = super().test_wide_configuration()
@@ -33,6 +38,8 @@ class UseVfsMixin:
             for dev_name in dev_names:
                 dev = getattr(host, dev_name)
                 sriov_devices = SRIOVDevices(dev, 1)
+                dev.vf_trust(0, self.params.vf_trust)
+
                 vf_dev = sriov_devices.vfs[0]
                 host.map_device(dev_name, {"ifname": vf_dev.name})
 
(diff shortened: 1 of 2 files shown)


lnst/Recipes/ENRT/UseVfsMixin.py
```diff
@@ -12,11 +12,16 @@ class UseVfsMixin:
     with VF Device instances. This allows user to interact with the network
     interfaces without additional changes to the code of recipe.
 
+    Mixin provides two boolean parameters:
```
jtluka, line 15: As they're not both boolean, I'll update the dosctring.

jtluka:
I noticed that there's a flaw in the implementation:
```
    Testwide configuration for recipe VlansOverBondRecipe description:
    Using vf device enp2s0f0v0 of pf enp2s0f0 for DeviceReq host1.eth0 trusted={self.params.vf_trust}
    Using vf device enp2s0f1v0 of pf enp2s0f1 for DeviceReq host1.eth1 trusted={self.params.vf_trust}
    Using vf device enp2s0f0v0 of pf enp2s0f0 for DeviceReq host2.eth0 trusted={self.params.vf_trust}
```

For this recipe we need to set vf trust only on host1 that configures the bond device. The other host does not use bonding so it is not necessary to configure vf trust.

jtluka:
> I noticed that there's a flaw in the implementation:
> 
> ```
>     Testwide configuration for recipe VlansOverBondRecipe description:
>     Using vf device enp2s0f0v0 of pf enp2s0f0 for DeviceReq host1.eth0 trusted={self.params.vf_trust}
>     Using vf device enp2s0f1v0 of pf enp2s0f1 for DeviceReq host1.eth1 trusted={self.params.vf_trust}
>     Using vf device enp2s0f0v0 of pf enp2s0f0 for DeviceReq host2.eth0 trusted={self.params.vf_trust}
> ```
> 
> For this recipe we need to set vf trust only on host1 that configures the bond device. The other host does not use bonding so it is not necessary to configure vf trust.

Fixed by adding `vf_trust_device_list` property.

lnst/Recipes/ENRT/UseVfsMixin.py
olichtne: the docstring still says "two boolean parameters"

lnst/Recipes/ENRT/UseVfsMixin.py
```diff
@@ -56,7 +63,7 @@ def generate_test_wide_description(self, config):
 
         if self.params.use_vfs:
             description += [
-                f"Using vf device {vf_dev.name} of pf {sriov_devices.phys_dev.name} for DeviceReq {host.hostid}.{vf_dev._id}"
+                f"Using vf device {vf_dev.name} of pf {sriov_devices.phys_dev.name} for DeviceReq {host.hostid}.{vf_dev._id} trusted={self.params.vf_trust}"
```
olichtne, line 66: this directly prints `self.params`... is there maybe also a `getter` for the `Device.vf_trust` value that should be used instead to reflect the real state of the device?

olichtne (commented):
minor questions asked. but overall looks ok

lnst/Recipes/ENRT/UseVfsMixin.py
jtluka: Hmm, so I think I've accidentally originally pushed the code to LNST-project's git tree instead of my fork. Then I pushed the updates into my fork's git, so they're not visible here. I'll fix that.

jtluka: There is no way to change the source branch of PR, I created a new one in https://github.com/LNST-project/lnst/pull/381

jtluka: This is fixed in new PR.

lnst/Recipes/ENRT/UseVfsMixin.py
jtluka: There's no getter atm. The "property" is implemented through a method. Maybe I could change this to property.

PR #381, opened 2024-10-03
UseVfsMixin: add vf_trust parameter

jtluka:
### Description

This replaces original messed up PR https://github.com/LNST-project/lnst/pull/379

This adds vf_trust parameter to UseVfsMixin. The Device class is extended with vf_trust() method.
Tests

This also updates BondRecipe, VlansOverBondRecipe and DoubleBondRecipe with fail_over_mac. To avoid code duplication I refactored these classes to us new BondingMixin class that adds the common bonding recipe parameters and provides a method to create the bonding devices.

J:10034309

Reviews

@olichtne

### DIFF
--- a/lnst/Recipes/ENRT/BondRecipe.py
+++ b/lnst/Recipes/ENRT/BondRecipe.py
@@ -218,3 +218,7 @@ class BondRecipe(BondingMixin, PerfReversibleFlowMixin, CommonHWSubConfigMixin, 
         """
         return [self.matched.host1.eth0, self.matched.host1.eth1,
             self.matched.host2.eth0]
+
+    @property
+    def vf_trust_dev_list(self):
+        return [self.matched.host1.eth0, self.matched.host1.eth1]
--- a/lnst/Recipes/ENRT/BondRecipe.py
+++ b/lnst/Recipes/ENRT/BondRecipe.py
@@ -1,8 +1,6 @@
 from collections.abc import Collection
 from lnst.Common.Parameters import (
     Param,
-    IntParam,
-    StrParam,
     IPv4NetworkParam,
     IPv6NetworkParam,
 )
@@ -18,10 +16,11 @@ from lnst.Recipes.ENRT.ConfigMixins.CommonHWSubConfigMixin import (
     CommonHWSubConfigMixin)
 from lnst.Recipes.ENRT.ConfigMixins.PerfReversibleFlowMixin import (
     PerfReversibleFlowMixin)
+from lnst.Recipes.ENRT.BondingMixin import BondingMixin
 from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
-from lnst.Devices import BondDevice
 
-class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConfigMixin,
+
+class BondRecipe(BondingMixin, PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConfigMixin,
     BaremetalEnrtRecipe):
     """
     This recipe implements Enrt testing for a network scenario that looks
@@ -43,15 +42,8 @@ class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConf
         |            host1            |     |      host2     |
         '-----------------------------'     '----------------'
 
-    The recipe provides additional recipe parameters to configure the bonding
+    Refer to :any:`BondingMixin` for parameters to configure the bonding
     device.
-
-        :param bonding_mode:
-            (mandatory test parameter) the bonding mode to be configured on
-            the bond0 device.
-        :param miimon_value:
-            (mandatory test parameter) the miimon interval to be configured
-            on the bond0 device.
 
     All sub configurations are included via Mixin classes.
 
@@ -73,9 +65,6 @@ class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConf
     net_ipv4 = IPv4NetworkParam(default="192.168.101.0/24")
     net_ipv6 = IPv6NetworkParam(default="fc00::/64")
 
-    bonding_mode = StrParam(mandatory=True)
-    miimon_value = IntParam(mandatory=True)
-
     def test_wide_configuration(self):
         """
         Test wide configuration for this recipe involves creating a bonding
@@ -91,12 +80,14 @@ class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConf
         host1, host2 = self.matched.host1, self.matched.host2
         config = super().test_wide_configuration()
 
-        host1.bond0 = BondDevice(mode=self.params.bonding_mode,
-            miimon=self.params.miimon_value)
-
-        for dev in [host1.eth0, host1.eth1]:
-            dev.down()
-            host1.bond0.slave_add(dev)
+        self.create_bond_devices(
+            config,
+            {
+                "host1": {
+                    "bond0": [host1.eth0, host1.eth1]
+                }
+            }
+        )
 
         ipv4_addr = interface_addresses(self.params.net_ipv4)
         ipv6_addr = interface_addresses(self.params.net_ipv6)
@@ -113,12 +104,6 @@ class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConf
         return config
 
     def generate_test_wide_description(self, config: EnrtConfiguration):
-        """
-        Test wide description is extended with the configured IP addresses, the
-        configured bonding slave interfaces, bonding mode and the miimon interval.
-        """
-
-        host1, host2 = self.matched.host1, self.matched.host2
         desc = super().generate_test_wide_description(config)
         desc += [
             "\n".join([
@@ -126,21 +111,9 @@ class BondRecipe(PerfReversibleFlowMixin, CommonHWSubConfigMixin, OffloadSubConf
                     dev.host.hostid, dev.name, dev.ips
                 )
                 for dev in config.configured_devices
-            ]),
-            "Configured {}.{}.slaves = {}".format(
-                host1.hostid, host1.bond0.name,
-                ['.'.join([host1.hostid, slave.name])
-                for slave in host1.bond0.slaves]
-            ),
-            "Configured {}.{}.mode = {}".format(
-                host1.hostid, host1.bond0.name,
-                host1.bond0.mode
-            ),
-            "Configured {}.{}.miimon = {}".format(
-                host1.hostid, host1.bond0.name,
-                host1.bond0.miimon
-            )
+            ])
         ]
+
         return desc
 
     def generate_ping_endpoints(self, config):
--- a/lnst/Recipes/ENRT/BondingMixin.py
+++ b/lnst/Recipes/ENRT/BondingMixin.py
@@ -1,0 +1,124 @@
+from lnst.Common.Parameters import (
+    IntParam,
+    StrParam,
+    ChoiceParam,
+)
+from lnst.Devices import RemoteDevice, BondDevice
+from lnst.Controller.Recipe import RecipeError
+from lnst.Recipes.ENRT.BaseEnrtRecipe import EnrtConfiguration
+
+
+class BondingMixin:
+    """
+    The recipe mixin provides additional recipe parameters to configure the
+    bonding device.
+
+        :param bonding_mode:
+            (mandatory test parameter) the bonding mode to be configured on
+            the bond0 device.
+        :param miimon_value:
+            (mandatory test parameter) the miimon interval to be configured
+            on the bond0 device.
+        :param fail_over_mac:
+            the fail_over_mac mode to be configured on the bond0 device.
+    """
+    bonding_mode = StrParam(mandatory=True)
+    miimon_value = IntParam(mandatory=True)
+    fail_over_mac = ChoiceParam(
+        StrParam, choices={"none", "active", "follow"}, default="none"
+    )
+
+    def create_bond_devices(
+        self,
+        config: EnrtConfiguration,
+        bond_devices_specs: dict[str: dict[str: list[RemoteDevice]]] = {}
+    ) -> None:
+        """
+        The derived class should call:
+
+        .. code-block:: none
+
+            config = super().test_wide_configuration()
+            self.create_bond_devices(
+                config,
+                {
+                    "host1": {
+                        "bond0": [host1.nic1, host1.nic2]
+                    },
+                    "host2": {
+                        "bond0": [host2.nic1, host2.nic2]
+                    }
+                }
+            )
+
+        That would create bonding devices accessible by `host1.bond0` and `host2.bond0`.
+
+        `host1.bond0` would be created with ports `host1.nic1` and `host1.nic2` and
+        `host2.bond0` with ports `host2.nic1` and `host2.nic2`.
+        """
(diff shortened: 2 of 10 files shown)


jtluka:
VF status can be fetched using:
```
[root@wsfd-advnetlab197 ~]# cat /sys/class/net/eno12399/device/sriov_numvfs 
0
[root@wsfd-advnetlab197 ~]# echo 3 > /sys/class/net/eno12399/device/sriov_numvfs 
[root@wsfd-advnetlab197 ~]# cat /sys/class/net/eno12399/device/sriov_numvfs 
3
[root@wsfd-advnetlab197 ~]# ip l show dev eno12399
4: eno12399: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether b4:83:51:01:ea:c4 brd ff:ff:ff:ff:ff:ff
    vf 0     link/ether d2:a3:e7:73:a5:f6 brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
    vf 1     link/ether 1e:53:d7:e2:e2:fd brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
    vf 2     link/ether c2:52:5e:b9:0d:a9 brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
    altname enp49s0f0

```

olichtne:
> VF status can be fetched using:
> 
> ```
> [root@wsfd-advnetlab197 ~]# cat /sys/class/net/eno12399/device/sriov_numvfs 
> 0
> [root@wsfd-advnetlab197 ~]# echo 3 > /sys/class/net/eno12399/device/sriov_numvfs 
> [root@wsfd-advnetlab197 ~]# cat /sys/class/net/eno12399/device/sriov_numvfs 
> 3
> [root@wsfd-advnetlab197 ~]# ip l show dev eno12399
> 4: eno12399: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
>     link/ether b4:83:51:01:ea:c4 brd ff:ff:ff:ff:ff:ff
>     vf 0     link/ether d2:a3:e7:73:a5:f6 brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
>     vf 1     link/ether 1e:53:d7:e2:e2:fd brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
>     vf 2     link/ether c2:52:5e:b9:0d:a9 brd ff:ff:ff:ff:ff:ff, spoof checking on, link-state auto, trust off
>     altname enp49s0f0
> ```

can you also try `ip --json l show dev eno12399` to see if this can easily be machine parsed? or if this is somehow accessible from the netlink messages that we already have?

lnst/Recipes/ENRT/UseVfsMixin.py
```diff
@@ -12,19 +13,24 @@ class UseVfsMixin:
...
+    vf_trust = ChoiceParam(choices={'on', 'off'})
 
     def test_wide_configuration(self):
         config = super().test_wide_configuration()
 
         if not self.params.use_vfs:
             return config
 
-        config.vf_config = {}
+        self.vf_config = {}
```
olichtne, line 33: why the change from `config` to `self`? i would imagine that we may want to export this value into our measurement result metadata to discern between the two possibilities and in that case it should probably be in the shared config object

jtluka:
> ip --json l show dev eno12399

Great idea, it works:
```python
[{"ifindex":4,"ifname":"eno12399","flags":["BROADCAST","MULTICAST"],"mtu":1500,"qdisc":"noop","operstate":"DOWN","linkmode":"DEFAULT","group":"default","txqlen":1000,"link_type":"ether","addr
ess":"b4:83:51:01:ea:c4","broadcast":"ff:ff:ff:ff:ff:ff","vfinfo_list":[{"vf":0,"link_type":"ether","address":"d2:a3:e7:73:a5:f6","broadcast":"ff:ff:ff:ff:ff:ff","vlan_list":[{}],"rate":{"max
_tx":0,"min_tx":0},"spoofchk":true,"link_state":"auto","trust":false},{"vf":1,"link_type":"ether","address":"1e:53:d7:e2:e2:fd","broadcast":"ff:ff:ff:ff:ff:ff","vlan_list":[{}],"rate":{"max_t
x":0,"min_tx":0},"spoofchk":true,"link_state":"auto","trust":false},{"vf":2,"link_type":"ether","address":"c2:52:5e:b9:0d:a9","broadcast":"ff:ff:ff:ff:ff:ff","vlan_list":[{}],"rate":{"max_tx"
:0,"min_tx":0},"spoofchk":true,"link_state":"auto","trust":false}],"altnames":["enp49s0f0"]}]
```

lnst/Recipes/ENRT/UseVfsMixin.py
```diff
@@ -33,17 +39,22 @@ def test_wide_configuration(self):
...
                 sriov_devices = SRIOVDevices(dev, 1)
+
                 vf_dev = sriov_devices.vfs[0]
                 host.map_device(dev_name, {"ifname": vf_dev.name})
 
-                host_config = config.vf_config.setdefault(host, [])
+                host_config = self.vf_config.setdefault(host, [])
                 host_config.append(sriov_devices)
 
+        if self.params.get("vf_trust"):
```
olichtne, line 49: the `vf_trust` parameter will always have either an `"on"` or `"off"` value so this condition will always be True... maybe you wanted to check `if self.params.use_vfs` but that's also not necessary since you check that higher in the method.

lnst/Recipes/ENRT/UseVfsMixin.py
```diff
@@ -56,10 +67,14 @@ def generate_test_wide_description(self, config):
 
         if self.params.use_vfs:
             description += [
-                f"Using vf device {vf_dev.name} of pf {sriov_devices.phys_dev.name} for DeviceReq {host.hostid}.{vf_dev._id}"
-                for host, sriov_devices_list in config.vf_config.items()
+                f"Using vf device {vf_dev.name} of pf {sriov_devices.phys_dev.name} for DeviceReq {host.hostid}.{vf_dev._id}" + (f" trusted={self.params.vf_trust}" if self.params.get("vf_trust") and sriov_devices.phys_dev in self.vf_trust_device_list else "")
```
olichtne, line 70: same here... the vf_trust parameter always has a value.

lnst/Recipes/ENRT/BondRecipe.py
```diff
@@ -245,3 +246,7 @@ def pause_frames_dev_list(self):
         """
         return [self.matched.host1.eth0, self.matched.host1.eth1,
             self.matched.host2.eth0]
+
+    @property
+    def vf_trust_device_list(self) -> list[RemoteDevice]:
+        return [sriov_devices.phys_dev for sriov_devices in self.vf_config[self.matched.host1]]
```
olichtne, line 252: i think in this case the `sriov_devices` is a `list[Device]` and the `phys_dev` is a `Device` property so you need to nest this one more level:
```
return [device.phys_dev for devices in self.vf_config[self.matched.host1] for device in devices]
```
similar to other recipes...

and probably adjust the variable naming so it's consistent

lnst/Recipes/ENRT/DoubleBondRecipe.py
```diff
@@ -118,3 +119,7 @@ def dev_interrupt_hw_config_dev_list(self):
     def parallel_stream_qdisc_hw_config_dev_list(self):
         host1, host2 = self.matched.host1, self.matched.host2
         return [host1.eth0, host1.eth1, host2.eth0, host2.eth1]
+
+    @property
+    def vf_trust_device_list(self) -> list[RemoteDevice]:
+        return [sriov_devices.phys_dev for sriov_devices_list in self.vf_config.values() for sriov_devices in sriov_devices_list]
```
olichtne, line 125: the nested loop returns a single device i think so probably better to name the variable `sriov_device` instead of `sriov_devices` right?

lnst/Recipes/ENRT/VlansOverBondRecipe.py
```diff
@@ -309,3 +310,7 @@ def pause_frames_dev_list(self):
         """
         host1, host2 = self.matched.host1, self.matched.host2
         return [host1.eth0, host1.eth1, host2.eth0]
+
+    @property
+    def vf_trust_device_list(self) -> list[RemoteDevice]:
+        return [sriov_devices.phys_dev for sriov_devices in self.vf_config[self.matched.host1]]
```
olichtne, line 316: same problem with list of devices here.

olichtne (changes requested):

lnst/Recipes/ENRT/UseVfsMixin.py
jtluka: I changed this to be able to access the vf_config in `test_wide_description()`. I can add this to both `config.vf_config()` and `self.vf_config()`

jtluka: Err, and when I checked the code now I see the config is passed as argument. I'll double check that, most likely you have a valid point.

lnst/Recipes/ENRT/UseVfsMixin.py
olichtne: ah ok, it doesn't have a default, and parameters are optional by default... ok

lnst/Recipes/ENRT/UseVfsMixin.py
jtluka: I changed this because I need this in https://github.com/LNST-project/lnst/blob/1d72d11f371312ac835c373b1d539e4757bf32eb/lnst/Recipes/ENRT/BondRecipe.py#L252.

To have the device list as property I need to store this in recipe object directly.

lnst/Recipes/ENRT/BondRecipe.py
jtluka: No, this is not necessary. SriovDevices is a collection of VFs over **single** PF. So iterating over devices in sriov_devices would generate several PFs that are identical.

jtluka: And `SriovDevices` class exports the `phys_dev` property that is common to all VFs in the class.

lnst/Recipes/ENRT/DoubleBondRecipe.py
jtluka: The class name is `SRIOVDevices`, so I followed this pattern. I thought of SRIOVDevices class as a collection of VFs over a PF.

The `self.vf_config.values()` returns `list[list[SRIOVDevices]]`.

olichtne: > The self.vf_config.values() returns list[list[SRIOVDevices]].

this is why i'm confused about this `vf_trust_device_list` property, in some recipes you have two nested loops and in some just one loop...

lnst/Recipes/ENRT/UseVfsMixin.py
jtluka: Latest version does not require this anymore, so I removed this change.

lnst/Recipes/ENRT/DoubleBondRecipe.py
jtluka: I changed this so that the individual recipes return list of the vf_devices (the remapped device requirements) and the UseVfsMixin that has access to SRIOVDevices structures would gather the PF devices.

lnst/Recipes/ENRT/VlansOverBondRecipe.py
jtluka: Updated per https://github.com/LNST-project/lnst/pull/381#discussion_r1805965954

lnst/Recipes/ENRT/BondRecipe.py
jtluka: Updated per https://github.com/LNST-project/lnst/pull/381#discussion_r1805965954

olichtne (approved):
just a minor nitpick, otherwise approved

lnst/Recipes/ENRT/BondingMixin.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+
+        if self.params.bonding_mode in ["active-backup", "1"]:
+            device_params["fail_over_mac"] = self.params.fail_over_mac
+
+        for host_str, spec in bond_devices_specs.items():
+            for bond_dev_name, bonded_devices in spec.items():
+                if len({dev.host for dev in bonded_devices}) > 1:
+                    raise RecipeError(
+                        f"Cannot create bond device with ports coming from different hosts, {bonded_devices}"
```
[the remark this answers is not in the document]
olichtne: the condition you use counts the number of hosts in "bonded_devices", if there's 2 then yes... the devices are from two different hosts and that doesn't work...

however, you can also provide 2 devices from one host, which is different than the host where the bond device should be created - your condition will count 1, but will pass even when it should fail.

for example:
```
        self.create_bond_devices(
            {
                "host1": {
                    "bond0": [host2.nic1, host2.nic2]
                },
            }
```

nics are from host2, but bond on host1.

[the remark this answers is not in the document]
jtluka: I'm not referring to the host_str in the spec. I'm just checking if all devices in the bond_devices list are on the same host.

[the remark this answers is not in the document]
jtluka: I could add your condition with a different exception string if you want.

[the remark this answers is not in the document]
olichtne: > if all devices in the bond_devices list are on the same host.

yes... but it could be the WRONG host. I know you're not checking `host_str`, i'm saying maybe we should?

but yeah... it's a nitpick like i've said, we can skip this if you want.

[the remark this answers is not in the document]
jtluka: You're right that with your proposal that would cover both cases. I'll do some testing and eventually update this.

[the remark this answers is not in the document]
jtluka: Fixed, I've also added check if the `host_str` can be resolved in `self.matched.hosts`

olichtne (dismissed):

lnst/Recipes/ENRT/BondingMixin.py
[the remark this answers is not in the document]
olichtne: looks ok

olichtne:
add doc rst file entries so that the new classes are exposed on readthedocs properly

lnst/Recipes/ENRT/VlansOverBondRecipe.py
```diff
@@ -200,6 +192,12 @@ def generate_test_wide_description(self, config: EnrtConfiguration):
                 host1.bond0.miimon
             )
         ]
+
+        if self.params.bonding_mode in ["active-backup", "1"]:
+            desc += ["Configured {}.{}.fail_over_mac = {}".format(
+                host1.hostid, host1.bond0.name,
+                host1.bond0.fail_over_mac
+            )]
```
jtluka, line 200: I think this shoud go into  the BondingMixin class. It's duplicated over individual recipes and it makes more sense to generate this within the BondingMixin class.

jtluka: Fixed.

jtluka:
Last test for fail_over_mac setting : J:10057391

jtluka:
> Last test for fail_over_mac setting : J:10057391

Tests passed.

PR #382, opened 2024-10-03
Fix iperf no connection and PerfResult warmup timestamp calculation

olichtne:
### Description
This should fix the issues we regularly see with lnst crashing with exception when iperf failed to establish a connection. The source of the issue is that the created perf result data structure was slightly different.

While testing this I also found issues with the warmup/warmdown timestamp calculation because the zero created zero iperf measurements don't have a duration long enough to include a warmup/warmdown... this however shouldn't crash the trimming and should instead return another zero measurement.

### Tests

### Reviews
@jtluka @enhaut @Axonis 

Closes: #

### DIFF
--- a/lnst/RecipeCommon/Perf/Measurements/NeperFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/NeperFlowMeasurement.py
@@ -200,8 +200,8 @@ class NeperFlowMeasurement(BaseFlowMeasurement):
         cpu_results = SequentialPerfResult()
 
         if not job.passed:
-            results.append(PerfInterval(0, 0, "transactions", time.time()))
-            cpu_results.append(PerfInterval(0, 0, "cpu_percent", time.time()))
+            results.append(PerfInterval(0, 1, "transactions", time.time()))
+            cpu_results.append(PerfInterval(0, 1, "cpu_percent", time.time()))
         elif job.what.is_crr_server():
             # Neper doesn't support server stats for tcp_crr due to memory issues.
             # Use perf_interval of 0.
--- a/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
@@ -84,33 +84,11 @@ class FlowMeasurementResults(BaseMeasurementResults):
 
     @property
     def warmup_end(self):
-        if self.warmup_duration == 0:
-            return self.start_timestamp
-
-        return max(
-            [
-                parallel[self.warmup_duration - 1].end_timestamp
-                for parallel in (
-                    *self.generator_results,
-                    *self.receiver_results,
-                )
-            ]
-        )
+        return self.start_timestamp+self.warmup_duration
 
     @property
     def warmdown_start(self):
-        if self.warmup_duration == 0:
-            return self.end_timestamp
-
-        return min(
-            [
-                parallel[-self.warmup_duration].start_timestamp
-                for parallel in (
-                    *self.generator_results,
-                    *self.receiver_results,
-                )
-            ]
-        )
+        return self.end_timestamp-self.warmup_duration
 
     def time_slice(self, start, end):
         result_copy = FlowMeasurementResults(
--- a/lnst/RecipeCommon/Perf/Measurements/Results/NeperFlowMeasurementResults.py
+++ b/lnst/RecipeCommon/Perf/Measurements/Results/NeperFlowMeasurementResults.py
@@ -12,7 +12,7 @@ class NeperFlowMeasurementResults(FlowMeasurementResults):
 
         return max(
             [
-                parallel[self.warmup_duration - 1].end_timestamp
+                parallel.start_timestamp + self.warmup_duration
                 for parallel in self.generator_results
             ]
         )
@@ -27,7 +27,7 @@ class NeperFlowMeasurementResults(FlowMeasurementResults):
 
         return min(
             [
-                parallel[-self.warmup_duration].start_timestamp
+                parallel.end_timestamp - self.warmup_duration
                 for parallel in self.generator_results
             ]
         )
--- a/lnst/RecipeCommon/Perf/Measurements/IperfFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/IperfFlowMeasurement.py
@@ -209,7 +209,7 @@ class IperfFlowMeasurement(BaseFlowMeasurement):
     def _parse_job_streams(self, job):
         result = ParallelPerfResult()
         if not job.passed:
-            result.append(PerfInterval(0, 0, "bits", time.time()))
+            result.append(SequentialPerfResult([PerfInterval(0, 1, "bits", time.time())]))
         else:
             for i in job.result["data"]["end"]["streams"]:
                 result.append(SequentialPerfResult())
--- a/lnst/RecipeCommon/Perf/Measurements/IperfFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/IperfFlowMeasurement.py
@@ -225,7 +225,7 @@ class IperfFlowMeasurement(BaseFlowMeasurement):
 
     def _parse_job_cpu(self, job):
         if not job.passed:
-            return PerfInterval(0, 0, "cpu_percent", time.time())
+            return PerfInterval(0, 1, "cpu_percent", time.time())
         else:
             cpu_percent = job.result["data"]["end"]["cpu_utilization_percent"]["host_total"]
             job_start = job.result["data"]["start"]["timestamp"]["timesecs"]
--- a/lnst/RecipeCommon/Perf/Measurements/TRexFlowMeasurement.py
+++ b/lnst/RecipeCommon/Perf/Measurements/TRexFlowMeasurement.py
@@ -173,10 +173,10 @@ class TRexFlowMeasurement(BaseFlowMeasurement):
 
         if not job.passed:
             timestamp = time.time()
-            results.generator_results.append(PerfInterval(0, 0, "packets", timestamp))
-            results.generator_cpu_stats.append(PerfInterval(0, 0, "cpu_percent", timestamp))
-            results.receiver_results.append(PerfInterval(0, 0, "packets", timestamp))
-            results.receiver_cpu_stats.append(PerfInterval(0, 0, "cpu_percent", timestamp))
+            results.generator_results.append(PerfInterval(0, 1, "packets", timestamp))
+            results.generator_cpu_stats.append(PerfInterval(0, 1, "cpu_percent", timestamp))
+            results.receiver_results.append(PerfInterval(0, 1, "packets", timestamp))
+            results.receiver_cpu_stats.append(PerfInterval(0, 1, "cpu_percent", timestamp))
         else:
             prev_time = job.result["start_time"]
             prev_tx_val = 0


olichtne:
this will need some testing on real data to see if the warmup/warmdown trimming isn't causing issues...

I tested this locally by printing the difference between the old calculation and the new one and the difference was minimal - `0.002s` so IMO... this shouldn't be a problem, but we want to be sure...

lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
```diff
@@ -84,33 +84,11 @@ def end_timestamp(self):
...
-        return max(
-            [
-                parallel[self.warmup_duration - 1].end_timestamp
-                for parallel in (
-                    *self.generator_results,
-                    *self.receiver_results,
-                )
-            ]
-        )
+        return self.start_timestamp+self.warmup_duration
```
jtluka, line 87: I think this is not correct.

The original code was basically returing the end_timestamp of the 3rd sample of the results of the **last started** measurement.

Now you return the start_timestamp (whatever) of the **first started** measurement. The `self.start_timestamp` returns `min(measurements)`, so it may be way off what we expect I think.

jtluka (changes requested):
Besides my concern about the warmup/warmdown timestamps, this looks ok.

jtluka:
> this will need some testing on real data to see if the warmup/warmdown trimming isn't causing issues...
> 
> I tested this locally by printing the difference between the old calculation and the new one and the difference was minimal - `0.002s` so IMO... this shouldn't be a problem, but we want to be sure...
> 
> the original implementation comes from #248 so that will be relevant for testing this

Did you test this on any parallel iperf test we have internally? Eg. 35469f23-2f89-4b8a-a40d-d29395f71fdf (96 parallel flows)

lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
olichtne: i wasn't trying to keep the old logic, IMO the old logic is the incorrect one because it uses the `warmup duration` (which is in seconds) as an index into an "arbitrary" list... which may or may not be samples (that may or may not be 1 second long)

this is why i think this should be fixed...

but yes, this changes the logic and the new results may be off from the old ones... I don't think it will be common, i think for most cases the difference will be minimal. But we definitely need to check

olichtne:
> > this will need some testing on real data to see if the warmup/warmdown trimming isn't causing issues...
> > I tested this locally by printing the difference between the old calculation and the new one and the difference was minimal - `0.002s` so IMO... this shouldn't be a problem, but we want to be sure...
> > the original implementation comes from #248 so that will be relevant for testing this
> 
> Did you test this on any parallel iperf test we have internally? Eg. 35469f23-2f89-4b8a-a40d-d29395f71fdf (96 parallel flows)

this is an arm test which i can't run at the moment...

instead i'll schedule a job with all tests that have `perf_parallel_processes > 1` on RHEL9 that are in `kernel-full` group and see what happens...

olichtne:
test job scheduled: `Submitted: ['J:9976969']`

lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
jtluka: Well consider this situation of parallel results:
```
                    samples

    measurement1    x   x   x   x   x   x   x   x
    measurement2      x   x   x   x   x   x   x   x
    measurement3        x   x   x   x   x   x   x
    measurement4          x   x   x   x   x   x   x

    time (s) -->    1   2   3   4   5   6   7
                    |        |
                    '--------'
                        |
                      warmup
```

With your implementation, the measurement4 would not be completely warmed up,
as we would take the measurement1.start_timestamp as the basis for the addition.

All I'm saying is that we can't use `self.start_timestamp` but most likely
`max([res.start_timestamp) for res in self.generator_results + self.receiver_results + ...])`.

olichtne:
> The reason was that we originally wanted to completely remove warm{up,down} samples

that doesn't work either... the samples are only APPROXIMATELY 1s long even if we set sampling at exactly 1s... so this assumption is simply wrong.

> Your approach could possibly include part of warm{up,down} sample in sliced results.

so could the old one

>  But I believe, that should be fine as no other parts of LNST expects results to have the exact number of samples.

we do additional `alignment` of measurements for multiple measurements (cpu vs flow measurement) as well to align everything properly... and that will also create "partial" samples... So actually... working with an EXACT number of samples is incorrect everywhere in the code that we currently have.

Besides... even if we do a 60s measurement with 1s samples we can still end up with 61 samples... even when doing no warmup/alignment etc...

lnst/RecipeCommon/Perf/Measurements/NeperFlowMeasurement.py
```diff
@@ -200,8 +200,8 @@ def _parse_job_samples(self, job: Job) ->\
         cpu_results = SequentialPerfResult()
 
         if not job.passed:
-            results.append(PerfInterval(0, 0, "transactions", time.time()))
-            cpu_results.append(PerfInterval(0, 0, "cpu_percent", time.time()))
+            results.append(PerfInterval(0, 1, "transactions", time.time()))
+            cpu_results.append(PerfInterval(0, 1, "cpu_percent", time.time()))
```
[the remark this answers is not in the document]
olichtne: hmmm, i haven't, interesting idea, i'll think about it a bit...

lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py
olichtne: The reason was that we originally wanted to completely remove warm{up,down} samples

that doesn't work either... the samples are only APPROXIMATELY 1s long even if we set sampling at exactly 1s... so this assumption is simply wrong.

    Your approach could possibly include part of warm{up,down} sample in sliced results.

so could the old one

    But I believe, that should be fine as no other parts of LNST expects results to have the exact number of samples.

we do additional alignment of measurements for multiple measurements (cpu vs flow measurement) as well to align everything properly... and that will also create "partial" samples... So actually... working with an EXACT number of samples is incorrect everywhere in the code that we currently have.

Besides... even if we do a 60s measurement with 1s samples we can still end up with 61 samples... even when doing no warmup/alignment etc...

olichtne: > Well consider this situation of parallel results:
> 
> ```
>                     samples
> 
>     measurement1    x   x   x   x   x   x   x   x
>     measurement2      x   x   x   x   x   x   x   x
>     measurement3        x   x   x   x   x   x   x
>     measurement4          x   x   x   x   x   x   x
> 
>     time (s) -->    1   2   3   4   5   6   7
>                     |        |
>                     '--------'
>                         |
>                       warmup
> ```
> 
> With your implementation, the measurement4 would not be completely warmed up, as we would take the measurement1.start_timestamp as the basis for the addition.
> 
> All I'm saying is that we can't use `self.start_timestamp` but most likely `max([res.start_timestamp) for res in self.generator_results + self.receiver_results + ...])`.

the diagram you created is almost correct and i aware of this situation... however the old code doesn't work correctly either due to bad assumptions and also simply crashes when the measurement hierarchy is created slightly differently

the point i am making is that the difference in start times in your diagram is IMO greatly exaggerated... when i tried this locally with 5 streams  the difference between the new calculation and the old calculation was... 0.002s which imo is simply not worth it for the code complexity. I have scheduled the tests with the large numbers of streams to see what really happens.

Where your diagram ISN'T correct is that you're writing `measurement 1...2...3...4` but this warmup warmdown is localized to a single measurement but different measurement "subparts": generator vs receiver and multiple streams... for this it should be almost irrelevant because this will be the same Iperf test module anyway...

For parallel processes the single measurement will run multiple iperf test modules but these are prepared and started in a way that should hopefully minimize delay between them... but i do want to see what we get from the scheduled beaker job.

jtluka: Yeah, I know I've exagerrated. This is solely for the purpose of flaw in the implementation.

We're considering only iperf/neper flow measurements - what if we add another where the startups are somewhat lazier?

There's (I believe) minimal complexity.

My proposal is simply:
```
        return max([res.start_timestamp for res inself.generator_results + self.receiver_results +...]) + self.warmup_duration
```

Whereas your implementation basically returns:
```
        return min([res.start_timestamp for res inself.generator_results + self.receiver_results +...]) + self.warmup_duration
```

Because `self.start_timestamp` is exactly the same except the it's `min()`: https://github.com/LNST-project/lnst/blob/de79872547e204f3cc4b91b496dfbf0fd923f60d/lnst/RecipeCommon/Perf/Measurements/Results/FlowMeasurementResults.py#L64

jtluka (dismissed):

olichtne:
scheduled a test for the other solution to check the difference `J:10033878`

olichtne:
Compared the "Option A" test run vs "Option B" test run here: 
[a_vs_b.txt](https://github.com/user-attachments/files/17434879/a_vs_b.txt)

There's a couple of tests where there are different results seen however IMO these are mostly due to the instable nature of multistream performance tests that we usually see anyway.

Both test runs evaluates just OK when compared to historical baselines.

So IMO both implementations are basically equivalent.

As one more step, I'll run a multistream test with 64 threads locally and print out the calculation of the `warmup_end` calculation to compare the option a, option b and the old original implementation to show here

olichtne:
So local comparison with running a test in containers is here.

The warmup_end code i used:
```
    @property
    def warmup_end(self):
        old = max(
            [
                parallel[self.warmup_duration - 1].end_timestamp
                for parallel in (
                    *self.generator_results,
                    *self.receiver_results,
                )
            ])
        optionA = self.start_timestamp + self.warmup_duration
        optionB = max(
                [
                    self.generator_results.start_timestamp,
                    self.generator_cpu_stats.start_timestamp,
                    self.receiver_results.start_timestamp,
                    self.receiver_cpu_stats.start_timestamp,
                ]
            ) + self.warmup_duration

        import logging
        logging.error(f"AAAAAAAAAAA old: {old}")
        logging.error(f"AAAAAAAAAAA A: {optionA}")
        logging.error(f"AAAAAAAAAAA B: {optionB}")
        return (
            max(
                [
                    self.generator_results.start_timestamp,
                    self.generator_cpu_stats.start_timestamp,
                    self.receiver_results.start_timestamp,
                    self.receiver_cpu_stats.start_timestamp,
                ]
            )
            + self.warmup_duration
        )
```

the output i got:
```
-> % rg 'AAAAAAAAAAA old' -A2 Logs
Logs/2024-10-18_14:00:06/SimpleNetworkRecipe_match_0/info
1465:2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA old: 1729252874.167374
1466-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA A: 1729252874
1467-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA B: 1729252874
1468:2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA old: 1729252971.003915
1469-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA A: 1729252971
1470-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA B: 1729252971
1471:2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA old: 1729253068.265373
1472-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA A: 1729253068
1473-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA B: 1729253068
1474:2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA old: 1729253163.374155
1475-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA A: 1729253163
1476-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA B: 1729253163
1477:2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA old: 1729253262.3002691
1478-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA A: 1729253262
1479-2024-10-18 14:08:48       (localhost)        -   ERROR: AAAAAAAAAAA B: 1729253262
```

We can see that the optionA and optionB are identical and they're always basically at most `0.3s` off of the old method.

The recipe i used for this locally was with this configuration:
```
recipe = SimpleNetworkRecipe(
    perf_parallel_processes=64,
    offload_combinations=[{'gro': 'on', 'gso': 'on', 'tso': 'on', 'tx': 'on', 'rx': 'on'}],
    perf_duration=60,
    ip_versions=['ipv4'],
    perf_tests=['tcp_stream'],
    perf_msg_sizes=[16384],
    minimal_idlestates_latency=0,
    drop_caches=True,
    perf_warmup_duration=3,
    disable_turboboost=True,
)
```

And it is worth noting that the startup of the iperf clients took a while as the difference between the first and last client is sizable:
```
2024-10-18 14:07:09       (localhost)        -   DEBUG: Host host2 executing job 271: type(module), host(host2), netns(None), IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12000)
2024-10-18 14:07:09       (localhost)        -   DEBUG: Job.what = IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12000)
2024-10-18 14:07:09       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host2), netns(None), IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12000)

...

2024-10-18 14:07:14       (localhost)        -   DEBUG: Host host2 executing job 334: type(module), host(host2), netns(None), IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12063)
2024-10-18 14:07:14       (localhost)        -   DEBUG: Job.what = IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12063)
2024-10-18 14:07:14       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host2), netns(None), IperfServer(bind=Ip4Address(192.168.101.2/24), oneoff=True, port=12063)
2024-10-18 14:07:14           (host2)        -   DEBUG: Running job 334 with pid "1010"
2024-10-18 14:07:14           (host2)        -   DEBUG: compiled command:  iperf3 -s -B 192.168.101.2 -J -p 12063 -1
2024-10-18 14:07:14           (host2)        -   DEBUG: running as server ...
2024-10-18 14:07:16       (localhost)        -   DEBUG: Host host1 executing job 272: type(module), host(host1), netns(None), IperfClient(server=Ip4Address(192.168.101.2/24), duration=60, warmup_duration=3, blksize=16384, port=12000, clien
t_port=12000)
2024-10-18 14:07:16       (localhost)        -   DEBUG: Job.what = IperfClient(server=Ip4Address(192.168.101.2/24), duration=60, warmup_duration=3, blksize=16384, port=12000, client_port=12000)
2024-10-18 14:07:16       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host1), netns(None), IperfClient(server=Ip4Address(192.168.101.2/24), duration=60, warmup_duration=3, blksize=16384, port=12000, cl

...

2024-10-18 14:07:39       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host1), netns(None), IperfClient(server=Ip4Address(192.168.101.2/24), duration=60, warmup_duration=3, blksize=16384, port=12063, cl
ient_port=12063)
2024-10-18 14:07:39           (host1)        -   DEBUG: Running job 335 with pid "1013"
2024-10-18 14:07:39       (localhost)        -   DEBUG: Waiting for Job 272 on Host host1 for 71 seconds.
2024-10-18 14:07:39           (host1)        -   DEBUG: Measuring for 66 seconds (perf_duration + perf_warmup_duration * 2).
2024-10-18 14:07:39           (host1)        -   DEBUG: compiled command:  iperf3 -c 192.168.101.2 -b 0/1000 -J -t 66   -l 16384  -p 12063 --cport 12063
2024-10-18 14:07:39           (host1)        -   DEBUG: running as client ...
```

olichtne:
Did the same with the `warmdown_start` for completion:

```
    @property
    def warmdown_start(self):
        old = min(
            [
                parallel[-self.warmup_duration].start_timestamp
                for parallel in (
                    *self.generator_results,
                    *self.receiver_results,
                )
            ])
        optionA = self.end_timestamp - self.warmup_duration
        optionB = min(
                [
                    self.generator_results.end_timestamp,
                    self.generator_cpu_stats.end_timestamp,
                    self.receiver_results.end_timestamp,
                    self.receiver_cpu_stats.end_timestamp,
               max(  ]
            ) - self.warmup_duration

        import logging
        logging.error(f"BBBBBBBBBBB old: {old}")
        logging.error(f"BBBBBBBBBBB A: {optionA}")
        logging.error(f"BBBBBBBBBBB B: {optionB}")
        return (
            min(
                [
                    self.generator_results.end_timestamp,
                    self.generator_cpu_stats.end_timestamp,
                    self.receiver_results.end_timestamp,
                    self.receiver_cpu_stats.end_timestamp,
                ]
            )
            - self.warmup_duration
        )

```

logs:
```
Logs/2024-10-18_14:30:00/SimpleNetworkRecipe_match_0/debug
6902:2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB old: 1729254727.004715
6903-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB A: 1729254727.135558
6904-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB B: 1729254727
--
6909:2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB old: 1729254829.009228
6910-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB A: 1729254829.005448
6911-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB B: 1729254829
--
6916:2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB old: 1729254930.013833
6917-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB A: 1729254930.123065
6918-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB B: 1729254930
--
6923:2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB old: 1729255027.000226
6924-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB A: 1729255027.104802
6925-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB B: 1729255027
--
6930:2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB old: 1729255120.022853
6931-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB A: 1729255120.036433
6932-2024-10-18 14:38:45       (localhost)        -   ERROR: BBBBBBBBBBB B: 1729255120
```

here we see some small difference between A and B, but when comparing A to old or B to old imo the difference is irrelevant.

olichtne:
with the new data and testing i believe the simpler option A implementation is better and there should be no real impact on any of our testing.

lnst/RecipeCommon/Perf/Measurements/NeperFlowMeasurement.py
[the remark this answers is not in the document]
olichtne: resolving this.

i'm still not decided on whether or not it could potentially be useful to create an "invalid interval" with 0 duration...

but either way, it imo doesn't make sense to include it in this PR anyway.

lnst/RecipeCommon/Perf/Measurements/Results/NeperFlowMeasurementResults.py
```diff
@@ -27,7 +27,7 @@ def warmdown_start(self):
 
         return min(
             [
-                parallel[-self.warmup_duration].start_timestamp
+                parallel.start_timestamp - self.warmup_duration
```
jtluka, line 30: Shouldn't this be `parallel.end_timestamp - self.warmup_duration` ?

jtluka (changes requested):
Besides my only comment, this looks ok.

lnst/RecipeCommon/Perf/Measurements/Results/NeperFlowMeasurementResults.py
olichtne: fixed.

jtluka (approved):

PR #384, opened 2024-10-23
Added XDP tx recipe

enhaut:
### Description
* exported parameters of `XDPBench` test tool to measurement
* added XDP tx test
  The test measures how fast the CPU can read an incoming packet, swap MACs and send it back using the same NIC. It's basically hairpinning

### Tests

Functional test ox xdp drop + xdp tx recipes in `J:10038813`

### DIFF
--- a/lnst/Tests/XDPBench.py
+++ b/lnst/Tests/XDPBench.py
@@ -63,7 +63,7 @@ class XDPBenchOutputParser:
         return results
 
     def _parse_line(self, line: str) -> tuple:
-        match = re.search(r"Summary\s+([\d,]+)\srx/s\s+([\d,]+)\serr/s?", line)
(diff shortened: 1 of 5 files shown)


lnst/Tests/XDPBench.py
```diff
@@ -108,7 +108,6 @@ class XDPBench(BaseTestModule):
     interval = IntParam(default=1)
 
     redirect_device = DeviceParam()
-    xdp_mode = ChoiceParam(type=StrParam, choices=("native", "skb"), default="native")
```
olichtne, line 111: removing this line should be done in a different commit since it is readded as a different type

olichtne (changes requested):
just a minor commit refactoring comment, otherwise looks ok

olichtne (approved):
looks ok, merging.

PR #390, opened 2025-01-06
pyproject.toml: add includes into wheel

olichtne:
### Description
After poetry-core update the default of the includes changed so that they're only in the sdist by default and this breaks our lnst installation as the schema file is used during runtime.

### Tests

### Reviews

Closes: #

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,11 @@ packages = [
     { include = "lnst"}
 ]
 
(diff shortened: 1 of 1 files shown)


jtluka:
Testing internally in `J:10416434`

jtluka:
> Testing internally in `J:10416434`

LNST recipe did not crash and runs fine.

PR #398, opened 2025-04-04
Added intel sst-tf support

enhaut:
### Description
Added support for enabling intel sst-tf. This requires changes to `DisableIdleStatesMixin` as well, it has mandatory parameter `minimal_idlestates_latency`, idle states are handled by sst-tf mixin itself. Since idle states mixin is deep in the inheritance tree for ENRT recipes, we would need to split that inheritance tree to skip setting idle states by `DisableIdleStatesMixin`. 

### Tests
`J:11094046` perfromance with no turbo on that machine is something like 44Gbit/s, with turbo enabled it's 55 and with intel sst, we are at 66Gbits.

### DIFF
--- a/lnst/Recipes/ENRT/BaremetalEnrtRecipe.py
+++ b/lnst/Recipes/ENRT/BaremetalEnrtRecipe.py
@@ -1,4 +1,5 @@
 from lnst.Recipes.ENRT.BaseEnrtRecipe import BaseEnrtRecipe
+from lnst.Recipes.ENRT.ConfigMixins.EnableIntelSSTTFMixin import EnableIntelSSTTFMixin
 from lnst.Recipes.ENRT.UseVfsMixin import UseVfsMixin
 from lnst.Recipes.ENRT.PerfTestMixins import CommonPerfTestTweakMixin
 from lnst.Recipes.ENRT.ConfigMixins.DisableTurboboostMixin import (
@@ -26,6 +27,7 @@ class BaremetalEnrtMeasurementGenerators(
 
 
 class BaremetalEnrtCommonMixins(
+    EnableIntelSSTTFMixin,
     DisableTurboboostMixin,
     DisableIdleStatesMixin,
     CommonPerfTestTweakMixin,


lnst/Recipes/ENRT/ConfigMixins/DisableIdleStatesMixin.py
```diff
@@ -31,7 +32,7 @@ def apply_sub_configuration(self, config):
         super().apply_sub_configuration(config)
 
         latency = getattr(self.params, "minimal_idlestates_latency", None)
-        if latency is not None:
+        if latency is not None and latency >= 0:
```
olichtne, line 35: i'm not sure i understand why this is necessary?

you say that
> recipes with `DisableIdleStatesMixin` deep in the inheritance tree.

but `self.params` are exposed to the top level of the Recipe... so you can always override it with a `=None` value? and using a `<0` value is the same - you basically have to at the top level do `minimal_idlestates_latency=-1`... but you can just as well do `minimal_idlestates_latency=None` right?

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+    def intel_sst_tf_host_list(self):
+        """
+        The value of this property is a list of hosts for which the intel
+        SST-TF should be turned on. Derived class can override this property.
+        """
+        return self.matched
+
+    def _is_sst_supported(self, host):
+
+        sst_info = host.run("intel-speed-select --info", job_level=ResultLevel.DEBUG)
```
olichtne, line 31: is this available by default? if not maybe this should be preceded by https://github.com/LNST-project/lnst/blob/master/lnst/Common/Utils.py#L308

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+            cpus.extend(self.params.perf_tool_cpu)
+
+        if "dev_intr_cpu" in self.params:
+            cpus.extend(self.params.dev_intr_cpu)
+
+        # TODO: multi dev intr config mixin support
+
+        return ",".join(str(cpu) for cpu in cpus)
+
+    def apply_sub_configuration(self, config):
```
olichtne, line 62: can we reorder the methods so that the main methods are on top and the helper methods are at the bottom of the class?

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+        super().apply_sub_configuration(config)
+
+        if not self.params.enable_intel_sst_tf:
```
olichtne, line 65: this needs to check that the parameter is defined since it doesn't have a default value.

olichtne (changes requested):

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
olichtne: ah my mistake i read that as "there's no default" for some reason...

lnst/Recipes/ENRT/ConfigMixins/DisableIdleStatesMixin.py
olichtne: parameters are optional by default and this one doesn't say `mandatory=True` so i guess this is already the case...

so if you don't specify the parameter you get `None` as the default returned from `getattr(self.params, "minimal_idlestates_latency", None)`

so then the question now becomes... why provide `minimal_idlestates_latency=-1` when you can simply skip the parameter?

jtluka: Yes, you can simply omit the parameter for the recipe and it would be skipped. See the example of a test that does exactly that, internal uuid is 10050e83-714c-4323-8d22-1199cde85e55.

The recipe prints: `configuration of idle states skipped`

lnst/Recipes/ENRT/ConfigMixins/DisableIdleStatesMixin.py
```diff
@@ -52,6 +53,7 @@ def generate_sub_configuration_description(self, config):
 
     def remove_sub_configuration(self, config):
         for host in self.disable_idlestates_host_list:
-            host.run("cpupower idle-set -E")
+            if self.params.minimal_idlestates_latency >= 0:
```
jtluka, line 56: I'd change this to: `if getattr(self.params, "minimal_idlestates_latency", None)`

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+
+        for host in self.intel_sst_tf_host_list:
+            if not self._is_sst_supported(host):
+                raise LnstError(f"Machine {host} doesn't support Intel SST-TF.")
```
jtluka, line 70: Please use RecipeError exception type instead of LnstError, that should be used at  the Controller level. Applies to all occurences in this MR.

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+from lnst.Recipes.ENRT.ConfigMixins import BaseSubConfigMixin
+
+
+class EnableIntelSSTTFMixin(BaseSubConfigMixin):
+    enable_intel_sst_tf = BoolParam(default=False)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if getattr(self.params, "disable_turboboost", True):
```
jtluka, line 15: It's probably not a big deal, but could we make this check not dependent on the other mixin that does not have to be in other Recipe instance hierarchy?

I'd check the value of `/sys/devices/system/cpu/intel_pstate/no_turbo` instead.

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if not self.params.enable_intel_sst_tf:
+            return
+
+        for host in self.intel_sst_tf_host_list:
+            if not self._is_sst_supported(host):
+                raise LnstError(f"Machine {host} doesn't support Intel SST-TF.")
+
+            idlestates = host.run("cpupower idle-set -E")
+            if not idlestates.passed:
```
jtluka, line 73: I'll leave that up to you, but if you don't need the idlestates variable for other purpose, maybe just squash this into:
```python
if not host.run("cpupower idle-set -E").passed:
```

Applies to other run() calls below.

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -0,0 +1,119 @@
...
+        if not tb_check.passed or tb_check.stdout != "0\n":
+            raise RecipeError("Intel SST-TF requires TurboBoost to be enabled.")
+
+        sst_info = host.run("intel-speed-select --info", job_level=ResultLevel.DEBUG)
+
+        result = sst_info.passed
+        if not re.search(
+            r"Intel\(R\) SST-TF \(feature turbo-freq\) is supported", sst_info.stderr
+        ):
+            result = ResultType.FAIL
```
olichtne, line 99: what's the point of this? this is returned from the `is_sst_supported` method but it's only used in a `if` condition in the `apply` method and not added to an actual `Result`... so this could just be a True/False

but maybe you want to do `sst_info.passed = ResultType.FAIL`?

lnst/Recipes/ENRT/ConfigMixins/EnableIntelSSTTFMixin.py
```diff
@@ -11,7 +11,7 @@
 class EnableIntelSSTTFMixin(BaseSubConfigMixin):
     TURBO_PATH = "/sys/devices/system/cpu/intel_pstate/no_turbo"
 
-    enable_intel_sst_tf = BoolParam(default=False)
+    enable_intel_sst_tf = BoolParam()
```
olichtne, line 14: since the entire `EnableIntelSSTTFMixin` is added in the previous commit can we squash the changes to it from this commit into the previous one?

i'd keep this commit only to add the mixin to the `BaremetalEnrtCommonMixins` class dependencies.

olichtne:
in the test logs for the sst enabled case there's a ping error reported:
```
2025-06-06 13:28:47       (localhost)        -    INFO: Result: PASS, What: Job started: type(module), host(host1), netns(None), Ping(dst=Ip6Address(fc00::2/64), interface=Ip6Address(fc00::1/64), count=100, interval=0.2, size=56)
2025-06-06 13:28:47           (host1)        -   ERROR: returncode = 2
2025-06-06 13:28:47           (host1)        -   ERROR: errors reported by ping
2025-06-06 13:28:47           (host1)        -   ERROR: 
    ping: bind icmp socket: Cannot assign requested address
    
```

is this related or can this be ignored?

olichtne (approved):

PR #400, opened 2025-04-15
Reworked pktgen test module, changable rx/tx queue size support

enhaut:
### Description

* Reworked pktgen to better represent its' structure and also, because the older version does not support multiple flows. Just single connection 5-tuple.
* Support of namespaces in `BaseHWConfigMixin`, previously we could not, e.g. pin interrupts of device that has been moved to separate namespace. 
* Added support for setting custom rx/tx ring size to `DevQueuesConfigMixin`
* Support for instancing lnst's IP address objects from python's IP address objects
* added `DevFlowsPinning` mixin which implements more granular ways of pinning individual flows to IRQ queues. E.g. you can pin flow based on its src IP.

### Tests

* reworked pktgen (compatibility with current test set) - `J:11077013`
* namespaces parsing and custom ring size  in `J:11075382`


lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+        super().hw_config(config)
+
+        config.hw_config["flow_steering_rules"] = {}
+
+        for combination in self.generate_flow_combinations(config):
+            for flow in combination:
+                if self.steer_flow_by(flow) is None:
+                    continue  # skip unpinned flow
+
+                for steering_nic in self.steer_flow_at(flow):
```
jtluka, line 71: Steer verb should use into or to, not at -> `steer_flow_to`.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+                    if steering_nic not in config.hw_config["flow_steering_rules"]:
+                        config.hw_config["flow_steering_rules"][steering_nic] = []
```
jtluka, line 73: You can use `dict.setdefault()` to avoid this if branch.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class DevFlowsPinningHWConfigMixin(BaseHWConfigMixin):
+    """
+        This class is an extension to the :any:`BaseEnrtRecipe` class that
+        enables flow steering across test device queues. Generated flows
+        are iterated and pinned to
+        :attr:`lnst.RecipeCommon.Perf.Measurements.BaseFlowMeasurement.Flow.receiver_cpupin`
+        CPU queue. Flow steering "keys" are configurable by overriding the
+        :meth:`.steer_flow_by` method and can be set independently.
```
jtluka, line 36: Could you explain this a bit more? The steer_flow_by() should be overriden by a derived class - how this would look like?

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+            the same flow to the same queue anyway (based on its' 5-tuple)."
+            )
+
+            logging.warning(f"Using only the first CPU: {cpupin[0]}")
+
+        steer_by = self.steer_flow_by(flow)
+        flow_type = self._get_flow_type(flow, steer_by)
+        steering_key = self._get_steering_key(flow, steer_by)
+
+        job = steering_nic.netns.run(
```
jtluka, line 110: I'm quite sure you can omit the `steering_nic` parameter as it can be fetched from `Flow.receiver_nic`. Yes, it is optional field of the `Flow` but for ENRT recipes it is always present.

To be 100% sure this works as expected, I'd just add a check if the receiver_nic is not None, if it is just skip that flow instance.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        This method should return an steering key, based on
+        which the flow should be steered across the queues.
+
+        :param flow: flow object to pin
+        :returns: steering key
+        """
+        raise NotImplementedError("Should be implemented in derived class.")
+
+    def steer_flow_at(self, flow) -> Iterable[Device]:
```
jtluka, line 213: Grammar fix, `steer_flow_to`

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,236 @@
...
+        not by `queue`. So other actions (by <0 action code) can
+        be specified as well.
+        See https://man7.org/linux/man-pages/man8/ethtool.8.html
+
+        :param flow: flow object to pin
+        :param steering_nic: NIC where the flow should be pinned
+
+        :returns: tuple of rule ID (for deconfig) and description of pinning.
+            That can't be dynamically generated in :meth:`.describe_hw_config`
+            as it woudln't know which flow belongs to which rule ID.
```
jtluka, line 94: Can't you save the Flow:ID mapping in the `config.hw_config` and use it like that ?

jtluka (changes requested):
I'll have probably more comments/questions on the implementation, posting initial review.

lnst/Recipes/ENRT/ConfigMixins/BaseHWConfigMixin.py
```diff
@@ -121,3 +116,28 @@ def _parse_device_settings(self, settings={}):
...
+        host = netns_path[0]
+
+        try:
+            host = getattr(self.matched, host)
+        except AttributeError:
+            raise Exception(f"Host {host} not found in matched hosts")
+
+        if len(netns_path) > 1:
+            try:
+                return getattr(host, netns_path[1])
```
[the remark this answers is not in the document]
jtluka: You're right I missed that both parts of split are used.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
jtluka: Ok, thanks for clarifying.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,242 @@
...
+
+        config.hw_config["flow_steering_rules"] = {}
+
+        for combination in self.generate_flow_combinations(config):
+            for flow in combination:
+                if self.steer_flow_by(flow) is None:
+                    continue  # skip unpinned flow
+
+                for steering_nic in self.steer_flow_to(flow):
+                    # if steering_nic not in config.hw_config["flow_steering_rules"]:
```
jtluka, line 74: You left some commented code here.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
jtluka: Ack, resolving.

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
jtluka: Looks better with the docstring.

lnst/Recipes/ENRT/ConfigMixins/DevRingConfigMixin.py
```diff
@@ -0,0 +1,98 @@
+from json import loads
+from lnst.Common.Parameters import DictParam
+from lnst.Recipes.ENRT.ConfigMixins.BaseHWConfigMixin import BaseHWConfigMixin
+
+
+class DevRingConfigMixin(BaseHWConfigMixin):
```
jtluka, line 6: Please add a docstring with example how to use this.

lnst/Recipes/ENRT/ConfigMixins/DevRingConfigMixin.py
```diff
@@ -0,0 +1,98 @@
...
+    #         "tx-max": 4096,
+    #         "rx": 256,
+    #         "tx": 256,
+    #         "tx-push": false,
+    #         "rx-push": false
+    # } ]
+
+    output = loads(output)[0]  # just single inf
+    result = {}
+    for configured in configured_values:
```
jtluka, line 95: This is just for clarity as I had troubles reading this part of the code.

Since the json output is a dict with keys and values, could you please change this to:
```
for key in configured_keys:
```

lnst/Recipes/ENRT/ConfigMixins/DevQueuesConfigMixin.py
```diff
@@ -4,16 +4,22 @@
 
 class DevQueuesConfigMixin(BaseHWConfigMixin):
     dev_queues = DictParam(mandatory=False)
+    dev_queue_sizes = DictParam(mandatory=False)
```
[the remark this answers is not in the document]
jtluka: That job ended with exception, but more importantly, the log does not contain the ring configuration and I think it's because it was run with old code before splitting the mixin as it contains:
```
    host2 device eno12399np0 queue sizes configured: rx 1024 tx 1024
    host2 device eno12409np1 queue sizes configured: rx 1024 tx 1024
```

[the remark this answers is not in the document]
jtluka: Please run a new job with updated code.

jtluka:
Besides my recent comments, looks ok. Could you add most recent test runs in Beaker to see if it works properly?

lnst/Recipes/ENRT/ConfigMixins/DevFlowsPinningHWConfigMixin.py
```diff
@@ -0,0 +1,241 @@
...
+        which the flow should be steered across the queues.
+        E.g. if steering by src-ip, it returns "src-ip" and
+        the flow will be steered based on its' source IP address.
+
+        If None is returned, flow won't be steered.
+
+        :param flow: flow object to pin
+        :returns: steering key
+        """
+        raise NotImplementedError("Should be implemented in derived class.")
```
olichtne, line 216: and this is implemented where? i don't see it in this PR or any other PRs from you either?

olichtne: right but... that means this specific method with `NotImplementedError` is implemented somewhere, but not here...

olichtne: my point is... you're adding a config mixin class... which isn't used anywhere, not in this PR... or any other PR I see from you. so why do we need it?

here you have an example, but is this something we plan on running or including somewhere? what's the actual use of the mixin class?

lnst/Tests/PktGen.py
```diff
@@ -107,116 +115,216 @@ def _parse_values(self, params) -> dict[str, str]:
...
 
-    def _pg_ctrl(self, cmd: str):
-        self._write_command("/proc/net/pktgen/pgctrl", cmd)
+        self._cmd("reset")
 
-    def _pg_thread(self, thread: int, cmd: str):
-        self._write_command(f"/proc/net/pktgen/kpktgend_{thread}", cmd)
+    def _check_cpu_pinning(self):
+        # check if each device is pinned to separate CPU
+        return len(set(dev["cpu"] for dev in self.params.config)) == len(
```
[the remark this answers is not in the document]
olichtne: we can keep it as is for now, we can lift the restriction later if needed.

lnst/Recipes/ENRT/PerfTestMixins/DevFlowsPinningMixin.py
```diff
@@ -0,0 +1,249 @@
...
+        which the flow should be steered across the queues.
+        E.g. if steering by src-ip, it returns "src-ip" and
+        the flow will be steered based on its' source IP address.
+
+        If None is returned, flow won't be steered.
+
+        :param flow: flow object to pin
+        :returns: steering key
+        """
+        return getattr(self.params, "steer_flows_by", None)
```
olichtne, line 224: is this enough? the function has a `flow` parameter which isn't used. or is the parameter supposed to be a single value that is used for all flows?

olichtne (changes requested):

olichtne (approved):

PR #406, opened 2025-07-02
TcRunMeasurementResults: rename description to describe

olichtne:
### Description
This is for consistency with other MeasurementResults, this should be a method called describe not a property called description. Not sure why it ended up this way...

### Tests

### Reviews

Closes: #


olichtne:
test scheduled: `J:11340126`

olichtne:
new test with a fix: `J:11345739`

PR #409, opened 2025-07-08
DevNfcRxFlowHashConfigMixin: disable symmetric hashing

olichtne:
### Description
If we're requesting any of the "non-symmetric" hash options we should disable symmetric hashing, otherwise the configuration is considered invalid and may error out.

This is based on an internal discussion investigating a potential regression, it turns out that when symmetric hashing is enabled configuring the `rx-flow-hash` option to use only one of src or dst fields is invalid and not supported.

This is explained in: https://elixir.bootlin.com/linux/v6.16-rc4/source/net/ethtool/ioctl.c#L1021

### Tests
tested via internal beaker job: `J:11347282`

### Reviews
@jtluka 

Closes: #

### DIFF
--- a/lnst/Recipes/ENRT/ConfigMixins/DevNfcRxFlowHashConfigMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/DevNfcRxFlowHashConfigMixin.py
@@ -1,3 +1,5 @@
+import logging
+
 from lnst.Common.Parameters import DictParam
 from lnst.Recipes.ENRT.ConfigMixins.BaseHWConfigMixin import BaseHWConfigMixin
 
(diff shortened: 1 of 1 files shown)


lnst/Recipes/ENRT/ConfigMixins/DevNfcRxFlowHashConfigMixin.py
```diff
@@ -26,6 +26,11 @@ def hw_config(self, config):
                     "original": original_flow_hash,
                 }
 
+                if protocol_setting not in {"sd", "fn", "sdfn"}:
+                    device.host.run(
+                        f"ethtool -X {device.name} xfrm none"
+                    )
```
jtluka, line 32: Could we add some log message that this is happening?

jtluka (dismissed):
Otherwise looks ok. Ack to merge.

lnst/Recipes/ENRT/ConfigMixins/DevNfcRxFlowHashConfigMixin.py
olichtne: added log message.

PR #410, opened 2025-07-09
Nftables rule scale recipe

SirPhuttel:
Add a recipe suited for testing nftables rule traversal performance (or the
other way round, a given rule's impact on packet processing). It takes a rule
and the number of times to add it to the ruleset as parameter and generates
results using FlowMeasurementGenerator. Sample instantiation looks like this:
```
r = NftablesRuleScaleRecipe(perf_tests=["udp_stream"],
                            offload_combinations=[],
			    rule="fib saddr . iif oif missing drop",
			    scale=1000)
Controller().run(recipe_instance)
```

The first commit in this PR adds SimpleNetnsRouterRecipe which enables testing
the kernel's forwarding path without the need for three hosts and a dedicated
link between two of them. It simply creates a netns on the second host and
inserts the netns-facing VETH interface as second endpoint instead of the
second host's physical NIC.

NftablesRuleScaleRecipe added by the second commit makes use of the netns
router by adding rules to the second host's netfilter forward hook. This should
prevent the test setup from disturbing control traffic between the agents.

### DIFF
--- a/lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
+++ b/lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
@@ -1,0 +1,55 @@
+from lnst.Recipes.ENRT.SimpleNetnsRouterRecipe import SimpleNetnsRouterRecipe
+from lnst.Recipes.ENRT.ConfigMixins.FirewallMixin import NftablesMixin
+from lnst.Common.Parameters import StrParam, IntParam, BoolParam
+
+class NftablesRuleScaleRecipe(SimpleNetnsRouterRecipe, NftablesMixin):
+    """
+    This recipe combines SimpleNetnsRouterRecipe and NftablesMixin for testing
+    routing throughput impact of specific nftables rules. To generate
+    meaningful results, the rule is simply repeated multiple times, thereby
+    amplifying the impact.
+    Rules are added to host2's forwarding hook which routes test traffic
+    between host1 and a local netns.
+
+    :param rule:
+        The actual rule to insert repeatedly into the router's forwarding
+        chain.
+    :type rule: :any:`StrParam` representing the nftables rule.
+
+    :param scale:
+        The number of times to insert :any:`rule` into the ruleset.
+    :type scale: :any:`IntParam` > 0
+
+    :param flowtable:
+        Whether to offload established connections to a flowtable or not.
+    :type flowtable: :any:`BoolParam`
+    """
+    rule = StrParam(mandatory=True)
+    scale = IntParam(mandatory=True)
+    flowtable = BoolParam(mandatory=False, default=False)
+
+    @property
+    def firewall_rulesets_generator(self):
+        rule = self.params.get('rule')
+        scale = self.params.get('scale')
+        ruleset = [
+            "flush ruleset",
+            "add table inet t",
+            "add chain inet t forward { type filter hook forward priority filter; }",
+        ] + [f"add rule inet t forward {rule}"] * scale
+        if self.params.get('flowtable'):
+            devs = f"{self.matched.host2.eth0.name}, {self.matched.host2.pn0.name}"
+            ftspec = f"hook ingress priority filter; devices = {{ {devs} }};"
+            ruleset.append(f"add flowtable inet t ft {{ {ftspec} }}")
+            ruleset.append("add rule inet t forward ct state established flow add @ft")
+        yield { self.matched.host2: "\n".join(ruleset) }
+
+    def generate_sub_configuration_description(self, config):
+        rule = self.params.get('rule')
+        scale = self.params.get('scale')
+        desc = super().generate_sub_configuration_description(config)
+        msg = f"NftablesScale: Ruleset with {scale} times '{rule}'"
+        if self.params.get('flowtable'):
+            msg += " and a flowtable"
+        desc.append(msg)
+        return desc
--- a/.github/runner_all_enrt.py
+++ b/.github/runner_all_enrt.py
@@ -85,6 +85,10 @@ for recipe_name in dir(enrt_recipes):
         }
         params["perf_tool_cpu"] = [0]
 
+    if recipe_name == "NftablesRuleScaleRecipe":
+        params["rule"] = "accept"
+        params["scale"] = 1
+
     if issubclass(recipe, OffloadSubConfigMixin):
         params['offload_combinations'] = []
 
--- a/docs/source/nftables_rule_scale_recipe.rst
+++ b/docs/source/nftables_rule_scale_recipe.rst
@@ -1,0 +1,6 @@
+NftablesRuleScaleRecipe
+^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autoclass:: lnst.Recipes.ENRT.NftablesRuleScaleRecipe.NftablesRuleScaleRecipe
+   :members:
+   :show-inheritance:
--- a/docs/source/specific_scenarios.rst
+++ b/docs/source/specific_scenarios.rst
@@ -29,3 +29,4 @@ Specific ENRT scenarios
     l2tp_tunnel_recipe
     linux_bridge_recipe
     linux_bridge_over_bond_recipe
+    nftables_rule_scale_recipe
--- a/lnst/Recipes/ENRT/__init__.py
+++ b/lnst/Recipes/ENRT/__init__.py
@@ -118,3 +118,4 @@ from .CTFulltableInsertionRateRecipe import CTFulltableInsertionRateRecipe
 from lnst.Recipes.ENRT.BaseEnrtRecipe import BaseEnrtRecipe
 from lnst.Recipes.ENRT.BaseTunnelRecipe import BaseTunnelRecipe
 from lnst.Recipes.ENRT.SimpleNetnsRouterRecipe import SimpleNetnsRouterRecipe
+from lnst.Recipes.ENRT.NftablesRuleScaleRecipe import NftablesRuleScaleRecipe


lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
```diff
@@ -0,0 +1,96 @@
...
+from lnst.RecipeCommon.Ping.PingEndpoints import PingEndpoints
+from lnst.Recipes.ENRT.helpers import ip_endpoint_pairs
+from lnst.Controller.NetNamespace import NetNamespace
+from lnst.Common.IpAddress import interface_addresses
+from lnst.Devices.VethPair import VethPair
+from collections.abc import Collection
+
+class SimpleNetnsRouterRecipe(SimpleNetworkRecipe):
+    """
+    This recipe implements Enrt testing for a simple network scenario that
```
jtluka, line 14: Please add some notes what this recipe is about.

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
```diff
@@ -0,0 +1,96 @@
...
+
+        host2.ns.np0._ipr_wrapper("route", "add", dst="default",
+                                  gateway=f"{pn0_addr}")
+        host2.ns.np0._ipr_wrapper("route", "add", dst="default",
+                                  gateway=f"{pn0_addr6}")
+
+        host2.run("sysctl -w net.ipv4.ip_forward=1")
+        host2.run("sysctl -w net.ipv6.conf.all.forwarding=1")
+
+        ipv4_addr = interface_addresses(self.params.net_ipv4)
```
jtluka, line 78: Somewhat confusing - could we change the previous ipv4_addr to ipv4_ns_addr or similar?

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
```diff
@@ -0,0 +1,96 @@
...
+        next(ipv4_addr)
```
jtluka, line 79: I'm not quite sure this is necessary? Why we need to skip one address?a

jtluka (changes requested):
Left few questions and comments, overall looks ok, but check my comments.

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
```diff
@@ -0,0 +1,50 @@
+from lnst.Recipes.ENRT.SimpleNetnsRouterRecipe import SimpleNetnsRouterRecipe
+from lnst.Recipes.ENRT.ConfigMixins.FirewallMixin import NftablesMixin
+from lnst.Recipes.ENRT.MeasurementGenerators.FlowMeasurementGenerator import FlowMeasurementGenerator
+from lnst.Common.Parameters import StrParam, IntParam, BoolParam
+
+class NftablesRuleScaleRecipe(SimpleNetnsRouterRecipe, NftablesMixin, FlowMeasurementGenerator):
```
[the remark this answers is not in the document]
SirPhuttel: Oh, thanks!

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
```diff
@@ -0,0 +1,50 @@
...
+        Whether to offload established connections to a flowtable or not.
+    :type flowtable: :any:`BoolParam`
+    """
+    rule = StrParam(mandatory=True)
+    scale = IntParam(mandatory=True)
+    flowtable = BoolParam(mandatory=False, default=False)
+
+    @property
+    def firewall_rulesets_generator(self):
+        rule = self.params.get('rule', None)
```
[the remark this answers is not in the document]
SirPhuttel: Oh, right. I took those calls from sample code and didn't check what that second parameter was for. A classic case of cargo cult! /o\

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
SirPhuttel: This is because I need the second address generated from `self.params.net_ipv4`. It is the host1-facing external IP of host2 which I use to create a route to the netns-network on host1.
As said, maybe an extra getter (or one I didn't discover yet) will simplify things here.

SirPhuttel:
Changes since v1:
- Improve docstrings of added classes
- Eliminate the need for temporary address variables in SimpleNetnsRouterRecipe by fetching the addresses from NetworkParams via an index
- Break some long lines
- NftablesRuleScaleRecipe does not need to explicitly inherit FlowMeasurementGenerator
- Drop pointless second parameter from self.params.get() calls
- Simplify ruleset construction in NftablesRuleScaleRecipe
- Add NftablesRuleScaleRecipe to docs/

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
SirPhuttel: I hope it's less chaotic now, noticed I can extract the IPs from `self.params.net_ipv4` directly via an index.

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
```diff
@@ -0,0 +1,95 @@
...
+        ipv6_addr = interface_addresses(self.params.netns_ipv6)
+        config.configure_and_track_ip(host2.pn0, next(ipv4_addr))
+        config.configure_and_track_ip(host2.pn0, next(ipv6_addr))
+        config.configure_and_track_ip(host2.ns.np0, next(ipv4_addr))
+        config.configure_and_track_ip(host2.ns.np0, next(ipv6_addr))
+        host2.pn0.up_and_wait()
+        host2.ns.np0.up_and_wait()
+
+        for gw in (self.params.netns_ipv4[1], self.params.netns_ipv6[1]):
+            host2.ns.np0._ipr_wrapper("route", "add", dst="default",
```
olichtne, line 71: not entirely sure i like this call of a "private" wrapper method from the recipe, this should probably added as a new public Device class method `add_route`... but i think for now let's keep it as is...

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
```diff
@@ -0,0 +1,95 @@
...
+                                      gateway=f"{gw}")
+
+        host2.run("sysctl -w net.ipv4.ip_forward=1")
+        host2.run("sysctl -w net.ipv6.conf.all.forwarding=1")
```
olichtne, line 75: unless these are default config values, i think these need the reverse action added to the `test_wide_deconfiguration` method, otherwise they may affect other tests running in sequence after this one.

we used to have a tester api for `sysctl` that used to take care of this automatically but we don't have this right now...

olichtne: this applies to other calls of `host.run('sysctl...')` as well...

olichtne (changes requested):

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
n0-1: Guess I could also just call `host2.run('ip -net host2ns route add ...')`, not sure if this improves things or not though.

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
n0-1: Ah, thanks! I missed that. Do I have to kill the netns as well? If so, will `host2.ns = None` do? (Please pardon my ignorance.)

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
olichtne: yeah, that's a bit better, it at least ensures that the recipe won't stop working if we decide to change the private method somehow...
i'd do this though:
```
host2.ns.run("ip route add...")
```
to run the command in the ns directly

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
olichtne: namespaces are cleaned up automatically so just the sysctls will do, thanks!

lnst/Recipes/ENRT/SimpleNetnsRouterRecipe.py
SirPhuttel: Oh, sorry Ondrej! I forgot about this open issue (and don't ask me what I assumed it was about when I marked this one as resolved. /o\

olichtne:
Looks good to me, i think this can be merged unless @jtluka has some other comments.

@SirPhuttel did you try to run this in beaker or just locally?

Maybe we can add one test run using the internal LNST infrastructure if you can provide some configuration example to just have a double check that everything works as expected. I can schedule that no problem if you provide me with the configuration.

SirPhuttel:
Hi @olichtne,

> @SirPhuttel did you try to run this in beaker or just locally?

Just locally, using some Fedora VMs.

> Maybe we can add one test run using the internal LNST infrastructure if you can provide some configuration example to just have a double check that everything works as expected. I can schedule that no problem if you provide me with the configuration.

That would be great! My goto is:
```
ctl = Controller()
r = NftablesRuleScaleRecipe(perf_tests=["udp_stream"],
			    offload_combinations=[],
			    rule="fib saddr . iif oif missing drop",
			    scale=10)
ctl.run(r)
```
Cheers, Phil

PR #411, opened 2025-07-09
Eliminate some harmless Error: lines from output

SirPhuttel:
Some commands are expected to fail especially with virtual NICs like VETH or loopback. Make sure `exec_cmd()` respects `log_outputs` parameter and pass that as `False` when probing for NIC features. Setters intentionally left untouched so failing deliberate action is not washed away.

The first commit is entirely unrelated, but too trivial to justify a separate PR.

### DIFF
--- a/lnst/Common/ExecCmd.py
+++ b/lnst/Common/ExecCmd.py
@@ -77,7 +77,6 @@ def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=F
             log_output(logging.debug, "Stderr", data_stderr)
     if subp.returncode and die_on_err:
         err = ExecCmdFail(cmd, subp.returncode, [data_stdout, data_stderr], report_stderr)
-        logging.error(err)
         raise err
 
     if json:
--- a/lnst/Devices/Device.py
+++ b/lnst/Devices/Device.py
@@ -1085,9 +1085,11 @@ class Device(object, metaclass=DeviceMeta):
 
     def _get_vf_count(self):
         # TODO: create sysfs api for Device, then run self.[get|set]_sysfs("device/sriov_numvfs")
-        stdout, _ = exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs")
-
-        return int(stdout)
+        try:
+            stdout, _ = exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs")
+            return int(stdout)
+        except ExecCmdFail:
+            raise DeviceFeatureNotSupported(f"Device {self.name} not SR-IOV capable")
 
     def _get_vf_representor(self, vf_index: int):
         pf_number = int(self.phys_port_name[1:])
--- a/lnst/Common/Parameters.py
+++ b/lnst/Common/Parameters.py
@@ -26,8 +26,8 @@ class ParamError(LnstError):
 class Param(object):
     """Base Parameter class
 
-    Can beused to define your own specific parameter type. Param derived classes
-    serve as *type checkers* to enable earlier failure of the recipe.
+    Can be used to define your own specific parameter type. Param derived
+    classes serve as *type checkers* to enable earlier failure of the recipe.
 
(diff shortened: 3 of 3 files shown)


SirPhuttel:
Changes since v1:
- Wrapped some long lines
- Added spaces around assignments
- Added SoB to commits

lnst/Common/ExecCmd.py
```diff
@@ -77,7 +77,8 @@ def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=F
             log_output(logging.debug, "Stderr", data_stderr)
     if subp.returncode and die_on_err:
         err = ExecCmdFail(cmd, subp.returncode, [data_stdout, data_stderr], report_stderr)
-        logging.error(err)
+        if log_outputs:
+            logging.error(err)
```
olichtne, line 81: imo this should just be removed... in this path:
1. die_on_err == True
2. we **raise an exception**

no need to log here... if `log_outputs == True`, the error should already be logged a couple of lines higher.

since an exception is raised, it should be the responsibility of the **caller** to handle that exception, and logging the exception is part of that handler.

This is probably some old relict where noone took a look at it for a long time.

So in the specific cases that this PR updates, it just changes the logging of this exception - it was logged, now it isn't

But these same callers, also handle the exception and either re-raise it as something different and that gets handled somewhere higher as a non-issue. So I think the specific cases of where we can "ignore" the situations are already covered - somewhere there's an exception handler that ignores this.

Having logs along the way IMO is useful, but they should be placed in points where they make sense - so for this exception, it should be in the handler of the exception. Basically don't log "i generated an exception!"... instead log "i caught and handled this exception!" if it is a useful information.

lnst/Devices/Device.py
```diff
@@ -855,7 +855,8 @@ def tx_pause_frames(self, value):
 
     def _read_pause_frames(self):
         try:
-            res, _ = exec_cmd("ethtool -a %s" % self.name)
+            res, _ = exec_cmd("ethtool -a %s" % self.name,
+                              log_outputs = False)
```
olichtne, line 860: with my previous comment i think this doesn't have to change, or is it a problem that the failed messages are also part of the `debug` logs?

on the other hand we may want to consider adding a debug log to the exception handler... if this REALLY interests us.

lnst/Devices/Device.py
```diff
@@ -1087,9 +1088,12 @@ def _get_vf_rep_devices(self):
...
         # TODO: create sysfs api for Device, then run self.[get|set]_sysfs("device/sriov_numvfs")
-        stdout, _ = exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs")
-
-        return int(stdout)
+        try:
+            stdout, _ = exec_cmd(f"cat /sys/class/net/{self.name}/device/sriov_numvfs",
+                                 log_outputs = False)
+            return int(stdout)
+        except ExecCmdFail:
+            return 0
```
olichtne, line 1096: alternatively i'd say we can also `raise DeviceFeatureNotSupported` so that the user of this may decide on if reading this is a MUST or if it is a NICE TO HAVE.

this may be useful when there's information difference between "vf_count == 0" and "vfs are NOT SUPPORTED".

SirPhuttel:
> > I'm in favor in fixing this annoying issue. But I think we should do this differently. Sometimes we're interested in functional bugs when querying device features that are supported suddenly stop working.
> > So, my proposal would be to have a feature dictionary per device type/driver (?) that would be checked and if the feature is not supported the error can be masked out.
> 
> I'm not sure checking the device type and driver is sufficient. Often the same driver serves a number of devices and probes features itself. Maybe including the PCI Vendor/Device IDs in that table could be enough, but I'm pretty sure there are cases where even this isn't.

A pointless warning which remains after my PR here is `Link down operation on LoopbackDevice disallowed`. For this one at least, the proposed table with features might be ideal - guarding all kinds of operations by `self.name != "lo"` statements certainly isn't!

jtluka:
> @jtluka what do you think about this solution?

Looks ok.

PR #417, opened 2025-08-11
HOTFIX: DevNfcRxFlowHashConfigMixin: delete xfrm_original only if defined

olichtne:
### Description
If the dictionary key isn't defined the deletion here would fail with a traceback.

### Tests
CI should be enough.

internal beaker job: `J:11524005`

### Reviews
@jtluka 

Closes: #


jtluka (approved):

olichtne:
test finished ok, merging.

PR #434, opened 2026-03-30
Containerized controller

enhaut:
### Description

Better support for controller containerization. Now, it suports agent auto setup by `test_environment.json` file and also tests db defined in `test_db.json` file. Container is build and after merging its pushed to quay https://quay.io/repository/lnst/lnst-controller repository (takes ~5minutes). Both `:latest` and `:commit sha` manifests are created, commit sha manifest refers to build based on specific LNST version on both controller and agents.

### Tests
Ran locally against agents without LNST. It installed LNST and ran tests defined in test db:
```
podman run -v $(pwd)/results/:/root/.lnst/results:z -v machine_keys:/root/.ssh:ro -v machines.json:/lnst/container_files/controller/pool/test_environment.json:z --rm -it quay.io/sdobron/lnst-test
```

Resolves https://github.com/LNST-project/lnst/issues/433

### DIFF
--- a/.github/workflows/container-build.yml
+++ b/.github/workflows/container-build.yml
@@ -1,0 +1,52 @@
+name: Build and Push Controller Container
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    paths:
+      - 'container_files/controller/**'
+
+concurrency:
+  group: container-build
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Source branch checkout
+      uses: actions/checkout@v4
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+
+    - name: Set LNST_SRC
+      run: |
+        if [ "${{ github.event_name }}" = "pull_request" ]; then
+          LNST_SRC="${{ github.event.pull_request.head.repo.clone_url }}@${{ github.event.pull_request.head.sha }}"
+        else
+          LNST_SRC="${{ github.server_url }}/${{ github.repository }}.git@${{ github.sha }}"
+        fi
+        echo "LNST_SRC=$LNST_SRC"
+        echo "LNST_SRC=$LNST_SRC" >> $GITHUB_ENV
+
+    - name: Login to Quay.io
+      if: github.event_name == 'push'
+      run: podman login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASS }} quay.io
+
+    - name: Build multi-arch container
+      run: |
+        podman manifest create quay.io/lnst/lnst-controller:latest
+        podman build \
+          --platform linux/amd64,linux/arm64 \
+          --build-arg LNST_SRC=${{ env.LNST_SRC }} \
+          --manifest quay.io/lnst/lnst-controller:latest \
+          container_files/controller
+
+    - name: Push multi-arch manifest
+      if: github.event_name == 'push'
+      run: |
+        podman manifest push quay.io/lnst/lnst-controller:latest
+        podman manifest push quay.io/lnst/lnst-controller:latest quay.io/lnst/lnst-controller:${{ github.sha }}
--- a/container_files/controller/Dockerfile
+++ b/container_files/controller/Dockerfile
@@ -16,8 +16,13 @@ RUN dnf install -y initscripts \
     libnl3-devel
 
 RUN mkdir -p /root/.lnst
-COPY . /lnst
-COPY container_files/controller/pool /root/.lnst/pool
+ARG LNST_SRC=https://github.com/LNST-project/lnst.git
+RUN repo="$LNST_SRC" && \
+    ref="" && \
+    case "$repo" in *@*) ref="${repo##*@}"; repo="${repo%@*}" ;; esac && \
+    git clone "$repo" /lnst && \
+    if [ -n "$ref" ]; then cd /lnst && git checkout "$ref"; fi
+ENV LNST_SRC=${LNST_SRC}
 
 ENV UV_PROJECT_ENVIRONMENT=/root/lnst_venv
 RUN cd /lnst && uv sync
--- a/container_files/controller/Dockerfile
+++ b/container_files/controller/Dockerfile
@@ -13,9 +13,13 @@ RUN dnf install -y initscripts \
     libnl3 \
     lksctp-tools-devel \
     git \
-    libnl3-devel
+    libnl3-devel \
+    jq \
+    sshpass \
+    openssh-clients
 
 RUN mkdir -p /root/.lnst
+# LNST_SRC: git URL used to clone LNST source, accepts "repo_url" or "repo_url@ref" (branch/tag/commit)
 ARG LNST_SRC=https://github.com/LNST-project/lnst.git
 RUN repo="$LNST_SRC" && \
     ref="" && \
@@ -25,7 +29,7 @@ RUN repo="$LNST_SRC" && \
 ENV LNST_SRC=${LNST_SRC}
 
 ENV UV_PROJECT_ENVIRONMENT=/root/lnst_venv
-RUN cd /lnst && uv sync
+RUN cd /lnst && uv sync --extra trex
 # UV_PROJECT_ENVIRONMENT places the venv outside /lnst
 # to prevent conflicts when user mounts host-machine's
 # lnst dir to /lnst
--- a/container_files/controller/container_runner.py
+++ b/container_files/controller/container_runner.py
@@ -1,17 +1,25 @@
+import json
 import os
+import ssl
 import sys
 import traceback
-from typing import Any, Type, Optional
+import zipfile
+from functools import reduce
+from typing import Any
+from urllib.request import urlopen
 
 from lnst.Recipes.ENRT import *
-from lnst.Controller.Recipe import BaseRecipe
+from lnst.Controller.Recipe import BaseRecipe, export_recipe_run
 from lnst.Controller.Controller import Controller
 from lnst.Controller.RecipeResults import ResultLevel, ResultType
 from lnst.Controller.MachineMapper import ContainerMapper
 from lnst.Controller.ContainerPoolManager import ContainerPoolManager
 
 from lnst.Controller.RunSummaryFormatters import *
-from lnst.Controller.RunSummaryFormatters.RunSummaryFormatter import RunSummaryFormatter
+
+RESULTS_DIR = "/root/.lnst/results"
+POOL_DIR = "/root/.lnst/pool"
+TEST_DB = os.getenv("TEST_DB", "/lnst/container_files/controller/test_db.json")
 
 
 class ContainerRunner:
@@ -22,7 +30,6 @@ class ContainerRunner:
     * DEBUG: Set to 1 to enable debug mode
     * RECIPE: Name of the recipe class to run
     * RECIPE_PARAMS: Parameters to pass to the recipe class
-    * FORMATTERS: List of formatters to use
     * MULTIMATCH: Set to 1 to enable multimatch mode
 
     Agents in containers-specific environment variables:
@@ -33,14 +40,30 @@ class ContainerRunner:
 
     def __init__(self) -> None:
         self._controller = Controller(**self._parse_controller_params())
-        self._recipe_params: dict[str, Any] = self._parse_recipe_params()
 
-        if not os.getenv("RECIPE"):
-            raise ValueError("RECIPE environment variable is not set")
-        self._recipe_cls: Type[BaseRecipe] = eval(os.getenv("RECIPE", ""))
-        self._recipe: Optional[BaseRecipe] = None
+        if os.getenv("RECIPE"):
+            self._test_db = [
+                {
+                    "recipe_name": os.getenv("RECIPE", ""),
+                    "params": self._parse_recipe_params(),
+                },
+            ]
+        else:
+            self._test_db = self._load_test_db()
 
-        self._formatters: list[Type[RunSummaryFormatter]] = self._parse_formatters()
+
+    @staticmethod
+    def _load_test_db() -> list[dict[str, Any]]:
+        uri = TEST_DB
+        if "://" not in uri:
+            uri = f"file://{uri}"
+
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+
+        with urlopen(uri, context=ctx) as resp:
+            return json.load(resp)
 
     def _parse_controller_params(self) -> dict:
         params = {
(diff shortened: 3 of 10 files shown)


container_files/controller/container_runner.py
```diff
@@ -75,34 +101,141 @@ def _parse_formatters(self) -> list[Type[RunSummaryFormatter]]:
...
+            results.append((test_id, test_result))
+            overall = ResultType.max_severity(overall, test_result)
+
+        print(f"\n{'=' * 60}")
+        print("Test Summary:")
+        print(f"{'=' * 60}")
+        for test_id, result in results:
+            status = "PASS" if result == ResultType.PASS else "FAIL"
+            print(f"  {test_id}: {status}")
+        overall_status = "PASS" if overall == ResultType.PASS else "FAIL"
```
jtluka: Could you please rename the `overall` to `overall_status` and remove this assignment line, where the assignment value can be used directly in the print statement below?

Having two variables with one named `overall` and the other `overall_status` in the same scope is quite confusing.

container_files/controller/container_runner.py
```diff
@@ -75,34 +101,141 @@ def _parse_formatters(self) -> list[Type[RunSummaryFormatter]]:
...
-        for formatter in self._formatters:
-            fmt = formatter(level=ResultLevel.IMPORTANT)
-            for run in self._recipe.runs:
-                print(fmt.format_run(run))
-                overall_result = ResultType.max_severity(
-                    overall_result, run.overall_result
-                )
-
-        return overall_result
+            sys.exit(1)
```
jtluka: Personally, I'd rather have this function return a bool value and move `sys.exit()` to the caller (__main__.py) based on the function value.

container_files/controller/create_pool.py
```diff
@@ -0,0 +1,94 @@
...
+
+    args = parser.parse_args()
+
+    pool_path = create_machine_pool(args.test_environment_description, args.output)
+    print(pool_path)
+
+
+def create_machine_pool(test_environment_description, pool_path):
+    # create a clean pool directory
+    pool_path.mkdir(parents=True, exist_ok=True)
```
jtluka: I'd add some error handling here in case the directory could not be created.

container_files/controller/create_pool.py
```diff
@@ -0,0 +1,94 @@
...
+        pretty_xml = minidom.parseString(xml_str).toprettyxml(indent="    ")
+
+        # Save to file
+        save_path = Path(pool_path) / f"{test_system_name}_agent.xml"
+        save_path.write_text(pretty_xml)
+        print(f"Generated XML at: {save_path}")
+
+    return pool_path
+
+def add_param(parent, name, value):
```
jtluka, line 72: I don't see much value in having this function. I't s a one-liner anyway.

container_files/controller/Dockerfile
```diff
@@ -13,14 +13,22 @@ RUN dnf install -y initscripts \
...
+    libnl3-devel \
+    jq \
+    sshpass \
+    openssh-clients
 
 RUN mkdir -p /root/.lnst
-COPY . /lnst
-COPY container_files/controller/pool /root/.lnst/pool
+ARG LNST_SRC=https://github.com/LNST-project/lnst.git
+RUN repo="$LNST_SRC" && \
```
jtluka, line 24: Please add a comment what this actually does, Gemini translated this to:
```
clone a Git repository (specified by an environment variable) and optionally switch to a specific branch, tag, or commit hash if one is provided after an @ symbol
```

container_files/controller/setup_agent.sh
```diff
@@ -0,0 +1,85 @@
...
+
+echo "Installing system packages..."
+dnf install -y \
+    python3-devel \
+    git \
+    gcc \
+    libnl3-devel \
+    iproute-tc \
+    tcpdump \
+    iperf3 \
```
jtluka, line 37: Please add NetworkManager-config-server

docs/source/extensions.rst
```diff
@@ -183,19 +183,97 @@ following parameters as environment variables to controller container:
...
 
-Using baremetal agents
+Test environment description
+````````````````````````````
+
+Instead of manually creating machine pool XMLs, you can provide a test environment
+description file (``test_environment.json``). The controller generates pool XMLs
+from it at startup.
+
+Place the file at ``container_files/controller/pool/test_environment.json``
```
jtluka: The file can be anywhere, no? Reading through the documentation further I do not see any guide how to mount the file into a controller container instance.

docs/source/extensions.rst
```diff
@@ -229,11 +326,89 @@ Before running the container, you need to provide environment variables:
...
+                "driver": "ice"
+            }
+        }
+    ]
+
+Tests run sequentially and each test is independent -- a failure in one test does not
+prevent subsequent tests from running. A summary of pass/fail results is printed at the
+end of the run.
+
+To use the test database, simply run the container without ``RECIPE`` and ``RECIPE_PARAMS``:
```
jtluka: I'd either remove this as you already have this at the beginning of the chapter or rephrase to:
```
To use the test database, ``RECIPE`` and ``RECIPE_PARAMS`` must not be specified.
```

jtluka (commented):
Overall OK, just minor issues and suggestions.

docs/source/extensions.rst
enhaut: rephrased

docs/source/extensions.rst
enhaut: Yes, anywhere. Added example

container_files/controller/setup_agent.sh
enhaut: added

container_files/controller/Dockerfile
enhaut: added

container_files/controller/create_pool.py
enhaut: done

container_files/controller/container_runner.py
enhaut: done, thanks

container_files/controller/create_pool.py
enhaut: Hmm, yes but i think its more readable this way compared to:
```python
ET.SubElement(eth_params, "param", {"name": "hwaddr", "value": str(mac_address)})
ET.SubElement(eth_params, "param", {"name": "driver", "value": "any"})
ET.SubElement(eth_params, "param", {"name": "nic_speed", "value": "1"})
ET.SubElement(eth_params, "param", {"name": "nic_model", "value": "1"})
```

container_files/controller/entrypoint.sh
```diff
@@ -1,2 +1,90 @@
 #!/bin/sh
```
Axonis: You are using bash features (e.g. [[ ]] (( )) ), yet shebang is sh

container_files/controller/container_runner.py
```diff
@@ -33,15 +42,32 @@ class ContainerRunner:
...
+            self._test_db = [
+                {
+                    "recipe_name": os.getenv("RECIPE", ""),
+                    "params": self._parse_recipe_params(),
+                },
+            ]
+        else:
+            self._test_db = self._load_test_db()
 
         self._formatters: list[Type[RunSummaryFormatter]] = self._parse_formatters()
```
Axonis: No longer used after you've moved into HumanReadableRunSummaryFormatter

container_files/controller/create_pool.py
```diff
@@ -0,0 +1,97 @@
+#!/bin/python
+
+import json
+import yaml
```
Axonis: It's kind of weird, that yaml is trex dependency, yet it's also depedency here... it can be run wtihout the trex - which would crash

container_files/controller/container_runner.py
```diff
@@ -75,34 +101,140 @@ def _parse_formatters(self) -> list[Type[RunSummaryFormatter]]:
...
+            )
+            sys.exit(1)
+
+
 if __name__ == "__main__":
+    _check_dir_access(POOL_DIR)
+    _check_dir_access(RESULTS_DIR)
     runner = ContainerRunner()
     exit_code = 0 if runner.run() == ResultType.PASS else 1
+    runner._zip_results()
```
Axonis: If the run crashes for whatever reason the zip_results might be unreachable due to exception jumping out, do we want this behavior?

container_files/controller/container_runner.py
```diff
@@ -75,34 +101,140 @@ def _parse_formatters(self) -> list[Type[RunSummaryFormatter]]:
...
         ]
 
+    def _export_results(self, recipe, result_dir):
+        os.makedirs(result_dir, exist_ok=True)
+
+        # Export human-readable log (with debug output)
+        hr_fmt = HumanReadableRunSummaryFormatter(level=ResultLevel.DEBUG)
+        with open(os.path.join(result_dir, "controller.log"), "w") as f:
+            for run in recipe.runs:
+                f.write(hr_fmt.format_run(run))
```
Axonis: Other formatters are wrapped in the try except blocks (you never know...), so if this throws on any run, the entiner export_results fails - losing json and lrc... maybe you could catch it in this method rather than above it.

Axonis (commented):
Overall looks good, just nits.

container_files/controller/create_pool.py
enhaut: I don't think this has much usecase outside of container that has `yaml` dependency installed due to trex recipes support. `create_pool.py` is supposed to be used with test environment desc file, which is as of now containerized controller container thing only.

But made it lazy loaded with raising exception if anyone decides to run it outside

container_files/controller/entrypoint.sh
enhaut: Fixed, thx

container_files/controller/container_runner.py
enhaut: Wrapped `runner.run()` to try/expect block. thx

container_files/controller/container_runner.py
enhaut: sure, wrapped to try/exec

container_files/controller/container_runner.py
enhaut: removed, thanks

container_files/controller/container_runner.py
enhaut: done

container_files/controller/create_pool.py
jtluka: Ok, we can keep it.

jtluka (dismissed):

Axonis (dismissed):

container_files/controller/container_runner.py
```diff
@@ -68,41 +91,133 @@ def _parse_recipe_params(self) -> dict[str, Any]:
...
-            eval(formatter)
-            for formatter in os.getenv("FORMATTERS", "").split(";")
-            if formatter
-        ]
+    def _export_results(self, recipe, result_dir):
+        os.makedirs(result_dir, exist_ok=True)
+
+        # Export human-readable log (with debug output)
+        hr_fmt = HumanReadableRunSummaryFormatter(level=ResultLevel.DEBUG)
+        with open(os.path.join(result_dir, "controller.log"), "w") as f:
```
olichtne: this only exports the "result summary"... that's useful but maybe i'd like it to be renamed to `result_summary.log`

additionally i'd also like to see the individual logs of the run be exported as:
```
controller.log
host1.log
host2.log
```

files, and these logs are located in `recipe.runs[0].log_list` dictionary with keys `'controller', 'host1, ...`

Or alternatively you can just move/copy the files as they should already be exported in `recipe.runs[0].log_dir`

container_files/controller/container_runner.py
```diff
@@ -68,41 +91,133 @@ def _parse_recipe_params(self) -> dict[str, Any]:
...
         """
         overall_result = ResultType.PASS
+        results: list[tuple[str, ResultType]] = []
+
+        for i, test in enumerate(self._test_db):
+            print(f"\n{'=' * 60}")
+            recipe_name = test["recipe_name"]
+            test_id = test.get("uuid", f"{i}_{recipe_name}")
+
+            test_result = ResultType.PASS
```
olichtne: this is not used since you redefine it in the try-except block

container_files/controller/container_runner.py
```diff
@@ -68,41 +91,133 @@ def _parse_recipe_params(self) -> dict[str, Any]:
...
+        except PermissionError:
+            print(
+                f"Permission denied accessing {path}. "
+                "If this directory is a mounted volume, SELinux may be "
+                "preventing access. Try running the container with "
+                "--security-opt label=disable",
+                file=sys.stderr,
+            )
+            return False
+    return True
```
olichtne, line 239: so this returns `True` if the path is not a directory or doesn't exist? is that intentional?

lnst/Controller/MachineMapper.py
```diff
@@ -311,6 +313,8 @@ def _check_interface_compatibility(self, req_if, pool_if):
                req_label != req_if["network"]:
                 return False
         for param, value in list(req_if["params"].items()):
+            if param in self._SKIP_IF_PARAMS:
```
olichtne: wait what? that basically means these parameters are not used ever again anywhere since this is included in the generic `lnst/Controller/MachineMapper.py` file... not even in our internal test runner where this is actually intended to be used?

the driver/speed/model parameters are something that most likely should be automatically used in this upstream controller container... but maybe that's a question of - it shouldn't be required by the recipes, maybe shouldn't be provided in the test db and likely shouldn't be part of the pool.

if we just blankly skip them here, that's going to effect also situations where we actually want to use these parameters.

.github/workflows/container-build.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Source branch checkout
+      uses: actions/checkout@v4
+
+    - name: Set up QEMU
```
olichtne, line 22: what do we need qemu for?

olichtne (changes requested):

.github/workflows/container-build.yml
enhaut: Podman uses qemu for multi arch building. We can either use arm based runners for arm builds but queue for these runners is way too long (i was waiting for 2 hours and didn't get the runner), so decided to just emulate it on x86...

olichtne: interesting, ok, that's fine then.

container_files/controller/container_runner.py
enhaut: makes sense.

I am now copying content of `.log_dir`

lnst/Controller/MachineMapper.py
enhaut: uuuh thanks, fixed. Definitely not intended.

Now, mapper will ignore matching for parameters with `any` as value

container_files/controller/container_runner.py
enhaut: thanks, fixed

lnst/Controller/MachineMapper.py
olichtne: i think this shouldn't be necessary.

there's two lists:
1. the pool parameters
2. the recipe requirements

The situations are:
* pool defines `param=X`, recipe... doesn't require param -> MATCH OK
* pool defines `param=X`, recipe.... specifies requirement `param=Y` -> MATCH NOT OK
* pool defines `param=X`, recipe.... specifies requirement `param=X` -> MATCH OK
* pool doesn't define `param`, recipe... doesn't require param -> MATCH OK
* pool doesn't define `param`, recipe... specifies requirement `param=X` -> MATCH NOT OK

I think all of these situations make sense, and should already be covered via the original implementation without any changes.

If there should be any support of `ANY` as a value... it should be on the REQUIREMENTS side... and it actually already maps to the situation where `recipe doesn't require param`... this is also the situation where the param simply doesn't have a value.

I think the issue we hit and why you originally included the code was that our TEST DB that we were providing to the container was **requiring** specific values.... and this was the bug, the test db simply shouldn't require those values.

At the same time, the machine pool that is automatically generated, simply shouldn't generate those descriptors for the autogenerated pool.

and we should end up in this situation:
* pool doesn't define `param`, recipe... doesn't require param -> MATCH OK

Please try out this approach as well. i believe it should work and simplify the code rather than including these weird workarounds...

enhaut: Ah... I thought it won't be matched because recipes now defines their requirements...

Just tried it out and it works since recipe reqs are empty strings...


Fixed.

olichtne (approved):

PR #436, opened 2026-04-08
pyproject.toml: losen restriction on urllib3

jtluka:
Check if this is still relevant. This blocks integration with other projects that require higher urllib version.


olichtne:
this will have to be updated on top of the uv changes to pyproject.toml

jtluka:
No longer relevant after migration to uv.

PR #439, opened 2026-04-23
Common/ExecCmd.py: handle gracefully terminated commands

jtluka:
Resolves the following recipe case:
```
dump = m1.run("tcpdump -i eth0", bg=True)
ctl.wait(2)
dump.kill(signal.SIGINT)
```

The agent would silently discard following error and will not cleanup the Job reference from its state.
```
Process Process-45:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/lnst/Agent/Job.py", line 107, in _run
    self._job_cls.run()
  File "/usr/local/lib/python3.9/site-packages/lnst/Agent/Job.py", line 238, in run
    stdout, stderr = exec_cmd(self._what["command"], json=self._what["json"])
  File "/usr/local/lib/python3.9/site-packages/lnst/Common/ExecCmd.py", line 65, in exec_cmd
    (data_stdout, data_stderr) = subp.communicate(input = stdin)
  File "/usr/lib64/python3.9/subprocess.py", line 1134, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib64/python3.9/subprocess.py", line 1995, in _communicate
    ready = selector.select(timeout)
  File "/usr/lib64/python3.9/selectors.py", line 416, in select
    fd_event_list = self._selector.poll(timeout)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib64/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.9/site-packages/lnst/Agent/Job.py", line 119, in _run
    result["result"] = job_result
UnboundLocalError: local variable 'job_result' referenced before assignment
```

Controller would report following in cleanup phase:
```
2026-04-21 13:20:49           (host1)        -   ERROR: [Errno 3] No such process
```

The code does not intentionally raise an exception as the keyboard interrupt is typically intentional and graceful termination. So, the stdout and stderr is collected and raising an exception is based on returncode check that is done further in the code.

### Description
(Please provide an overall description of what your Merge request is trying to
do.)

### Tests

J:12626902

### DIFF
--- a/lnst/Common/ExecCmd.py
+++ b/lnst/Common/ExecCmd.py
@@ -62,7 +62,12 @@ def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=F
     subp = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE, stdin=subprocess.PIPE,
                             close_fds=True)
-    (data_stdout, data_stderr) = subp.communicate(input = stdin)
+    try:
+        (data_stdout, data_stderr) = subp.communicate(input = stdin)
+    except KeyboardInterrupt:
+        data_stdout = subp.stdout.read()
+        data_stderr = subp.stderr.read()
+
     data_stdout = data_stdout.decode()
     data_stderr = data_stderr.decode()
 


lnst/Recipes/ENRT/IpsecEspAeadRecipe.py
```diff
@@ -81,10 +81,9 @@ def test_wide_configuration(self, config):
         ipv6_addr = {host1: interface_addresses(self.params.net1_ipv6),
                      host2: interface_addresses(self.params.net2_ipv6)}
         for host in [host1, host2]:
-            host.eth0.down()
+            host.eth0.up()
```
olichtne: do we know why this happens or is required? i think when we originally wrote the tests the assumption was that the ipaddress must be configured in this order and that's why it was there... so it would be interesting to learn a bit more about how this is working

lnst/Common/ExecCmd.py
```diff
@@ -62,7 +62,12 @@ def exec_cmd(cmd, die_on_err=True, log_outputs=True, report_stderr=False, json=F
     subp = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE, stdin=subprocess.PIPE,
                             close_fds=True)
-    (data_stdout, data_stderr) = subp.communicate(input = stdin)
+    try:
+        (data_stdout, data_stderr) = subp.communicate(input = stdin)
```
olichtne, line 66: so since exec_cmd is used in... a lot of places... we should do a very wide comprehensive test for this to check there are no unpredictable side effects, i think one rhel version is ok for that - e.g. i'd run it with rhel10 only, but probably all possible tests are good to check just to be extra sure.

olichtne (commented):
some discussion comments first.

Also, you mention in the description of the PR that you'll add tests, but so far no tests can be seen other than the CI for this PR

lnst/Common/ExecCmd.py
jtluka: Scheduled full internal candidate test run on RHEL10: job id 12626902

lnst/Recipes/ENRT/IpsecEspAeadRecipe.py
jtluka: I hit several times that without the fix, the expected IP addresses were not available on the interfaces and ping/iperf connections failed. Once applied, this did not happen anymore. This could be related to the LNST deployment - containerized controller with setup_agents.sh script + test execution.

I did not investigate why this happens further. Also I think the interfaces should be up when configuring addresses, however we have the same configuration flow in other recipes (e.g. SimpleNetworkRecipe does not explicitly set up() the interfaces) and we did not have this problem.

jtluka: Now I realized that this could be the result of NetworkManager stepping in. In the containerized controller + setup_agents.sh test bed the NetworkManager deployment did not have the NetworkManager-config-server package installed, so the activated interface was managed by NetworkManager.

jtluka: I removed the commit from this PR.

lnst/Common/ExecCmd.py
jtluka: All failures in that run were unrelated to changes in this MR. This should be safe to merge.

jtluka:
> some discussion comments first.
> 
> Also, you mention in the description of the PR that you'll add tests, but so far no tests can be seen other than the CI for this PR

I dropped the ipsec recipe commit from the MR, irrelevant now.

olichtne (approved):

PR #442, opened 2026-05-12
lnst/Common/Logs.py: reset export logger between set/unset_recipe

jtluka:
### Description

Without this the export_handler logger handler would append all recipes executed by a single Controller instance into a shared list.

The export_handler initialization/teardown is moved to set_recipe() and unset_recipe() methods to avoid that.

### Tests

Tested locally with attached script: [test-containers.py](https://github.com/user-attachments/files/27626144/test-containers.py)
```
uv run  ./test-containers.py
```

And then inspecting the log files with another script: [dump-lrc-log.py](https://github.com/user-attachments/files/27626164/dump-lrc-log.py)
```
uv run ./dump-lrc-log.py /tmp/lnst_container_recipe1_run.lrc controller| less
uv run ./dump-lrc-log.py /tmp/lnst_container_recipe2_run.lrc controller| less
```

In the dumped logs, check for `Summary of used Recipe parameters` ... the second dump will include also run from previous recipe (ipv6 AND ipv4).
### Reviews

@olichtne @enhaut @Axonis

### DIFF
--- a/lnst/Common/Logs.py
+++ b/lnst/Common/Logs.py
@@ -101,7 +101,7 @@ class LoggingCtl:
     agents = {}
     transmit_handler = None
     _id_seq = 0
-    log_list = {}
+    log_list = None
 
     def __init__(self, debug=False, log_dir=None, log_subdir="", colours=True):
         #clear any previously set handlers
@@ -130,11 +130,6 @@ class LoggingCtl:
         self.display_handler = logging.StreamHandler(sys.stdout)
         self.display_handler.setFormatter(MultilineFormatter(colours))
 
-        # the export_handler will add log messages to lists, so it could be exported to .lrc file
-        self.log_list["controller"] = []
-        self.export_handler = self._create_export_handler(self.log_list["controller"], colours)
-        self.export_handler.setLevel(logging.DEBUG)
-
         if not debug:
             self.display_handler.setLevel(logging.INFO)
         else:
@@ -146,7 +141,6 @@ class LoggingCtl:
         logger = logging.getLogger()
         logger.setLevel(logging.NOTSET)
         logger.addHandler(self.display_handler)
-        logger.addHandler(self.export_handler)
 
     def unset_formatter(self):
         self.display_handler.setFormatter(None)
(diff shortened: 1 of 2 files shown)


lnst/Common/Logs.py
```diff
@@ -182,11 +176,23 @@ def set_recipe(self, recipe_path, clean=True, prepend=False, expand=""):
         logger.addHandler(recipe_info)
         logger.addHandler(recipe_debug)
 
+        # reset the export_handler
+        logger.removeHandler(self.export_handler)
```
olichtne: is this required if it is also part of the `unset_recipe` method and you removed the export handler creation from initialization>

lnst/Controller/Controller.py
```diff
@@ -179,6 +179,7 @@ def run(self, recipe, **kwargs):
                     raise
                 finally:
                     self._cleanup_agents()
+                    self._log_ctl.unset_recipe()
```
olichtne, line 182: hmmm actually looking at it now... maybe this shouldn't be here and the "remove handler" from "set_recipe" should be kept

just in case there are some logs between this finally block and the next iteration of the matcher/controller run - these will be appended to the previous recipe run, which may not be 100% correct, but at least they will be part of SOME lrc export, just in case they're useful.

jtluka: Well, I added this to resolve the issue when:
```
ctl.run(recipe1)
ctl.run(recipe2)
```

The exported lrc for recipe2 would contain any logging messages that are related to pool preparation (e.g. containers setup) that happen between the two runs.

Additionally, the current placement of `LogginCtl.set_recipe()` (and all logging setup) results in recipe1 not being able to include this pool setup logs in any way.

Imo, the pool setup is not relevant for the RecipeRun context. Only the match description.

lnst/Common/Logs.py
jtluka: Yeah, most likely makes sense. I was just following the original logic.

olichtne: ok, remove this as it doesn't do anything

lnst/Controller/Controller.py
olichtne: ok keep this as is.

lnst/Common/Logs.py
jtluka: Removed

PR #445, opened 2026-06-08
Fix XDP cpu eval filter

Axonis:
### Description

XDP recipes (drop, tx) evaluate aggregate CPU utilization on both test hosts. The generator host only pushes traffic toward the receiver - it does not run the XDP program. Because the generator spreads packet transmission across multiple cores while the receiver pins IRQs to a single CPU, their aggregate CPU profiles are very different.

When machines rotate between generator and receiver roles across pipeline runs, the rolling baseline for a given host absorbs values from both roles. This makes the baseline unstable and causes false positive CPU eval detections that have nothing to do with actual XDP performance.

#### Changes

- Add an overridable `cpu_perf_evaluation_hosts` property to `BaseEnrtRecipe` that controls which hosts are included in CPU baseline evaluation. Defaults to all matched hosts, so existing recipes are unaffected.
- Override it in `XDPDropRecipe` and `XDPTxRecipe` to only evaluate CPU on host2 (the receiver), where the XDP workload actually runs.

CPU data is still collected and saved on both hosts - only the evaluation step is scoped to the receiver.

### Tests
TODO - run XDP

### Reviews
@olichtne @enhaut @jtluka 

Closes: #

### DIFF
--- a/lnst/Recipes/ENRT/BaseEnrtRecipe.py
+++ b/lnst/Recipes/ENRT/BaseEnrtRecipe.py
@@ -502,6 +502,16 @@ class BaseEnrtRecipe(
         return evaluators
 
     @property
+    def cpu_perf_evaluation_hosts(self):
+        """Hosts to include in CPU performance evaluation.
+
+        Override to limit CPU evaluation to specific hosts, e.g. only the
+        receiver in an XDP drop test where the generator's aggregate CPU
+        is irrelevant.
+        """
+        return list(self.matched)
+
+    @property
     def cpu_perf_evaluators(self):
         """CPU measurement evaluators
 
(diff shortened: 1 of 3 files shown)


olichtne:
this could theoretically be a useful feature... so not discarding it YET.

however, the initial analysis of xdp related issues is incorrect, and IMO there is no reason to evaluate only the `host2` part of the recipe for xdp... which would mean that we would only keep the base enrt definition of "all hosts" and well... that changes nothing for now...

so for now let's work a bit more on the analysis side of things, and until it is proven that this is needed i'm blocking this.

Axonis:
Built on wrong assumptions, closing.

PR #446, opened 2026-06-26
CTRecipes: fix ct module unload on RHEL 10+

enhaut:
### Description
`nft_counter` was merged into `nf_tables` in kernel 5.17 and doesn't exist as separate module on RHEL 10+. Added check if loaded before attempting removal.

### Tests
rhel9 J:12809859

rhel10 J:12809860

### Reviews
@jtluka

### DIFF
--- a/lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
+++ b/lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
@@ -48,5 +48,8 @@ class NftablesConntrackMixin(NftablesMixin):
 
         host2.run("systemctl stop nftables")
         time.sleep(2)
-        host2.run("modprobe -rv nft_ct nf_conntrack_netlink nf_conntrack nft_counter")
+        # nft_counter merged into nf_tables in kernel 5.17, doesn't
+        # exist as separate module on RHEL 10+
+        host2.run("modprobe -rv nft_ct nf_conntrack_netlink nf_conntrack")
+        host2.run("if grep -q '^nft_counter ' /proc/modules; then modprobe -rv nft_counter; fi")
 


lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
```diff
@@ -48,5 +48,8 @@ def remove_sub_configuration(self, config):
 
         host2.run("systemctl stop nftables")
         time.sleep(2)
-        host2.run("modprobe -rv nft_ct nf_conntrack_netlink nf_conntrack nft_counter")
+        # nft_counter merged into nf_tables in kernel 5.17, doesn't
+        # exist as separate module on RHEL 10+
+        host2.run("modprobe -rv nft_ct nf_conntrack_netlink nf_conntrack")
+        host2.run("if grep -q '^nft_counter ' /proc/modules; then modprobe -rv nft_counter; fi")
```
jtluka, line 54: The command can be simplified to:
```
grep -q '^nft_counter ' /proc/modules && modprobe -rv nft_counter
```

olichtne (approved):

jtluka (commented):
Overall looks ok, I suggest to simplify the command.

lnst/Recipes/ENRT/ConfigMixins/NftablesConntrackMixin.py
enhaut: I don't think so... `and commands list` will return nonzero exit code if any command from list fails which will be evaluated as a fail in lnst. We can do

```
grep -q '^nft_counter ' /proc/modules && modprobe -rv nft_counter || true
```

But that'll mask `modprobe` failure as well, which i don't like...

jtluka: Missed that, you're right. Resolving and approve.

jtluka (approved):

PR #447, opened 2026-07-24
Extend NftablesRuleScaleRecipe parameterization

SirPhuttel:
Most crucial change is in the last commit which changes 'rule' parameter into ListParam. This allows users to specify multiple rules which are added repeatedly.

The next best thing is in commit three which opens the possibility to add arbitrary ruleset elements, e.g. to test chain jumps with a rule like 'jump mychain'. This requires for that chain to exist and passing 'extras=["add chain {family} {tablename} mychain"]' makes it happen.

To add a bit of convenience to that new 'extras' parameter, parameterize family, table and chain names in commit 2 (not really necessary but a low hanging fruit if one wants to support format strings in 'extras'.

Finally, the first commit fixes up docs after the 'chainspec' enhancement.

### DIFF
--- a/.github/runner_all_enrt.py
+++ b/.github/runner_all_enrt.py
@@ -97,7 +97,7 @@ for recipe_name in dir(enrt_recipes):
                 "add chain {family} {tablename} otherchain",
                 "add rule {family} {tablename} {chainname} counter"
         ]
-        params["rule"] = "accept"
+        params["rule"] = ["jump otherchain", "accept"]
         params["scale"] = 1
 
     if issubclass(recipe, OffloadSubConfigMixin):
--- a/.github/runner_all_enrt.py
+++ b/.github/runner_all_enrt.py
@@ -93,6 +93,10 @@ for recipe_name in dir(enrt_recipes):
         params["family"] = "ip"
         params["tablename"] = "mytable"
         params["chainname"] = "mychain"
+        params["extras"] = [
+                "add chain {family} {tablename} otherchain",
+                "add rule {family} {tablename} {chainname} counter"
+        ]
         params["rule"] = "accept"
         params["scale"] = 1
 
--- a/.github/runner_all_enrt.py
+++ b/.github/runner_all_enrt.py
@@ -90,6 +90,9 @@ for recipe_name in dir(enrt_recipes):
         params["perf_tool_cpu"] = [0]
 
     if recipe_name == "NftablesRuleScaleRecipe":
+        params["family"] = "ip"
+        params["tablename"] = "mytable"
+        params["chainname"] = "mychain"
         params["rule"] = "accept"
         params["scale"] = 1
 
--- a/lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
+++ b/lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
@@ -12,6 +12,18 @@ class NftablesRuleScaleRecipe(SimpleNetnsRouterRecipe, NftablesMixin):
     :any:`chainspec` parameter) which routes test traffic between host1 and a
     local netns.
 
+    :param family:
+        The family to use for the tested ruleset.
+    :type family: :any:`StrParam` (default "inet")
+
+    :param tablename:
+        The name to use for the tested ruleset's table.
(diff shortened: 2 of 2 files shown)


.github/runner_all_enrt.py
```diff
@@ -90,7 +90,14 @@
         params["perf_tool_cpu"] = [0]
 
     if recipe_name == "NftablesRuleScaleRecipe":
-        params["rule"] = "accept"
+        params["family"] = "ip"
+        params["tablename"] = "mytable"
+        params["chainname"] = "mychain"
+        params["extras"] = [
+                "add chain {family} {tablename} otherchain",
+                "add rule {family} {tablename} {chainname} counter"
```
olichtne, line 98: so since `counter` is also specified as a `params["rule"]` the final result is going to have that rule appear also in `extras` but also in the `rule*scale` generation yes?

i know this is a test, just trying to understand the usecase of extras compared to the rule generation

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
```diff
@@ -31,36 +50,51 @@ class NftablesRuleScaleRecipe(SimpleNetnsRouterRecipe, NftablesMixin):
...
+        extras = self.params.get('extras')
         rule = self.params.get('rule')
         scale = self.params.get('scale')
         desc = super().generate_sub_configuration_description(config)
-        msg = f"NftablesScale: Ruleset with {scale} times '{rule}'"
-        msg += f" in a chain of '{chainspec}'"
+        msg = "NftablesScale: Ruleset with"
+        if extras:
+            msg += f" extra commands '{extras}' and"
+        msg += f" {scale} times '{rule}' in a chain of '{chainspec}'"
```
olichtne: no need to also include the `{f_t_c}` description here since it is now also configurable?

.github/runner_all_enrt.py
SirPhuttel: Yeah, this is a goofy test which adds that counter rule to the chain upfront before the 'scale'(=1) times counter rule and accept rule.
I wanted to "test" referring to available format strings. Maybe I should change the rules in 'rule' property to make it less confusing. Turning one test rule to 'jump otherchain' is a natural one since it kind of justifies the existence of that 'extras' list.

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
SirPhuttel: Well, table and chain names are really just names without functional effect but indeed, that description should mention the family as well.

.github/runner_all_enrt.py
olichtne: ok np, makes sense to me, no need to change the rules

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
olichtne: yes, but being configurable suggests that there's a reason to potentially change them sometimes and that to me implies that it's also better to report that explicitly for cases where it could have value for debugging or understanding failures.

.github/runner_all_enrt.py
SirPhuttel: I changed from ['counter', 'accept'] to ['jump otherchain', 'accept'].

lnst/Recipes/ENRT/NftablesRuleScaleRecipe.py
SirPhuttel: Included table and chain names now. The sentence is a bit long now, though (from a local test run):
`NftablesScale: Ruleset with extra commands '['add chain {family} {tablename} foo', 'add chain {family} {tablename} bar']' and 50 times '['fib saddr . iif oif missing counter drop', 'tcp dport 27374 counter drop']' in a table named 'mytable' and chain named 'testchain' with 'type filter hook forward priority filter' in family 'ip'`

olichtne: i think that's fine for now... we'll see how that evolves into the future, changing the format is relatively easy later on anyway... thanks for the update!

olichtne (approved):

PR #448, opened 2026-08-10
Recipes: added XDPRedirectCPU recipe

enhaut:
### Description
Adding `XDPRedirectCPU` recipe that benchmarks redirecting packets between CPUs using XDP. We only support `l4-dhash` because we need to spread packets to separate cpus. LNST generates multiple flows with the same IP but src/dst port differs and only `l4-{s,d}port` program supports src/dst port hashing and so, spreading between cpus.

This also extends `XDPBench.py` but we need to have separate classes handling `xdp-bench redirect-cpu`, it uses a completely different output format...

#### Results structure
`XDPRedirectCpuMeasurement` returns **single** result object with following structure. It's similar to what forwarding recipes return. It cannot return multiple results - one per flow (the same way as iperf measurements does) as `xdp-bench` returns stats aggregated per receiver cpu. It actually spreads flows to separate cpus but this solely depends on `xdp-bench` so we don't know what cpu is handling what flow. Therefore, it returns single, aggregated result.
```
generator_results:
SequentialPerfResult (5 runs)         # iterations
    ParallelPerfResult (4 streams)    # pktgen threads
        SequentialPerfResult               # intervals
            PerfInterval (60)

receiver_results:
SequentialPerfResult (5 runs)         # iterations
    ParallelPerfResult (4 CPUs)        # one per target CPU in cpumap
        SequentialPerfResult               # intervals
            PerfInterval (60)

forwarded_results:
SequentialPerfResult (5 runs)         # iterations
    ParallelPerfResult (1)                   # single receive IRQ CPU
        SequentialPerfResult               # intervals
            PerfInterval (60)
```

### Tests
`J:13005666`

### DIFF
--- a/lnst/Recipes/ENRT/MeasurementGenerators/XDPRedirectCPUMeasurementGenerator.py
+++ b/lnst/Recipes/ENRT/MeasurementGenerators/XDPRedirectCPUMeasurementGenerator.py
@@ -1,0 +1,71 @@
+from lnst.Common.Parameters import IntParam, ListParam, StrParam
+from lnst.Recipes.ENRT.MeasurementGenerators.BaseFlowMeasurementGenerator import (
+    BaseFlowMeasurementGenerator,
+)
+from lnst.RecipeCommon.Perf.Measurements.XDPRedirectCPUMeasurement import (
+    XDPRedirectCPUMeasurement,
+)
+
+
+class XDPRedirectCPUMeasurementGenerator(BaseFlowMeasurementGenerator):
+    """
+    Measurement generator mixin for XDP redirect-cpu recipes.
+
+    :param perf_tool_cpu: target CPU IDs on the receiver host for cpumap
+        redirect (mandatory). Also used to pin pktgen threads on the generator
+        host (round-robin). The IRQ CPU must not appear in this list.
+    :type perf_tool_cpu: :any:`ListParam`
+
+    :param ratep: pktgen rate limit in pkt/s, -1 for unlimited (default -1)
+    :type ratep: :any:`IntParam`
+
+    :param burst: pktgen burst size (default 1)
+    :type burst: :any:`IntParam`
+
+    :param backlog_size: cpumap queue depth per CPU, passed as ``-q`` to
+        xdp-bench (default 512)
+    :type backlog_size: :any:`IntParam`
+
+    :param xdp_redirect_program: XDP program variant, passed as ``-p`` to
+        xdp-bench. Supported: ``l4-dport`` (default), ``l4-sport``.
+    :type xdp_redirect_program: :any:`StrParam`
+
+    :param xdp_redirect_remote_action: action after cpumap, passed as ``-r``
+        to xdp-bench (default ``drop``)
+    :type xdp_redirect_remote_action: :any:`StrParam`
+    """
+
+    perf_tool_cpu = ListParam(mandatory=True)
+    ratep = IntParam(default=-1)
+    burst = IntParam(default=1)
+    backlog_size = IntParam(default=512)
+    xdp_redirect_program = StrParam(default="l4-dport")
+    xdp_redirect_remote_action = StrParam(default="drop")
+
+    @property
+    def net_perf_tool_class(self):
+        """
+        Partial application - same pattern as :class:`XDPFlowMeasurementGenerator`.
+        Needed to inject extra params into XDPRedirectCPUMeasurement since
+        BaseFlowMeasurementGenerator calls net_perf_tool_class with a fixed
+        signature. See https://github.com/LNST-project/lnst/pull/310#discussion_r1305763175
+        """
+
+        def _factory(*args, **kwargs):
+            return XDPRedirectCPUMeasurement(
+                *args,
+                cpus=self.params.perf_tool_cpu,
+                xdp_program=self.params.xdp_redirect_program,
+                xdp_remote_action=self.params.xdp_redirect_remote_action,
+                backlog_size=self.params.backlog_size,
+                ratep=self.params.ratep,
+                burst=self.params.burst,
+                **kwargs,
+            )
+
+        return _factory
+
+    def generator_cpupin(self, flow_id: int) -> list[int]:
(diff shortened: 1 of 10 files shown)


lnst/Tests/XDPBench.py
```diff
@@ -168,3 +169,184 @@ def _prepare_arguments(self):
...
+                else:
+                    in_kthread_section = False
+
+        if received is None or not forwarded_per_cpu:
+            raise ValueError("Could not parse received/forwarded from block")
+
+        return received, forwarded_per_cpu
+
+
+class XDPBenchRedirectCpu(XDPBench):
```
olichtne: i feel like this entire test module can be separated into its own file and exist standalone without the XDPBench parent.

from what i see the only things these two share are 3 parameters: `interface, interval, duration` or am i missing something?

is there any other reason to keep them tied together other than it feels nice since both are related to "XDPBench"?

lnst/Recipes/ENRT/XDPRedirectCPURecipe.py
```diff
@@ -0,0 +1,108 @@
...
+            logging.warning(f"Could not parse RSS hash key from ethtool -x output")
+            config.rss_original_hash_key = None
+
+    def _restore_rss_hash_key(self, host, dev, config):
+        original_key = getattr(config, "rss_original_hash_key", None)
+        if original_key:
+            host.run(f"ethtool -X {dev.name} hkey {original_key}")
+            logging.info(f"Restored RSS hash key on {dev.name}")
+
+    def _cpu_measurement_kwargs(self):
```
olichtne: what is this? it's not used anywhere, not in this PR, not in anywhere else in LNST code base

lnst/Recipes/ENRT/MeasurementGenerators/XDPRedirectCPUMeasurementGenerator.py
```diff
@@ -0,0 +1,64 @@
...
+    perf_tool_cpu = ListParam(mandatory=True)
+    ratep = IntParam(default=-1)
+    burst = IntParam(default=1)
+    backlog_size = IntParam(default=512)
+    xdp_redirect_program = StrParam(default="l4-dport")
+    xdp_redirect_remote_action = StrParam(default="drop")
+
+    @property
+    def net_perf_tool_class(self):
+        def _factory(*args, **kwargs):
```
olichtne, line 54: you have a measurement generator that doesn't generate measurements but instead overrides the class property by generating a custom init method to create measurements?

this is a very weird pattern, i'd like to see some explanation for it, why not just use standard pattern of generating measurements with a measurement generator?

olichtne (changes requested):

lnst/Recipes/ENRT/MeasurementGenerators/XDPRedirectCPUMeasurementGenerator.py
enhaut: The factory/partial is needed to pass generator configuration (from recipe params) to the test module. `BaseFlowMeasurementGenerator.generate_perf_measurements_combinations` calls `self.net_perf_tool_class(flow_combination, recipe_conf=config)` with a fixed signature, so there's no other way to inject extra params without overriding the whole method.

This was actually discussed in https://github.com/LNST-project/lnst/pull/310#discussion_r1305763175.

iperf recipes doesn't need it because `IperfFlowMeasurement.__init__` only takes (`flows`, `recipe_conf`). everything it needs is in the flow objects (src/dst, ports, duration). `XDPRedirectCPUMeasurement` needs cpus, xdp_program, backlog_size, ... which are generator-level, not flow-level config.

Params could technically be packed into Flow objects but cpus and xdp_program are test-wide (one xdp-bench instance for all flows), not per-flow and putting them to flows is ... meh.

The alternative is to override `generate_perf_measurements_combinations` directly with something like:

```python
def generate_perf_measurements_combinations(self, config):
    combinations = super(BaseFlowMeasurementGenerator, self).generate_perf_measurements_combinations(config)
    for flow_combination in self.generate_flow_combinations(config):
        combinations.append([
            XDPRedirectCPUMeasurement(
                flow_combination,
                cpus=self.params.perf_tool_cpu,
                xdp_program=self.params.xdp_redirect_program,
                xdp_remote_action=self.params.xdp_redirect_remote_action,
                backlog_size=self.params.backlog_size,
                ratep=self.params.ratep,
                burst=self.params.burst,
                recipe_conf=config,
            )
        ])
    return combinations
```
But I would say current implementation is more "elegant"

lnst/Recipes/ENRT/XDPRedirectCPURecipe.py
enhaut: leftovers, removed, thanks.

lnst/Tests/XDPBench.py
enhaut: Originally had it in a separate file but moved it to `XDPBench.py` since it's the same underlying tool. Separated it back out, i don't really mind

lnst/Recipes/ENRT/MeasurementGenerators/XDPRedirectCPUMeasurementGenerator.py
olichtne: ah yes, i see that now, in that linked thread i also said:

> this would probably be best to document in a docstring for anyone else in the future looking at this.

and we do have that here: https://github.com/LNST-project/lnst/blob/master/lnst/Recipes/ENRT/MeasurementGenerators/XDPFlowMeasurementGenerator.py#L30

probably best to also include a similar comment here.

---

Now i also think that there is an alternative proposal - to extend the `Flow` class into a `XDPFlow` which would contain the extra parameters... but at this point imo this partial application is fine as well. just adding this as a note for consideration/historical context if we ever return to this PR discussion.

enhaut: Added docstring, for now i think its fine with this partial application

olichtne (approved):

PR #449, opened 2026-09-11
Fix packet assert job accumulation

olichtne:
### Description
Seen in testing with the containerized controller runner, when running multiple recipes in a row packet assert accumulates data over multiple recipes and reports incorrect results.

This can also eventually lead to OOM killer intervening as a long enough test set simply keeps growing on RAM usage.

### Tests

### Reviews

Closes: #


Axonis:
Cool, makes sense also `LinuxPerfMeasurement` has the exact same pattern, worth fixing here imo.
`start()` appends to the class list, `finish()` reassigns to shadow it per-isntance. Since it has already has init, fix is just moving it there.

Hmmm research this deeper, this could probably also go into linter ruler - flagging mutable class attributes (list, dict, set...) on classes with `__init__`. 

Ruff has `RUF012` which flags this exact thing, but it cannot distinguish the sematnics (sharing on purpose vs unintentionally), so maybe adding custom rule that specifically catches "mutable class attr + mutating method call on self.<attr>" would nail this pattern without false positives.

olichtne:
added an additional commit:

Used uv run ruff check --isolated --select RUF012 lnst/ to check for any
additional issues with using mutable class variables and changing their
state.

This commit removes them as class variables and redefines them as
instance variables to hopefully avoid potential issues or fix hidden
ones that we don't know about.

RUF012 identifies a bunch more, i check all of those individually and
these are all cases where the variable is readonly and it's state isn't
changed. These should be retyped to tuples ideally so that a new issue
doesn't arise, but i would separate that work into it's own PR where we
can address just generic linter warnings and issues.

jtluka:
Looks good, ack.

jtluka (dismissed):

olichtne:
cherrypicked a commit @enhaut : https://github.com/enhaut/lnst/commit/dbe024ae6a899fe8411f7c8b557fce9050f88999 to fix the ci test running for XDP Redirect CPU recipe

PR #450, opened 2026-09-17
Devices.Device: extend up_and_wait check for running and lower_up

olichtne:
### Description
Checking only the IFF_UP state is insufficient for any practical purposes as it only refers to the administrative UP state.

In practice, we almost always want to wait for the device to be actually "running" which requires checking 2 more state flags - LOWER_UP and RUNNING.

I considered separating this into a separate condition/method/check for semantic logic reasons. the up_and_wait method sets only the admin UP state and the wait is introduced due to netlink returning immediatelly without confirmation that the request is applied. So it makes sense to wait for just the UP state which was requested.

However, I checked and all of the uses of the up_and_wait method are in cases where it is semantically understood that the device is actually running after this condition is met so because of that IMO it makes sense to keep this as a single method.

### Tests
I tested this manually in the CNB lab on bnxt nics where we saw the issue appear the most frequently, likely due to:
* bnxt is potentially a bit slower to start running
* the switch in the CNB lab is historically known to maybe have a slightly slower auto-negotiation setup

The monitored issue was that ping rate for GreTunnelRecipe would be <100 SOMETIMES... and we would see ~90 pings pass, where it would be the first pings in sequence that are dropped.

What increased the likelyhood of the issue was to drop the ARP cache as often as possible ideally during the eth and gre tunnel setup.

With that i managed to reproduce the issue 7 out of 14 attempted test runs - 50% likelyhood of the issue.

After this fix was applied, the issue went away and i had 17 successfully passing test runs.

In addition to this i'll schedule a quick pipeline run in our lab to check if anything serious could be breaking somewhere else...


olichtne:
`J:1200171870`

jtluka:
Code wise looks ok.

jtluka (dismissed):

olichtne:
in beaker testing there's one failure reported:

```
    PASS 41_TestResult:	Ping result --- From: <host1 (192.168.6.2)> To: <host2 (192.168.7.2)>
    PASS 42_TestResult:	measured rate 100 is more than min_rate(50)
    PASS 43_TestResult:	Packet assert successful, packets received 100, expected min(100) max(0)
    PASS 52_TestResult:	Ping result --- From: <host1 (6001:db8:ac10:fe01::2)> To: <host2 (7001:db8:ac10:fe01::2)>
    PASS 53_TestResult:	measured rate 100 is more than min_rate(50)
    FAIL 54_TestResult:	Packet assert unsuccessful, packets received 99, expected min(100) max(0)
```

i think, this is unrelated to this MR, the ping rate itself is 100 and that is the main point of this PR...

the packet assert is potentially a separate issue that needs more investigation

with that i'd like to proceed with merging this in.

jtluka (approved):

